PR #10, opened 2015-12-17
Fix test setup

camjc:
- We forgot to change cfg/test.js to use isparta instead of
  istanbul instrumenter.


robmunro:
LGTM

Merging?

robmunro:
:trophy:

PR #12, opened 2015-12-18
Added custom checkbox and radio buttons

robmunro:
Adding Checkboxes and Radio buttons to library
![screen shot 2015-12-18 at 12 28 07 pm]
screen shot 2015-12-22 at 11 00 33 am


robmunro:
Now adding tests.

src/components/Main.js
```diff
@@ -98,6 +101,48 @@ class AppComponent extends React.Component {
...
+          <Checkbox label="Disabled" disabled />
+        </div>
+        <div className="example-component-panel">
+          <Checkbox label="Checked and Disabled" checked disabled/>
+        </div>
+        <h1>Radio Buttons</h1>
+        <div className="example-component-panel">
+          <RadioGroup name="testRadio" value="2">
+            <Radio
+              label="Unchecked"
```
omgaz: double indent props

src/components/Main.js
```diff
@@ -98,6 +101,48 @@ class AppComponent extends React.Component {
...
+          <Checkbox label="Unchecked" />
+        </div>
+        <div className="example-component-panel">
+          <Checkbox label="Checked" checked />
+        </div>
+        <div className="example-component-panel">
+          <Checkbox label="Disabled" disabled />
+        </div>
+        <div className="example-component-panel">
+          <Checkbox label="Checked and Disabled" checked disabled/>
```
omgaz: space after disabled and before /

package.json
```diff
@@ -71,6 +72,7 @@
     "react-addons-test-utils": "^0.14.3",
     "react-bootstrap": "^0.28.1",
     "react-hot-loader": "^1.3.0",
+    "react-icheck": "^0.1.1",
```
omgaz: Note, but ok. react-icheck seems to have build errors, dependencies out of date, and still noted as in development.

src/components/Main.js
robmunro: Think that is more of a jade thing. Jade it is hard to tell the diff between attributes and the start of an element. Not so much of a problem IMO in JSX.

package.json
robmunro: Yeah noticed that. This is still being worked on - and like the fact that it was built off same style/markup structure as iCheck. The svg's are all there now though, so I we decide to change then shouldn't be too hard.

src/components/Main.js
omgaz: :+1:

omgaz:
Looks good, had a chat about tests and everything #Hairy needs proper convos in the new year

omgaz:
Merging :meat_on_bone:

robmunro:
Totes @omgaz  - not super happy with it - but I think we are going to need to use the standard renderer rather than shallow so that we can use ```findRenderedDOMComponentsWithTag``, ```scryRenderedDOMComponentsWithTag``` if we want to fix that. But that is up to you guys:)

PR #18, opened 2016-01-04
Extract css and update bootstrap export variables

camjc:
- Bugfix button overrides for active/hover/focus and the disabled color


robmunro:
Nice buddio - happy to merge this?

camjc:
@robmunro Yeah :)

PR #25, opened 2016-01-06
Update Dependencies, notably Alexandria and ESLint

camjc:
- add webpack-sources as dev dep until webpack/extract-text-webpack-plugin#142
- Clean up scss and examples for Button.
- Remove :active outlines created by bootstrap.
- Run JSCS in `posttest`


src/components/Main.js
```diff
@@ -31,37 +32,41 @@ class AppComponent extends React.Component {
       <div className="index">
         <h1>Buttons</h1>
         <div className="btn-panel">
-          <Button className="btn-inverse">
-            Default
+          <Button>
```
etaoins: Should this be `<Button className="btn-inverse">`?

etaoins:
LGTM :green_apple:

src/components/Main.js
camjc: Yeah probably. Do you think its wise to enforce no default button by making default button look wrong?

robmunro:
Merging

PR #26, opened 2016-01-06
New: Add react-select component

omgaz:
Using [react-select](https://github.com/JedWatson/react-select) this PR aims to integrate multi-select boxes into adslot-ui.

 - Custom style overrides
 - single and multi examples and tests
 - Downgraded sass naming convention errors to warnings until we can do inline/file based rules/exclusions
 - Added spacing between example components
 - `allowCreate` dependant on react-select issue 660 (not linking directly because github seems to spam the projects issue with audits :speaker: )

screen shot 2016-01-12 at 11 54 34

### DIFF
--- a/src/components/distributionEntry.js
+++ b/src/components/distributionEntry.js
@@ -2,6 +2,7 @@
 
 require('styles/_bootstrap-custom.scss');
 require('styles/_icheck-custom.scss');
+require('styles/_react-select-custom.scss');
 require('styles/_react-toggle-custom.scss');
 
 import Button from 'react-bootstrap/lib/Button';
@@ -24,6 +25,8 @@ import {
   Totals,
 } from 'alexandria-adslot';
 
+import Select from 'react-select';
+
 module.exports = {
   Alert,
   Breadcrumb,
@@ -37,6 +40,7 @@ module.exports = {
   Radio,
   RadioGroup,
   Search,
+  Select,
   Slicey,
   Tab,
   Tabs,
--- a/src/styles/App.scss
+++ b/src/styles/App.scss
@@ -23,4 +23,8 @@ body {
   > .btn-panel + .btn-panel {
     margin-top: $spacing-etalon;
   }
+
+  > br {
+    line-height: $spacing-etalon;
+  }
 }
--- a/src/styles/_react-select-custom.scss
+++ b/src/styles/_react-select-custom.scss
@@ -1,0 +1,100 @@
+@import 'variable';
+
+// control options
+$select-input-border-color: $color-border-lighter;
+$select-input-border-radius: $border-radius-base;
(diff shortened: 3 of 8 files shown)


src/components/distributionEntry.js
```diff
@@ -3,6 +3,7 @@
 require('styles/_bootstrap-custom.scss');
 require('styles/_icheck-custom.scss');
 require('styles/_react-toggle-custom.scss');
+require('styles/components/Select.scss');
```
camjc: I think we should be using the same pattern as above (since its overriding styles for an existing component rather than a new custom component.

src/styles/App.scss
```diff
@@ -22,4 +22,8 @@ body {
   > .btn-panel + .btn-panel {
     margin-top: $etalon-spacing;
   }
+
+  > .example-component-panel {
```
camjc: I don't know about this. Lets look at this holistically together soon.

src/styles/components/Select.scss
```diff
@@ -0,0 +1,83 @@
+@import '../variable';
+
+// control options
```
camjc: I think these are global now. Should be scoped under `.Select`

src/styles/components/Select.scss
```diff
@@ -0,0 +1,83 @@
...
+
+
+.Select {
+  &:not(.is-disabled) {
+    .Select-control {
+      box-shadow: 0 1px 0 0 $color-border-light;
+
+      &:hover,
+      &:focus {
+        box-shadow: 0 2px 0 0 $color-border-light;
```
camjc: you generally don't need the last 0 in box-shadows

src/styles/App.scss
camjc: come to think of it we may as well just use `<br />`s

src/styles/components/Select.scss
omgaz: Need these here so they override the defaults and so that .Select can use these from react-selects entry point.

src/styles/_react-select-custom.scss
```diff
@@ -0,0 +1,83 @@
+@import 'variable';
+
+// control options
+$select-input-border-color: $color-border-lighter;
+$select-input-border-radius: $border-radius-base;
+$select-input-border-focus: $select-input-border-color;
+$select-input-height: 30px;
```
camjc: $spacing-etalon ?

omgaz: Just a coincidence that they're the same. It's 26 now anyway.

camjc:
LGTM

mcslash:
LGTM :peach:

mcslash:
Merging!

PR #33, opened 2016-01-13
Fix: Improve styling of Nav (Bootstrap Tabs)

camjc:
- Stop users being able to select the text of the tabs.
- Prevent size popping when async loading tab icons.
- Make last tab's hover `border-right` present when its `:active`.


omgaz:
LGTMerging :ramen:

PR #34, opened 2016-01-13
Update: Link colour

camjc:
- Link colour has changed from cyan to normal blue.
- This is to be more consistent with the rest of the internet.


omgaz:
Merging :ramen:

camjc:
Going to have to do an audit when we update to this version of adslot-ui in direct-web. Not this repo's concern though.

PR #35, opened 2016-01-14
Update: TreePickerNode include style from success to primary

camjc:
- Reflects the semantics of the action more accurately.


omgaz:
LGTMerging :ramen: :fish_cake:

PR #36, opened 2016-01-14
Update: TreePicker remove nodes from available if they are selected.

camjc:
- This lets us simplify the Node component to always be shown first if
  it is selected.


omgaz:
LGTMerging

PR #41, opened 2016-01-19
Fix: Implicit line height declaration breaking external Select use

omgaz:
- Set line height of .Select-value so labels and icons appear correctly aligned outside of examples

From this (notice white padding at bottom of select option value)
screen shot 2016-01-19 at 15 40 14

To this (notice how it fills the space)
screen shot 2016-01-19 at 15 40 56


camjc:
LGTM merging.

PR #43, opened 2016-01-19
Breaking: TreePicker modifications to make it usable

camjc:
- When there is no type or no path don't render the node's metadata.
- Make getSelected an Object instead of a function,
  call it initialSelection.
- Provide a callback for getSubtree
  so that it can be used asynchronously.


omgaz:
:fish_cake: LGTMerging :ramen:

PR #48, opened 2016-01-21
Breaking: TreePicker modifications

camjc:
- Since TreePicker state relies on the callback from getSubtree,
  just return null inside render if the callback hasn't fired yet.
- Sort nodes for display because its easier to read through.
- return the selected node ids grouped by rootTypeIds because we don't
  know if the consumer will have the ids unique across all rootTypes.
- Stop modifying node.path directly,
  clone it first so we don't keep reversing in a loop.
- Generalise the required message so it doesn't matter
  what the names of the rootTypes are.
- Fix CSS issue where switching tabs may resize the 50% panes.
- Tests: Make rootTypes use letters for ids to disambiguate between
  rootType ids and node ids.
- Modal: explicit font-weight instead of relying on cascade.


vinteo:
LGTM

omgaz:
:fish_cake: LGTMerging :ramen:

PR #49, opened 2016-01-21
Fix: Reset to initial selection when picker is cancelled

vinteo:
Clone the initial selection to break reference when loading into the component state so it can then reset to the initial selection when cancelled


camjc:
I guess this needs a test? `is not equal` but `is deep equal`

camjc:
Great. LGTM. Merging.

PR #50, opened 2016-01-22
Update: Add `allowEmptySelection` prop to MultiPicker/UserMultiPicker

vinteo:
When `allowEmptySelection` is set to true, the `Apply` button of the pickers will be disabled when no items are selected


src/components/adslotUi/MultiPickerComponent.js
```diff
@@ -67,7 +74,7 @@ class MultiPickerComponent extends React.Component {
           <Button className="btn-inverse" onClick={this.cancelAction}>
             Cancel
           </Button>
-          <Button bsStyle="primary" onClick={this.applyAction}>
+          <Button bsStyle="primary" onClick={this.applyAction} disabled={this.disableApplyButton()}>
```
camjc: Pretty sure this would occur a performance hit like it does in angular (set a prop don't always be calling a function)

vinteo:
@camjc changes made

camjc:
LGTM. Merging

PR #57, opened 2016-01-28
Fix: TreePicker lifecycle

camjc:
- Had issues with mounting and unmounting Treepicker due to async data.
- Handle the async data as suggested at
  https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html
- This is annoying for shallow rendering in tests, so we manually call
  componentDidMount after we render the element to mimic
  what would happen in a normal render.


src/components/adslotUi/TreePickerComponent.js
```diff
@@ -19,21 +19,32 @@ class TreePickerComponent extends React.Component {
...
 
+    this.state = {};
+  }
+
+  componentDidMount() {
+    this._isMounted = true;
     this.loadData();
   }
 
+  componentWillUnmount() { this._isMounted = false; }
```
antklim: Where do we call this method?

omgaz:
Coverage then LGTM

vinteo:
LGTM

src/components/adslotUi/TreePickerComponent.js
camjc: Yeah, will have to test it.
React calls it, its a lifecycle method.
Our linting guarantees that lifecycle methods will always be at the top of the class btw.

vinteo:
Merging...

PR #59, opened 2016-01-28
Breaking: TreePicker add averageWithinRootType

camjc:
- averageWithinRootType: false causes each node to be summed
- averageWithinRootType: true
  causes each node to be averaged per rootType
- To achieve the old behaviour set averageWithinRootType: true.


etaoins:
LGTM :+1:

vinteo:
LGTM

PR #69, opened 2016-02-04
Update: TreePicker reorder subcomponents

camjc:
- change `contains` to `includes` to prepare for lodash v4
- Add FlexSpacer, then use it to fill out empty spaces in TreePicker
- Move Totals to the bottom, with Alerts just above them.
- Add deepSetObjectMutability helper. This is a good candidate for
  splitting into its own repo in the future. Right now its just being
  used for the tests to guarantee we aren't mutating initial mock data.

screen shot 2016-02-04 at 5 04 57 pm
screen shot 2016-02-04 at 5 04 48 pm
screen shot 2016-02-04 at 5 04 25 pm


src/styles/adslotUi/TreePickerSelected.scss
```diff
@@ -11,8 +11,10 @@
   }
 
   &-scrollable {
-    $scrollable-height: $treepicker-height + $header-height - 128px;
+    $scrollable-height: $treepicker-height + $header-height - 139px;
```
antklim: The other `magicKraken`?

test/components/adslotUi/MultiPickerComponentTest.js
```diff
@@ -4,8 +4,9 @@
 import createComponent from 'helpers/shallowRenderHelper';
 import Button from 'react-bootstrap/lib/Button';
 import Modal from 'react-bootstrap/lib/Modal';
-import React from 'react';
 import MultiPickerComponent from 'components/adslotUi/MultiPickerComponent';
+import React from 'react';
+import { deepFreeze } from 'helpers/deepSetObjectMutability';
```
antklim: FYI: there is [ImmutableJS](https://facebook.github.io/immutable-js/). Made by facebook as well as reactjs.

test/components/adslotUi/TreePickerNodeComponentTest.js
```diff
@@ -17,6 +18,7 @@ describe('TreePickerNodeComponent', () => {
     path: ['USA', 'NY'],
     isExpandable: true,
   };
+  deepFreeze(newYorkNode);
```
antklim: Why do we need to freeze it? Since `newYorkNode` is constant why is it possible to change it?

src/styles/adslotUi/TreePickerSelected.scss
omgaz: :ocean: :octocat: :ocean:

src/styles/adslotUi/TreePickerSelected.scss
```diff
@@ -22,16 +24,22 @@
...
 
+    // First rootType Grid
     .grid-component {
       &:first-child {
         border-top: 0;
       }
     }
+
+    // Totals Grid
+    + .grid-component {
```
omgaz: spacer +

test/components/adslotUi/MultiPickerComponentTest.js
camjc: Yep, its cool but for a slightly different purpose.
ImmutableJS gives you perf gains when you want to modify objects only by reference,
what I'm doing here is making it so you can't modify the objects at all.

test/components/adslotUi/TreePickerNodeComponentTest.js
camjc: const makes it so you can't reassign a variable once its assigned.
freeze makes it so you can't change properties of an Object.
deepFreeze makes it so you can't change the properties of any sub-object within the Object.

antklim:
Superb!!! Thanks for explanations :sunglasses: 
LGTM :tropical_drink:

src/styles/adslotUi/TreePickerSelected.scss
camjc: We realised that we can't do this because the flexspacer-component is **inside** the treepickerselected-component-scrollable

src/styles/adslotUi/TreePickerSelected.scss
camjc: You're completely right to hassle me about this. I fixed it and also realised that `+ $header-height` was supposed to be `- $header-height` anyway.

omgaz:
Merging :ramen:

PR #76, opened 2016-02-09
Breaking: TreePicker remove child on select parent

camjc:
- BREAKING CHANGE: Path inside each node is now
  an Array of Objects {id, label}.
- This lets us remove subnodes when selecting
  a parent node.


omgaz:
LGTM

etaoins:
LGTM

omgaz:
Merging

PR #77, opened 2016-02-10
Build: Refactor to remove duplication

camjc:
- Reuse node propType.
- Reuse mocks for MultiPicker.
- Abstract out some similar checks across
  multiple tests.


test/components/adslotUi/MultiPickerComponentTest.js
```diff
@@ -1,28 +1,21 @@
...
-  const items = [teamMember1, teamMember2, teamMember3];
-
-  deepFreeze([teamMember1, teamMember2, teamMember3, itemHeaders, items]);
-
+  const {
+    itemHeaders,
+    items,
+    labelFormatter,
+    teamMember1,
+    teamMember2,
```
OndrejCholeva: did you lose teamMember3 on purpose?

test/components/adslotUi/TreePickerComponentTest.js
```diff
@@ -263,19 +267,13 @@ describe('TreePickerComponent', () => {
...
-    runComponentDidMount({ shallowRenderer });
+    const componentRender = getShallowTreePickerPure();
+    const { shallowRenderer } = componentRender;
+    let { treePickerPureElement } = componentRender;
 
-    let treePickerPureElement = getTreePickerPureElement(component);
     treePickerPureElement.props.expandNode(actNode);
 
-    component = shallowRenderer.getRenderOutput();
+    let component = shallowRenderer.getRenderOutput();
```
antklim: Is it supposed to be changed? Otherwise it's better to use `const`

test/components/adslotUi/TreePickerComponentTest.js
```diff
@@ -293,19 +291,15 @@ describe('TreePickerComponent', () => {
...
-    runComponentDidMount({ shallowRenderer });
-    let treePickerPureElement = getTreePickerPureElement(component);
+    const componentRender = getShallowTreePickerPure();
+    const { shallowRenderer } = componentRender;
+    let { treePickerPureElement } = componentRender;
 
     treePickerPureElement.props.expandNode(actNode);
-    component = shallowRenderer.getRenderOutput();
+
+    let component = shallowRenderer.getRenderOutput();
```
antklim: Same here. @326L you used `const`. Looks quite inconsistent.

test/components/adslotUi/TreePickerComponentTest.js
camjc: I abstracted the first use into `getShallowTreePickerPure`.
This still gets reassigned (after the expander in github) so it needs to be a `let`.
The linters automatically flag any of these that are wrong so you don't need to look for them in code review.
(if its assigned twice it makes sure its a let, if its assigned once it makes sure its a const. It never allows a `var`)

test/components/adslotUi/TreePickerComponentTest.js
camjc: See above, @326 I only assign it once.

antklim:
@camjc thanks for explanations :book: 
LGTM

pphminions:
LGTM :candy:

etaoins:
LGTM. Merging!

PR #108, opened 2016-02-24
Update: Optional link buttons for ListPicker

vinteo:
image


camjc:
Would be cool to use indices for the tests. Looks awesome otherwise. :trophy:

test/components/adslotUi/ListPickerComponentTest.js
```diff
@@ -19,8 +19,19 @@ describe('ListPickerComponent', () => {
     teamMember2,
   } = ListPickerMocks;
 
+  const componentsHash = {
```
camjc: We've been calling it indices and having a new object for each level so you can see the relationship properly for when you have to reorder them. Also lease freeze the Objects so we can guarantee they aren't messed with later on in the tests.

camjc:
LGTM. Merging!

PR #109, opened 2016-02-24
Fix: TreePicker new initialSelection rebuilds

camjc:
- When initialSelection is changed to a set of new
  ids, we rebuild the state using the new initialSelection.
- Remove requirement of TreePickerNav because the lower
  components should take care of that.


vinteo:
LGTM

omgaz:
LGTM :fish_cake:

vinteo:
Merging...

PR #111, opened 2016-02-26
Update: SVG Symbol support for empty states

vinteo:
image

image

image

image


camjc:
I guess you need to add svgforeverybody and do some cross browser tests too.
The icon you're adding looks too big for its circle.

omgaz:
Does the test runner spew a bunch of web-server 404 warnings in the console, like with direct-web? If so is there a way we can mute it?

Otherwise, LGTM!

vinteo:
svgforeverybody is bundled with alexandria and should run with it. Tested with IE, it works

no 404 errors

vinteo:
@camjc @omgaz merge?

camjc:
merging!

PR #114, opened 2016-02-26
New: Treepicker totals suffix

camjc:
- Can pass a totalsSuffix to TreePicker for display.

screen shot 2016-02-26 at 3 35 50 pm


pphminions:
LGTM :hocho:

OndrejCholeva:
Looks a bit weird to me (like a currency). Can it be suffix of the label instead?

omgaz:
Should this not just be part of the valueFormatter?

camjc:
@OndrejCholeva It needs to be linked stronger to the pricing. Design and product teams have looked at heaps of alternatives and this is the best so far. It may change in the future.

@omgaz Nope, we only want to show it in totals, not on every node.

omgaz:
Merging :ramen:

PR #117, opened 2016-03-02
Breaking: TreePicker Svg Icons replacing imgs

camjc:
- Uses latest Alexandria, so exports SvgSymbol,
  SvgSymbolCircle.


vinteo:
LGTM

vinteo:
when we put this into direct-web will need to make sure the styles for the product mapper empty state still holds

scripts/commit-msg
```diff
@@ -46,7 +47,7 @@ function checkGitCommit() {
   }
 
   // Only check non-release messages
```
omgaz: only check non-release messages and ignore greenkeeper style chores with super-long package names

omgaz:
Merging :ocean:

PR #124, opened 2016-03-07
New: Confirm modal

vinteo:
image


src/components/adslotUi/ConfirmModalComponent.js
```diff
@@ -0,0 +1,55 @@
...
+    <Modal className="confirm-modal-component" show={show} bsSize="small" keyboard={false}>
+      <Modal.Header>
+        <Modal.Title>{modalTitle}</Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        <p>{modalDescription}</p>
+      </Modal.Body>
+      <Modal.Footer>
+        <Button className="btn-inverse" onClick={cancelAction}>Cancel</Button>
+        <Button bsStyle="primary" onClick={applyAction}>Confirm</Button>
```
omgaz: Should this `confirm` text be configurable through props. Seems like Adslot will want a more conversation tone, something more like `I'm Sure`, giving the user ownership of their actions.

src/components/adslotUi/ConfirmModalComponent.js
```diff
@@ -0,0 +1,55 @@
...
+      </Modal.Footer>
+    </Modal>
```
omgaz: We have a lot of boilerplate for modals, is this maybe a good point to create a wrapper?

vinteo: I don't know, it would probably end up with too many options

src/components/adslotUi/ConfirmModalComponent.js
vinteo: Yea, I think we should only add it when we actually need it

OndrejCholeva: If you make it customizable, you can reuse the component to get rid of following ones: https://github.com/Adslot/direct-web/tree/master/common/client/angular/modalConfigs/campaignManagement

vinteo:
@omgaz added button label options and made cancel button optional

vinteo:
modal title is also optional

test/components/adslotUi/ModalConfirmComponentTest.js
```diff
@@ -0,0 +1,143 @@
...
+    modal: {
+      header: 0,
+      body: 1,
+      footer: 2,
+    },
+    modalFooter: {
+      cancelButton: 0,
+      applyButton: 1,
+    },
+  };
```
omgaz: It'll be nice when we have enzyme and can ignore all of this

omgaz:
LGTMerging :ramen:

PR #134, opened 2016-03-15
Update: Add dts for treepicker components

OndrejCholeva:
Adding dts for key treepicker components, so it can be later used in e2e test for line item targeting.

### DIFF
--- a/src/components/adslotUi/SplitPaneComponent.js
+++ b/src/components/adslotUi/SplitPaneComponent.js
@@ -4,8 +4,8 @@ import React, { PropTypes } from 'react';
 
 require('styles/adslotUi/SplitPane.scss');
 
-const SplitPaneComponent = ({ children }) => (
-  <div className="splitpane-component">
+const SplitPaneComponent = ({ children, dts }) => (
+  <div className="splitpane-component" if dts data-test-selector={dts}>
     {children}
   </div>
 );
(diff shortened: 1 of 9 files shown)


pphminions:
LGTM :chestnut:

src/components/adslotUi/SplitPaneComponent.js
```diff
@@ -13,5 +13,6 @@ const SplitPaneComponent = ({ children }) => (
 SplitPaneComponent.displayName = 'AdslotUiSplitPaneComponent';
 SplitPaneComponent.propTypes = {
   children: PropTypes.node,
+  dts: PropTypes.string,
```
camjc: should there be a default one? can we require this? (especially since it will just be `data-test-selector=""` otherwise)

camjc:
Should we update the tests to use these? (guaranteeing they're there, and cleaning up the tests a little)

src/components/adslotUi/SplitPaneComponent.js
OndrejCholeva: Hmm, default dts doesn't sound cool on second thought, neither does requiring it...
How about `<div className="splitpane-component" if dts data-test-selector={dts}>`?

test/components/adslotUi/SplitPaneComponentTest.js
```diff
@@ -6,6 +6,7 @@ describe('SplitPaneComponent', () => {
   it('should have its component name as default className', () => {
```
omgaz: …and no data-test-selector

omgaz:
LGTM :computer:

OndrejCholeva:
@pphminions @camjc @omgaz merge?

pphminions:
Merging :rocket:

PR #137, opened 2016-03-18
Fix: dist containing DTS

OndrejCholeva:
Forgot to `run dist` with https://github.com/Adslot/adslot-ui/pull/134, fixing it now.
(the adslot-ui-main.js is a result of dist of current master)

screenshot 2016-03-18 13 31 47


OndrejCholeva:
No need for that as there is already merge that includes newest version of dist anyway. Just need to release that.

PR #146, opened 2016-04-01
Fix: Adding gitattributes to get Git always use LF for line ending.

davepoon:
Adding .gitattributes to get Git always use LF for line ending. This config will resolve the line ending issue on windows.
Because git on windows will automatically convert LF to CRLF for line endings by default, which cause eslint and jscs to throw validateLineBreaks error.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf


.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
vinteo: does `text=auto` work? We use that for our other repos

davepoon: No, I tried it, it doesn't. It's same as "core.autocrlf true" cmd, it still convert LF to CRLF on windows.

davepoon:
@vinteo Merged? or any concern or test for this change?

vinteo:
Merging...

PR #149, opened 2016-04-05
Update: Allow for unselectable tree nodes

etaoins:
For segment targeting sometimes we need e.g. country nodes to show a path to specific regions but the country itself isn't selectable. To support this add an optional `isSelectable` property to the tree nodes which defaults to true.


test/mocks/TreePickerMocks.js
```diff
@@ -22,8 +22,14 @@ const rootTypes = [
 const auPath = { id: 'au', label: 'AU' };
 const actPath = { id: 'au-act', label: 'ACT' };
 
-const actNode =
-  { id: 'au-act', label: 'Australian Capital Territory', type: 'State', path: [auPath], value: 1000, rootTypeId: 'a' };
+const actNode = {
+  id: 'au-act',
+  label: 'Australian Capital Territory',
+  type: 'State',
+  path: [auPath], value: 1000, rootTypeId: 'a',
```
omgaz: Can these three be on their own lines?

omgaz:
`Gary requesting 2 change(s)`

etaoins:
@omgaz Didn't even notice that. Looking good now?

omgaz:
`omgaz approving 1 change(s)`

camjc:
@etaoins Ya'll don't want to assert that a node is at least either *selectable* or *expandable*?
Shouldn't slow down perf much and would stop consumers putting their nodes into an unusable state.

PR #156, opened 2016-04-08
Fix: Flex overflow issues in Safari/IE

omgaz:
Uses flex for width

Dependent on https://github.com/Adslot/alexandria/pull/146/

IE Bug:
ie

Safari Bug:
screen shot 2016-04-08 at 11 55 13


omgaz:
Tested with Ryan in IE/Chrome/FF/Safari

vinteo:
Merging...

PR #183, opened 2016-04-22
Fix: Pin sass-lint at 1.5.1

etaoins:
It appears that sass-lint 1.6.0 barfs on some adjacency (+) selectors we use without giving a useful message. Pin at 1.5.1 for now.


tchcxp:
LGTM and Merge.

Thanks for fixing this.

PR #195, opened 2016-04-29
New: Hide Treepicker empty on null subtree

etaoins:
Currently there is no way to suppress the empty state of the treepicker during load. This means if the treepicker is being asynchronously populated (e.g. from REST) there will be an empty state flash.

Fix by considering a subtree of [] empty but hiding the empty state on null/undefined


mcslash:
Yup, that'll do it. ⚡

etaoins:
@mcslash Merge?

mcslash:
⚡ ⚡ Merging ⚡ ⚡

mcslash:
🚀

PR #199, opened 2016-05-02
Fix: fastStatelessWrapper props check

etaoins:
When checking for property changes fastStatelessWrapper was checking to see if any property matched instead of all properties matching. Fix and add a test.


tchcxp:
LGTM 👍

davepoon:
LGTM :thumbsup:

etaoins:
@tchcxp @davepoon Merge?

tchcxp:
Merging...

PR #209, opened 2016-05-09
New: List picker supports controllerFormatter

tchcxp:
This PR adds controllerFormatter in list picker component.

Current list picker component has a label to describe items, and a checkbox/radio button to select/deselect the item.

This PR adds support for adding an extra button to apply more business rule (eg. `required` or not) on the item with an external component, which is `controllerFormatter`.

For further details, please check:
https://trello.com/c/FKN8jNjE/5312-segments-adslot-ui-components

### Screenshot:

With `controllerFormatter`:

image

Without `controllerFormatter`:

image


src/components/Main.js
```diff
@@ -113,6 +115,7 @@ class AppComponent extends React.Component {
     const listPickerItemHeaders = {
       left: 'Team',
       right: 'Member',
+      optional: 'Required',
```
antklim: This is confusing 😄

test/components/adslotUi/ListPickerPureComponentTest.js
```diff
@@ -59,6 +59,24 @@ describe('ListPickerPureComponent', () => {
...
+
+    const headerGridElement = component.find(Grid).first();
+    const gridHeaderElement = headerGridElement.find(GridRow);
+    const gridHeaderCellElements = gridHeaderElement.find(GridCell);
+    expect(gridHeaderCellElements.last().children().text()).to.equal('Required');
+
+    const gridElement = component.find(Grid).last();
+    const gridRowElements = gridElement.find(GridRow);
+    gridRowElements.forEach((gridRowElement) => {
+      expect(gridRowElement.find(GridCell).last().find(Checkbox).length).to.equal(1);
```
vinteo: the last cell in a row can also have 1 checkbox if the `controllerFormatter` is not there, either:

- also assert the number of elements with `.grid-component-cell-toggle` for each row, or
- pass a different component into `controllerFormatter` other than `Checkbox` or `Radio`

vinteo:
LGTM

davepoon:
LGTM :thumbsup:

OndrejCholeva:
Having column of checkboxex next to column of radios feels strange to me. Given that radio allows user to only select single item, there can be only single required that is obviously applied to selected item. 

@pedroscoulloupas ?

tchcxp:
@OndrejCholeva Please check the trello card attached above.
The checkbox in here is an external component and it can be a checkbox, radio button or other UI component. The screenshot attached above is only for demo purpose to clarify that now we are able to add an external component in the list picker.

src/components/Main.js
```diff
@@ -90,6 +90,8 @@ class AppComponent extends React.Component {
 
     const labelFormatter = (item) => `${item.givenName} ${item.surname}`;
 
+    const controllerFormatter = () => (<Checkbox />);
```
OndrejCholeva: `controllerFormatter` accepts no parameters, but you are passing `item` on line https://github.com/Adslot/adslot-ui/pull/209/files#diff-04d820ed82558ab61555c230f537aa84R69

tchcxp: The `main.js` is the file that we can preview the component. As an external component, `controllerFormatter` will definitely receive parameters (in this case, `item`), for applying further business rules (eg. when deselecting the radio button on the left, the checkbox on the right will be deselected and disabled.).

The reason why we did not put a parameter in `main.js` is because we don't need to apply a complex `controllerFormatter` in here for previewing purpose. FYI, please check the use of `labelFormatter` in direct-web for greater details.

OndrejCholeva:
Thanks @tchcxp  :-) LGTM

tchcxp:
@OndrejCholeva My pleasure. 👍 

@davepoon @vinteo @antklim merge please?

vinteo:
Merging...

PR #210, opened 2016-05-09
Fix remove console log in dist

tchcxp:
I attached local version of `Adslot-Alexadria` with adding `console.log` in the `dist`.

 This PR simply removes the console.log and rebuilds dist files.


vinteo:
Merging...

PR #213, opened 2016-05-12
Update: list picker component header name

tchcxp:
As discussed with @omgaz , we update the header name to `addon`.


sharmajyo:
LGTM

src/components/Main.js
```diff
@@ -115,7 +115,7 @@ class AppComponent extends React.Component {
     const listPickerItemHeaders = {
       left: 'Team',
       right: 'Member',
-      optional: 'Required',
+      addon: 'Required',
```
mcslash: Add on is two words or hyphenated, therefore it should probably be `addOn` (camel case).

omgaz: We use addon internally as does bootstrap.

omgaz:
LGTM merging 🍜  🍥

PR #215, opened 2016-05-17
Upgrade: Alexandria dependency

omgaz:
- Also fixes the distribution file that was incorrectly resolved with conflicts.


sharmajyo:
LGTM

PR #223, opened 2016-05-23
Breaking: Fix inline Datepicker

omgaz:
Input elements should not be `inline` by default, inline is an opinionated view that breaks the bootstrap grid system.

- Make the input container block, and use Bootstraps grid system for constraining floats and positioning as per guideline.
- Add a pointer cursor and placeholder to the input for improved experience.

--

> **N.B** This is seen as a breaking change as existing usages of datepicker will need to ensure they are properly wrapped in Bootstrap grid system (see example in `Main.js`).

screen shot 2016-05-23 at 10 24 02


davepoon:
LGTM :thumbsup:

sharmajyo:
LGTM

PR #240, opened 2016-06-21
New: Add button and tab icons

patrick-ivers:
#### Changes
- Change port number to 8001 to prevent conflicts when running alongside alexandria
- Refactor to use FlexibleSpacer and .flexible-wrapper-inline from alexandria
- Add icons to buttons and tabs with FlexibleSpacer
- Upgrade alexandria

#### Trello
https://trello.com/c/OWT71p6Z/5420-adslot-ui-add-support-for-icons-in-buttons-also-update-adslot-ui-port-number-to-8001

#### Screenshots
screen shot 2016-06-22 at 11 30 04 am


src/styles/helpers.scss
```diff
@@ -0,0 +1,9 @@
+.flex-wrapper-inline {
+  display: flex;
+  flex-direction: row;
+
+  & .flexspacer-component {
```
vinteo: isn't the `&` redundant here?

etaoins:
It seems like the `FlexWrapper`/`FlexSpacer` name is describing the implementation more than the functionality or usage. Maybe `LabelRow`/`LabelSpacer`?

mcslash:
There's too much space between the icon and the label in the tab component. Also, it's probably not in the scope of this PR, but we need a better + icon if that's the one we're using.

mcslash:
I think the reason it looks off is because its taller than it is wide.

patrick-ivers:
@mcslash The problem here is with the icon itself. That particular SVG icon has a bit of padding either side because it is tall. I could put another "squarer" icon in as a better example.

mcslash:
Yeah I was just thinking that haha. Yeah, if its just an example, maybe just switch out the icon. We would try to use a squarish icon in most of these cases anyway.

patrick-ivers:
@etaoins Good point. `LabelRow / LabelSpacer` isn't quite generic enough though. It's main purpose is to add flexible spacing. It just happens that the underlying implementation is using `flex`. I think that in this regard it accurately describes the functionality. Maybe we could change it to `FlexibleSpacer`?

src/styles/helpers.scss
patrick-ivers: Will fix.

src/components/Main.js
```diff
@@ -227,8 +239,19 @@ class AppComponent extends React.Component {
...
-            <Tab eventKey="Billing" title="Billing" disabled>Billing content</Tab>
+            <Tab
+              eventKey="Audience"
+              title={<span className="flex-wrapper-inline"><SvgSymbol /><FlexSpacer />Audience</span>}
+            >
+              Audience content
+            </Tab>
+            <Tab
+              eventKey="Billing"
+              title={<span className="flex-wrapper-inline">+<FlexSpacer />Billing</span>}
```
davepoon: That could be an alternative way of outputting the customized svg icon and the title using an array,
`title={[<SvgSymbol key="Billing(or a dynamic key)" href="/Images/SvgSymbols/svg-symbols.svg#add" />, 'Billing']}`

Looks good, man!!

mcslash:
Looks good mate. Did you end up switching out the icon?

patrick-ivers:
@mcslash No, because the extra padding I thought was on the icon was actually a margin-right that didn't need to be there. Unless you think it still needs to be swapped out?

mcslash:
Ahhhhh. Nah all good man. Nice one! That fixed it.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "adslot-ui",
-  "version": "11.0.1",
+  "version": "10.2.1",
```
OndrejCholeva: this doesn't seem right.

package.json
```diff
@@ -90,7 +90,7 @@
     "react-select": "^1.0.0-beta13",
     "react-toggle": "^2.0.1",
     "redux": "^3.3.1",
-    "sass-lint": "1.8.0",
+    "sass-lint": "1.7.0",
```
OndrejCholeva: dtto

omgaz:
Merging 🍜

PR #255, opened 2016-06-30
Fix: fix assigning the correct value to emptySymbol in tree picker pure

pphminions:
Fix assigning the correct value to emptySymbol in the tree picker pure.


omgaz:
LGTM, but I'm not sure if your distjs file is correct. I'm compiling locally and getting diff errors. Which version of node are you running and have you ran npm i before committing / running dist?

omgaz:
LGTM, Merging when green :)

PR #261, opened 2016-07-01
Add dts to FilePicker component

OndrejCholeva:
usage in tests:
``` coffeescript
filePickerElement.sendKeys('absolute/path') #simulates user selecting file by filepicker
```


NamrataTrivedi:
LGTM :open_file_folder:

omgaz:
LGTM

src/components/adslotUi/FilePickerComponent.js
```diff
@@ -50,6 +50,7 @@ class FilePickerComponent extends React.Component {
                 type="file"
                 onChange={this.onChange}
                 accept={this.props.filter}
+                data-test-selector={this.props.dts}
```
vinteo: should we put this on the main `div` and classNames/dts on individual elements so they can be targeted as well or is that overkill?

OndrejCholeva: I was thinking about that, but for e2e purpose we only need to target the `<input type="file">` element (protractor sendKeys simulates both selecting and submitting action).

OndrejCholeva:
@omgaz / @NamrataTrivedi / @vinteo  merge?

PR #269, opened 2016-07-12
Fix: File upload button

patrick-ivers:
#### Changes
- Add styles back to file upload and elements styled as buttons (but aren't actually buttons), by changing style selector from `:enabled` to `:not(:disabled)`

#### Trello
https://trello.com/c/JooIv3Ud/5489-inverse-button-styled-elements-style-is-broken-3h
https://trello.com/c/bmd0ibrR/5490-filepicker-remove-button-style-is-broken


vinteo:
LGTM

vinteo:
Actually should we include some of the different styles in Main.js so we can see if it breaks again?

OndrejCholeva:
As for the https://trello.com/c/bmd0ibrR/5490-filepicker-remove-button-style-is-broken - I think the issue there is more about wrong classes being used for the [X] button, so even though this PR would make it appear right, it is not the right fix.

omgaz:
@OndrejCholeva Yea, the file-picker issue is now more of a technical-debt refactor issue now. I can update that card so the fix and issue is apparent.

patrick-ivers:
@vinteo I can definitely do that, but some of the file picker styles (specifically the styles that make the actual button hidden) are applied on a case by case basis in direct-web. As such, a file picker in adslot-ui would not look quite right, despite having the correct button styles.

omgaz:
@patrick I think @vinteo just means to add an adslot-ui <a href> and <span> with classes .btn and make sure they look correct

patrick-ivers:
@omgaz Ah! 😝

omgaz:
LGTMerging 🐏 🐟

PR #286, opened 2016-07-27
Build: When running dist create an unminified copy

omgaz:
- Split the dist file (non-breaking change)
- CSS is tricky, haven't been able to accomplish it
- Export Tag component (missed in last PR)
- Remove redundant (implied) 'loader' from loaders


OndrejCholeva:
do we need to include `dist/adslot-ui-main-ugly.js` into commit?

omgaz:
That's the purpose of the commit... so that it get's generated because we can't validate if a minified copy is valid otherwise.

pphminions:
LGTM 🍒

vinteo:
LGTM

vinteo:
Merging...

PR #287, opened 2016-07-27
Update: Add dts in treepicker and related components

NamrataTrivedi:
#### Changes
- Added itemType parameter to tree picker and related components to create generic data-test-selectors, combined with node.id
- Corrected itemType to itemProp in listPickerComponent
- Updated tests

#### Trello Card
https://trello.com/c/cCT5Zf3V/5523-e2e-test-for-add-publisher-modal


test/components/adslotUi/ListPickerComponentTest.js
```diff
@@ -24,7 +25,7 @@ describe('ListPickerComponent', () => {
   };
 
   it('should render with defaults', () => {
-    const component = createAndMountComponent(<ListPickerComponent />);
+    const component = createAndMountComponent(<ListPickerComponent itemType={itemType} />);
```
OndrejCholeva: `it('should render with defaults'` means you cannot pass the `itemType={itemType}`.
Fix you default by adding something like: `itemType='item'` on line 206 of `src/components/adslotUi/ListPickerComponent.js`

test/mocks/ListPickerMocks.js
```diff
@@ -1,5 +1,7 @@
 import immutable from 'seamless-immutable';
 
+const itemType = 'example item type';
```
OndrejCholeva: would not be needed once the default will be set properly.

test/components/adslotUi/ListPickerComponentTest.js
NamrataTrivedi: Had a chat with @omgaz about this. Since itemType isn't in default props, so we can pass itemType. Also the test throws an ugly warning if something required isn't passed through.

OndrejCholeva: Yeah, that's what I am telling - if itemType is not in default props and someone will forget to pass it our dts would become `-15402487` (for item.id = 15402487). Is that what we want @omgaz?

omgaz:
LGTM

src/components/adslotUi/TreePickerGridComponent.jsx
```diff
@@ -51,6 +53,7 @@ TreePickerGridComponent.propTypes = {
   emptyText: PropTypes.any.isRequired,
   expandNode: PropTypes.func,
   includeNode: PropTypes.func,
+  itemType: PropTypes.string.isRequired,
```
vinteo: Not too sure about having them as required props ... I think if we have it as required we should set default values. Even a default of `item type`?

NamrataTrivedi: I have set a default value to itemType prop for TreePickerPure and TreePickerSimplePure components which should trickle down to TreePickerGrid and other components now.

OndrejCholeva: @NamrataTrivedi so you can now remove the `itemType` from mock data and unit test it loads correctly with defaults? see https://github.com/Adslot/adslot-ui/pull/287/files#r72378236

NamrataTrivedi: 'should render with defaults' verifies with defaults in ListPicker components tests. I added itemType in there to reuse in rest of the steps.

test/components/adslotUi/ListPickerComponentTest.jsx
```diff
@@ -211,20 +213,20 @@ describe('ListPickerComponent', () => {
   });
 
   it('should change `selectedItems` state after a `deselectItem` action', () => {
-    const props = { initialSelection: getInitialSelection(), items: users };
+    const props = { initialSelection: getInitialSelection(), items: users, itemType };
```
OndrejCholeva: could you remove the unnecessary `itemType` param from tests that are testing other things?

NamrataTrivedi:
@vinteo @OndrejCholeva made the changes. 
Also tested modals on direct-web and it is working as expected.

vinteo:
LGTM

test/components/adslotUi/ListPickerComponentTest.jsx
```diff
@@ -219,12 +221,12 @@ describe('ListPickerComponent', () => {
   });
 
   it('should show modal when `show` is true', () => {
-    const component = createAndMountComponent(<ListPickerComponent show />);
+    const component = createAndMountComponent(<ListPickerComponent itemType={itemType} show />);
```
OndrejCholeva: I still don't understand why we need to pass `itemType={itemType}` here.

test/components/adslotUi/ListPickerComponentTest.jsx
```diff
@@ -219,12 +221,12 @@ describe('ListPickerComponent', () => {
...
     expect(component.prop('show')).to.equal(true);
   });
 
   it('should hide modal when `show` is false', () => {
-    const component = createAndMountComponent(<ListPickerComponent show={false} />);
+    const component = createAndMountComponent(<ListPickerComponent itemType={itemType} show={false} />);
```
OndrejCholeva: and here

test/components/adslotUi/TreePickerComponentTest.jsx
```diff
@@ -311,12 +315,12 @@ describe('TreePickerComponent', () => {
   });
 
   it('should show modal when `show` is true', () => {
-    const component = createAndMountComponent(<TreePickerComponent show />);
+    const component = createAndMountComponent(<TreePickerComponent itemType={itemType} show />);
```
OndrejCholeva: and here... basically the `itemType={itemType}` is only required in `it should render with props` and `it specifically tests something that depends on certain itemType`, not anywhere else.

NamrataTrivedi:
@vinteo @OndrejCholeva merge?

vinteo:
Merging...

PR #312, opened 2016-08-11
Fix: Enter key press for category search

patrick-ivers:
#### Changes
- Add check for event.which (as well as event.keyCode), as Chrome has a bug where it always returns 0 for event.keyCode


vinteo:
LGTM

NamrataTrivedi:
LGTM :cookie:

vinteo:
Merging...

PR #325, opened 2016-08-23
update: treeNodeRenderer method added to treePickerGrid

sharmajyo:
Created a treeNodeRenderer function in treePickerGrid so that we can customise the value in cell instead of showing only label.

screen shot 2016-08-23 at 11 56 19 am


vinteo:
LGTM

src/components/adslotUi/TreePickerNodeComponent.jsx
```diff
@@ -48,6 +49,14 @@ const TreePickerNodeComponent = ({
     { onClick: expandNodeBound } :
     {};
 
+  const renderLabel = (node) => {
```
OndrejCholeva: Why you can't make `treeNodeRenderer: PropTypes.func.isRequired` with `treeNodeRenderer: (node) => node.label,` as a default? (see `valueFormatter`)

OndrejCholeva: and actually rename it all to `labelFormatter` or `labelRenderer` as it describes a bit better what it is doing. (the `tree` word seems extra, as it is just a prop - we don't name props by component)

sharmajyo: its not technically a required property for this component so there is no good reason for forcefully making it required, I suppose...
Also in various open source react components we are using (like react-select) such custom rendering properties are always optional.

OndrejCholeva: Default should work even without required, I just don't see a reason for reinventing existing functionality.

pphminions:
LGTM 🍒

sharmajyo:
@OndrejCholeva review comments added.

vinteo:
Merging...

PR #331, opened 2016-08-26
Update: update tree picker pure to support given empty text component

pphminions:
This will support having a loader component for the tree picker pure. ( should send up the loader component conditionally)

screen shot 2016-08-29 at 1 34 18 pm


src/components/adslotUi/TreePickerSimplePureComponent.jsx
```diff
@@ -17,7 +17,7 @@ const TreePickerSimplePureComponent = ({
   disableInclude,
   emptySvgSymbol,
   expandNode,
-  subtree,
+  loaderStateNode,
```
OndrejCholeva: `i` < `l`

src/components/Main.jsx
```diff
@@ -55,6 +55,7 @@ class AppComponent extends React.Component {
       'setCategory',
       'setCategorySearchString',
       'setSearchValue',
+      'searchOnClear',
```
lteacher: Actually I feel `onSearchClear` or `onClearSearch` are better assuming there is some search field that can be cleared. This way round seems like we are "_searching when clear is performed_"

src/components/adslotUi/TreePickerSimplePureComponent.jsx
```diff
@@ -17,25 +17,28 @@ const TreePickerSimplePureComponent = ({
...
+  subtree,
   svgSymbolCancel,
   svgSymbolSearch,
 }) => {
   const selectableNodes = removeSelected({ subtree, selectedNodes });
-  const emptyText = initialStateNode && _.isEmpty(searchValue) ? initialStateNode : 'No items to select.';
+  let emptyText = initialStateNode && _.isEmpty(searchValue) ? initialStateNode : 'No items to select.';
   const emptySymbol = initialStateSymbol && _.isEmpty(searchValue) ? initialStateSymbol : emptySvgSymbol;
 
+  emptyText = loaderStateNode ? loaderStateNode : emptyText;
```
OndrejCholeva: `emptyText = loaderStateNode || emptyText;` to save few characters?

pphminions:
@OndrejCholeva Incorporated the review comments

src/components/Main.jsx
pphminions: This is a good suggestion. But this parameter was there for a long time with tree-picker and changing this will cause to change direct-web.
and this is not introduced from this PR :)

vinteo:
What do you think of just re-using `emptyText`? Means we don't have to pass in another prop

pedroscoulloupas:
@pphminions Is this loading using the spinner @patrick-ivers has used in the new marketplace?
Also please refer to the design …loading publishers

pphminions:
@pedroscoulloupas yes will be using the same spinner as the marketplace. :)

pphminions:
@vinteo - Incorporated the review comments

pphminions:
Notice that in direct-web, even when we are not sending `emptySvgSymbol` (TreePickerSimplePure) there is an empty circle. But in adslot-ui if `emptySvgSymbol` is not sending it is not showing correctly (as per the screenshot) @mcslash want your thoughts about this :)

vinteo:
LGTM

test/components/adslotUi/TreePickerSimplePureComponentTest.jsx
```diff
@@ -76,13 +76,25 @@ describe('TreePickerSimplePureComponent', () => {
     expect(rightPaneElement.find(FlexibleSpacer)).to.have.length(1);
   });
 
-  it('should render with empty state props', () => {
+  it('should render with give empty state node', () => {
```
OndrejCholeva: given

test/components/adslotUi/TreePickerSimplePureComponentTest.jsx
```diff
@@ -76,13 +76,25 @@ describe('TreePickerSimplePureComponent', () => {
...
+
+    const component = shallow(<TreePickerSimplePure {...loaderStateProps} />);
+    expect(component.find(TreePickerGrid).first().prop('Loading...'));
+  });
+
+  it('should render with default empty state props', () => {
     const emptyStateProps = _.assign({}, props, {
       searchValue: 'Victoria',
       initialSelection: [],
+      emptyText: null,
```
OndrejCholeva: do you need to set the emptyText explicitly here?

pphminions:
@OndrejCholeva Incorporated the review comments

pphminions:
According to the new design, should support the `selected` side empty text and svg image for a given text and image. So I made the changes. @vinteo, @OndrejCholeva please review

screen shot 2016-08-29 at 1 38 55 pm 2

vinteo:
LGTM

src/components/adslotUi/TreePickerSimplePureComponent.jsx
```diff
@@ -76,8 +80,8 @@ const TreePickerSimplePureComponent = ({
         <TreePickerGrid
           {...{
             disabled,
-            emptySvgSymbol,
-            emptyText: 'Nothing selected.',
+            emptySvgSymbol: emptySvgSymbolSelected || emptySvgSymbol,
```
OndrejCholeva: :confused: Can we have better variable names? `emptySvgSymbolSelected` is highly misleading (is it empty symbol which is selected, selected symbol which is empty, both?)...

OndrejCholeva: The whole `empty` concept is just strange as it hold few different states depending on external conditions and is never actually empty.

pphminions:
@vince please merge

src/components/adslotUi/TreePickerSimplePureComponent.jsx
```diff
@@ -29,11 +31,13 @@ const TreePickerSimplePureComponent = ({
   searchPlaceholder,
   searchValue,
   selectedNodes,
+  subtree,
   svgSymbolCancel,
   svgSymbolSearch,
 }) => {
   const selectableNodes = removeSelected({ subtree, selectedNodes });
-  const emptyText = initialStateNode && _.isEmpty(searchValue) ? initialStateNode : 'No items to select.';
+  let searchTextNode = emptyText || 'No items to select.';
```
OndrejCholeva: `let searchStateNode` (similarly to `initialStateNode`)

src/components/adslotUi/TreePickerSimplePureComponent.jsx
```diff
@@ -16,8 +16,10 @@ const TreePickerSimplePureComponent = ({
   disabled,
   disableInclude,
   emptySvgSymbol,
+  emptySelectedListSvgSymbol,
+  emptyText,
```
OndrejCholeva: `searchStateNode` as a prop, as it is no more just a text and it better represents the purpose to component user.

vinteo:
Merging...

PR #358, opened 2016-09-28
Fix: Remove dropdown from category search

patrick-ivers:
#### Changes
- Remove category search dropdown and rename component to "SearchBar"
- Rename export for Search (from alexandria) to "SearchField"

#### Trello
https://trello.com/c/bkglEyJm/5768-marketplace-remove-category-drop-down-from-search-bar


OndrejCholeva:
`ButtonSearch` sound a bit silly, why not a `SearchButton`?

patrick-ivers:
@OndrejCholeva Search button means it's just a button. This means it's a search input with a button. I guess I could call it `SearchWithButton`, but that sounds even sillier. 😝

OndrejCholeva:
`SearchWithButton` at least describes what is it (note that adslot-ui is a public module with no documentation). How about `SearchBar`?

patrick-ivers:
@OndrejCholeva I like `SearchBar`. @omgaz Thoughts?

omgaz:
SearchBar is good but how would that differ from the Alexandria Search Component. Maybe we should export from adslot-ui SearchField and SearchBar to make it easier to differentiate?

patrick-ivers:
@omgaz We could, but is renaming the Search field component something we want to do right now? It seems like a big task. I'll rename `ButtonSearch` to `SearchBar` for now and leave the refactor of `Search` for another PR.

patrick-ivers:
@omgaz Done! 👍

src/components/Main.jsx
```diff
@@ -209,8 +172,12 @@ class AppComponent extends React.Component {
     this.setState({ startDate: newValue });
   }
 
-  performCategorySearch() {
-    console.log(`Searching "${this.state.categorySearchString}"...`);
+  searchOnClear() {
+    this.setSearchValue('');
+  }
+
+  performSearchBarSearch() {
```
OndrejCholeva: What about `searchString` and `performSearch()` instead of forcing searchbar everywhere?
LGTM otherwise.

patrick-ivers: This is just a dummy function in Main.jsx. Also, there's also the TreePicker search stuff that is confusing things so I didn't think it would hurt to be a little more specific.

OndrejCholeva: Oh I see, the `SearchBarComponent` props are ok. The `main.js` always confuses me :-)

patrick-ivers:
@omgaz @NamrataTrivedi @OndrejCholeva Merge please?

omgaz:
Merging 🍜

PR #360, opened 2016-09-29
[WIP] Heading and grouping support in TreePickerSimplePure

mdotwills:
todo:
- props passing and prop args
- testing

There is a lot of copy/paste duplication between `TreePickerSimplePure` and the `TreePickerSelectedComponent` implementation at the moment. Need to get behaviour right before refactor.


mdotwills:
This has now already been implemented in #498

PR #369, opened 2016-10-10
Fix: Shows metadata in treepicker when nodes contain ancestors data

mdotwills:
**Background**
Nodes loaded from the api as already selected describe their path in terms of ancestry with the `ancestors` key.

**Changes**
- Checks both `node.path` and `node.ancestory` when rendering metadata.

**Trello**
https://trello.com/c/Xbs2mGJ0/5656-multiple-segments-description-missing-for-selected-cities-and-states


src/components/adslotUi/TreePickerNodeComponent.jsx
```diff
@@ -34,9 +34,12 @@ const TreePickerNodeComponent = ({
...
   valueFormatter,
 }) => {
-  const pathElement = !_.isEmpty(node.path) ?
+
+  const pathElement = !(_.isEmpty(node.path) && _.isEmpty(node.ancestors)) ?
     <span className={`${baseClass}-path`}>
-      {_(node.path).map('label').clone().reverse().join(', ')}
+      { !_.isEmpty(node.path) ?
+        _(node.path).map('label').clone().reverse().join(', ') :
+        _(node.ancestors).map('label').clone().join(', ') }
```
OndrejCholeva: I think the clone is extra here, in fact `_.map(node.ancestors, 'label').join(', ')` should be just fine.

src/components/adslotUi/TreePickerNodeComponent.jsx
```diff
@@ -34,9 +34,12 @@ const TreePickerNodeComponent = ({
   selected,
   valueFormatter,
 }) => {
-  const pathElement = !_.isEmpty(node.path) ?
+
+  const pathElement = !(_.isEmpty(node.path) && _.isEmpty(node.ancestors)) ?
```
patrick-ivers: Could you change `!(_.isEmpty(node.path) && _.isEmpty(node.ancestors))` to `_.some([node.path, node.ancestors])` ?

src/components/adslotUi/TreePickerNodeComponent.jsx
```diff
@@ -34,9 +34,12 @@ const TreePickerNodeComponent = ({
...
     <span className={`${baseClass}-path`}>
-      {_(node.path).map('label').clone().reverse().join(', ')}
+      { !_.isEmpty(node.path) ?
+        _(node.path).map('label').clone().reverse().join(', ') :
```
patrick-ivers: Is `.clone()` necessary?

OndrejCholeva: clone is only necessary when operating on original array as reverse mutates it. Given that map is creating new array, it is safe to skip it. both reverse and join are native so there no need to even initiate lodash chain:
```
_.map(node.path, 'label').reverse().join(', '):
_.map(node.path, 'label').join(', ')
```

mdotwills: @OndrejCholeva, seems the `clone()` is needed after all (unit tests fail) 😞 Could just be with the way they were setup, but I'm not going to change them for sake of removing a `clone` that was already present.

src/components/adslotUi/TreePickerNodeComponent.jsx
mdotwills: @patrick-ivers does not work, even when specifying the `predicate` arg.

mdotwills:
Merge?

src/components/adslotUi/TreePickerNodeComponent.jsx
OndrejCholeva: `!_.every([node.path, node.ancestors], _.isEmpty)` would work, but anyway...

src/components/adslotUi/TreePickerNodeComponent.jsx
omgaz: The intent of this line I think get's lost in lodash. Can you maybe take another look or at least do this outside of the JSX wrapped in an explanatory function.

test/components/adslotUi/TreePickerNodeComponentTest.jsx
```diff
@@ -7,7 +7,7 @@ import { GridCell, GridRow } from 'alexandria-adslot';
 import { shallow, mount } from 'enzyme';
 
 describe('TreePickerNodeComponent', () => {
-  const { cbrNode, actNode, itemType, nodeRenderer } = TreePickerMocks;
+  const { cbrNode, cbrNodeAlreadySelected, actNode, itemType, nodeRenderer } = TreePickerMocks;
```
omgaz: What was cbr again?

OndrejCholeva: **C**an**b**e**r**ra?

PR #370, opened 2016-10-11
Update: handle disable property in the inverse button

pphminions:
Handle `disabled` for the inverse button. 
There are situation where we use 'ng-disabled' for the buttons. And this is adding `disabled= disabled` to the button. At this point the button should have the disabled properties. 

Removed default button font size 100px style


pphminions:
@vinteo, @omgaz please merge

vinteo:
Merging...

PR #361, opened 2016-09-29
Fix: Downgrade react-icheck to v0.3.3

patrick-ivers:
#### Changes
- Force v0.3.3 of react-icheck to fix problem with checkboxes not registering clicks

PR #381, opened 2016-10-18
Fix: Downgrade react-icheck to v0.3.4

OndrejCholeva:
React-icheck rewrote thier release history, so v0.3.3 is no more available, so latest version was installed instead. That caused E2E to break as well as issue from PR #361 to reappear.


package.json
```diff
@@ -85,7 +85,7 @@
     "react-bootstrap": "^0.29.5",
     "react-datepicker": "^0.29.0",
     "react-hot-loader": "^1.3.0",
-    "react-icheck": "=0.3.3",
+    "react-icheck": "=0.3.4",
```
bchew: `=` is actually not needed https://docs.npmjs.com/misc/semver

OndrejCholeva: I've read that, but decided to go with minimum change. I am fine to remove it if we fell it is easier to read.

omgaz:
LGTM

Make sure you've:
1. rebased with master
2. ran `npm i`
3. ran `npm run dist`

OndrejCholeva:
@omgaz I've spent last 30 minutes doing that (+tests), but still only 95% confident about it 🙈

package.json
bchew: Considering there are other places in the file where this is done, perhaps leave it as is. Could probably be done as a separate PR if warranted..

lteacher:
LGTM!

OndrejCholeva:
@omgaz is there anything more I should do?
```
(running of freshmost master)
n 4.6.0
rm -f node_modules
npm i
npm run dist
npm run test
```

If not, please merge.

omgaz:
merging 🚀

PR #385, opened 2016-10-21
Fix: eslint

omgaz:
ESLint has been broken for a few months now.

Problem has existed since we migrated from .js to ,jsx for our jsx files as per airbnb eslint rules.

 - Update pre-commit hooks to only check staged files
 - Updated messages when running pre-commit
 - Added magnifying glass 🔍 when checking things out
 - Added broken hearts 💔 when something goes wrong
 - Renamed `lint` script to `eslint`
 - Added new `lint` script that lints everything (not just eslint)
 - Mapped `post`test` to `lint`
 - Fix broken lint


README.md
```diff
@@ -56,7 +56,7 @@ npm run dist
 
 # Lint all files in src (also automatically run after tests)
 
-npm run lint
```
omgaz: This should still be lint

src/components/adslotUi/PagedGridComponent.jsx
```diff
@@ -50,7 +50,7 @@ class PagedGridComponent extends React.Component {
         {totalPages > 1 ?
           <div className="pagedgrid-component-pagination">
             <span className="pagedgrid-component-pagination-info">
-              {(activePage - 1) * perPage + 1}–{Math.min(activePage * perPage, items.length)} of {items.length}
+              {((activePage - 1) * perPage) + 1}–{Math.min(activePage * perPage, items.length)} of {items.length}
```
OndrejCholeva: is this necessary?

src/components/adslotUi/TreePickerNodeComponent.jsx
```diff
@@ -34,24 +34,27 @@ const TreePickerNodeComponent = ({
...
+            .map('label')
+            .clone()
+            .reverse()
+            .join(', ')
+      :
+        _(node.ancestors)
+          .map('label')
+          .join(', ')
+      }
+    </span> : null;
```
OndrejCholeva: should be:
```
    </span> 
 :
    null;
```
but I don't think putting semicolons on own line improves anything...

omgaz: Null is kinda the exception I think; @patrick-ivers?

src/components/adslotUi/PagedGridComponent.jsx
omgaz: The lint rules comlain about mixed * and +. I'd have assumed bodmas would take precedence.

ESLint defines the benefit as:

> Enclosing complex expressions by parentheses clarifies the developer’s intention, which makes the code more readable. This rule warns when different operators are used consecutively without parentheses in an expression.

OndrejCholeva: maybe can use the rule with groups: http://eslint.org/docs/rules/no-mixed-operators#groups which makes much more sense (only forcing parens around logical and bitwise operators).

omgaz: Agreed, sounds good :)

src/components/adslotUi/TreePickerNodeComponent.jsx
```diff
@@ -34,24 +34,27 @@ const TreePickerNodeComponent = ({
   selected,
   valueFormatter,
 }) => {
-
-  const printPathText = (node) => _(node.path).map('label').clone().reverse().join(', ');
-  const printAncestorText = (node) => _(node.ancestors).map('label').join(', ');
-
   const pathElement = !(_.isEmpty(node.path) && _.isEmpty(node.ancestors)) ?
     <span className={`${baseClass}-path`}>
-      { !_.isEmpty(node.path) ? printPathText(node) : printAncestorText(node) }
```
OndrejCholeva: This was definitely more descriptive/readable. What was the reason for the change/what's wrong with arrow functions http://eslint.org/docs/rules/func-style#allowarrowfunctions?

omgaz: Good spot, I meant to move these further up :)

omgaz:
Ran direct-web e2e: https://ci.adslot.com/view/Direct/job/ui-test-direct/5328/ all passing

test/components/adslotUi/ListPickerComponentTest.jsx
```diff
@@ -159,12 +159,32 @@ describe('ListPickerComponent', () => {
...
 
     const itemInfoPropertyElements = itemInfoGridElement.find(GridRow);
-    expect(itemInfoPropertyElements.at(1).find(GridCell).first().children().text()).to.equal('Name');
-    expect(itemInfoPropertyElements.at(1).find(GridCell).last().children().text()).to.equal('Jill Smith');
-    expect(itemInfoPropertyElements.at(1).find(GridCell).last().prop('dts')).to.equal('name');
-    expect(itemInfoPropertyElements.at(2).find(GridCell).first().children().text()).to.equal('Age');
-    expect(itemInfoPropertyElements.at(2).find(GridCell).last().children().text()).to.equal('21');
-    expect(itemInfoPropertyElements.at(2).find(GridCell).last().prop('dts')).to.equal('age');
+    expect(itemInfoPropertyElements.at(1).find(GridCell)
+      .first()
```
OndrejCholeva: are we forcing chains in tests as well or is this due to length constraint?

omgaz: Length here, I believe; although in this instance the accessors are being repeated. I'll clean it up; but what are your thoughts on the chaining?

OndrejCholeva: I think multi-line chains are easier to read/simplify especially when there is any nesting. However with exception of chains like `selecting().one().particular().element.property` (i.e. no mutating or complex set of dynamic rules to get the result, but just a simple way to navigate a relatively static object).

Also in this case I don't like the partial wrapping.

vinteo:
Merging...

PR #402, opened 2016-11-07
Fix: Allow more than one type of node structure

mdotwills:
Changes:

- Relaxed `path` property of node type

### DIFF
--- a/src/helpers/propTypes/TreePickerPropTypes.js
+++ b/src/helpers/propTypes/TreePickerPropTypes.js
@@ -11,7 +11,7 @@ export default {
         id: PropTypes.string.isRequired,
         label: PropTypes.string.isRequired,
       }).isRequired
-    ).isRequired,
+    ),
     type: PropTypes.string.isRequired,
     value: PropTypes.number,
   }),


src/helpers/propTypes/TreePickerPropTypes.js
```diff
@@ -2,19 +2,27 @@ import { PropTypes } from 'react';
...
+      id: PropTypes.string.isRequired,
+      isExpandable: PropTypes.bool,
+      label: PropTypes.string.isRequired,
+      type: PropTypes.string.isRequired,
+    }),
+    node: PropTypes.shape({
+      id: PropTypes.string.isRequired,
+      isExpandable: PropTypes.bool,
+      label: PropTypes.string.isRequired,
+      path: PropTypes.arrayOf(
```
OndrejCholeva: What is the reason of having two almost identical node types? What about just making path optional?

mdotwills: ^ much cleaner solution, thought there was a need for explicit `path` on some components but we can do without it.

omgaz:
I don't see why a node would not have a path? Can you explain?

mdotwills:
After offline discussion, we've decided to keep the `path` property, regardless if used explicitly, as a required property as it is fundamental to the tree picker structure concept.

PR #443, opened 2016-11-21
Update: Tree picker node component to support child nodes

pphminions:
Support tree picker node component with child nodes. If `isChildNode` property of a node is true, then it will contain a padding to indent correctly .

screen shot 2016-11-21 at 4 02 05 pm

### DIFF
--- a/src/components/Main.jsx
+++ b/src/components/Main.jsx
@@ -127,7 +127,7 @@ class AppComponent extends React.Component {
       ],
       treePickerPureSubtree: [],
       selectedNodes: [
-        { id: '2', label: 'Norfolk Island', path: [{ id: '11', label: 'AU' }], type: '', isSelectable: false },
+        { id: '2', label: 'Norfolk Island', path: [], type: '', isSelectable: false },
         { id: '3', label: 'Queensland', path: [{ id: '12', label: 'AU' }], type: '' },
       ],
       searchBarString: '',
--- a/src/components/adslotUi/TreePickerNodeComponent.jsx
+++ b/src/components/adslotUi/TreePickerNodeComponent.jsx
@@ -57,7 +57,7 @@ const TreePickerNodeComponent = ({
   const labelCellProps = expanderElement ? { onClick: expandNodeBound } : {};
 
   return (
-    <div className={`${baseClass}`}>
+    <div className={!_.isEmpty(node.path) || !_.isEmpty(node.ancestors) ? `${baseClass} child-node` : `${baseClass}`}>
       <GridRow dts={`${_.kebabCase(itemType)}-${node.id}`}>
         {selected ?
           <GridCell classSuffixes={['button']} dts="button-remove">
--- a/test/components/adslotUi/TreePickerNodeComponentTest.jsx
+++ b/test/components/adslotUi/TreePickerNodeComponentTest.jsx
@@ -11,7 +11,7 @@ describe('TreePickerNodeComponent', () => {
 
   it('should render a node with defaults', () => {
     const component = shallow(<TreePickerNodeComponent itemType={itemType} node={cbrNode} />);
-    expect(component.prop('className')).to.equal('treepickernode-component');
+    expect(component.prop('className')).to.equal('treepickernode-component child-node');
     expect(component.type()).to.equal('div');
 
     const rowElement = component.find(GridRow);
@@ -141,6 +141,16 @@ describe('TreePickerNodeComponent', () => {
     expect(rowElement.prop('children')).to.have.length(5);
     const valueCellElement = rowElement.prop('children')[3];
     expect(valueCellElement).to.be.a('null');
+  });
+
+  it('should not have the child node class for root nodes', () => {
+    const component = shallow(<TreePickerNodeComponent itemType={itemType} node={maleNode} />);
+    expect(component.prop('className')).to.equal('treepickernode-component');
+  });
+
+  it('should have the child node class for child nodes', () => {
+    const component = shallow(<TreePickerNodeComponent itemType={itemType} node={cbrNode} />);
+    expect(component.prop('className')).to.equal('treepickernode-component child-node');
   });
 
   it('should fire expandNode when clicking on the `expand` cell', () => {


mdotwills:
I didn't notice the `isChildNode` property on the node type props, maybe update that too and set the default to false with default props?

src/components/adslotUi/TreePickerNodeComponent.jsx
```diff
@@ -58,7 +58,13 @@ const TreePickerNodeComponent = ({
 
   return (
     <div className={`${baseClass}`}>
-      <GridRow dts={`${_.kebabCase(itemType)}-${node.id}`}>
+      <GridRow dts={`${_.kebabCase(itemType)}-${node.id}`} >
+        {node.isChildNode ?
+          <GridCell>
+            <div className="child-node" />
```
omgaz: If this isn't actually a child-node can we call this child-node-indent-spacer maybe? But again, would much rather see this purely a css/class related thing on the top node.

src/components/adslotUi/TreePickerNodeComponent.jsx
```diff
@@ -58,7 +58,13 @@ const TreePickerNodeComponent = ({
 
   return (
     <div className={`${baseClass}`}>
```
omgaz: Why not add the class, `child-node` here, and apply the padding to `.child-node > .grid-row-component`? It'd be a lot cleaner than adding extra conditionals in the markup and adding extra dom nodes and cells to create something that is purely a spacer.

test/components/adslotUi/TreePickerNodeComponentTest.jsx
```diff
@@ -127,7 +127,7 @@ describe('TreePickerNodeComponent', () => {
     />);
 
     const rowElement = component.find({ dts: `${_.kebabCase(itemType)}-${cbrNode.id}` });
-    const valueCellElement = rowElement.prop('children')[3];
+    const valueCellElement = rowElement.prop('children')[4];
```
omgaz: If you go the css route none of this would have to change either.

src/components/adslotUi/TreePickerNodeComponent.jsx
pphminions: I tried this @omgaz , if you add a padding here it looks like below. The padding apply to the bottom boarder as well
screen shot 2016-11-22 at 8 58 57 am

pphminions:
@mdotwills - added `isChildNode` as a prop of the node, but there I don't see any of the other similar properties having the default value, for example prop `isExpandable`

pphminions:
@omgaz, @mdotwills - incorporated the review comments

src/helpers/propTypes/TreePickerPropTypes.js
```diff
@@ -14,6 +14,7 @@ export default {
     ).isRequired,
     type: PropTypes.string.isRequired,
     value: PropTypes.number,
+    isChildNode: PropTypes.bool,
```
OndrejCholeva: LGTM, just move it before `isExpandable`? 🔠

src/helpers/propTypes/TreePickerPropTypes.js
```diff
@@ -4,6 +4,7 @@ import { SvgSymbol } from 'alexandria-adslot';
 export default {
   node: PropTypes.shape({
     id: PropTypes.string.isRequired,
+    isChildNode: PropTypes.bool,
```
OndrejCholeva: Isn't the `isChildNode` computable from (or at least related to) the `path` prop?
Also elsewhere we are checking `_.isEmpty(node.ancestors))` to figure out if it is child node or not.

@omgaz thoughts?

pphminions: my understanding is 'path' and 'node.ancestors' is used to display the ancestor name with the node label. 
I want something purely to add just a padding 
screen shot 2016-11-22 at 11 46 23 am

OndrejCholeva: @pphminions my point is that we should either think about it from higher level perspective and refactor existing code to use the `isChildNode` prop (to avoid duplication and confutsion) or name it `extraPadding`.

I know the ancestors and path is mostly coming from geo segments and the way tree picker used to be only used for targetting, but I think we should add props wisely with some plan/vision. That's why I am calling Gary, as a I believe he knows the plan.

OndrejCholeva: (the `isSite` was a joke to show that adding props because of specific need at time may lead to bad code design)

omgaz: IIRC

 - Ancestors with type make up that (State in AU) as (type in ancestor)

 - Path makes up breadcrumbs

We could still pursue with adding an is-child class but, as Ondrej mentioned, calculated from whether we have ancestors/path rather than defined by the user. We don't need to add or pass an extra prop. We just apply it to all child nodes; but what about children's childrens? **Maybe we need an `is-child` and `child-depth-x`? So for the future we know how deep.**

Then direct-web can choose to style this how it pleases. This way we provide the information about the node, without dictating how we should style it.

If that's the case, styling in adslot-ui isn't so maintainable or extensible, we'd either end up with all pickers showing this indentation (which we don't want) or we have to add a prop type flag to decide whether or not we style it one way or the other (however adding proptypes to markup that dictate css/style in this case creates too much coupling.

pphminions: @omgaz, @OndrejCholeva - removed the css style for `child-node` and will handle that from the direct-web

src/components/adslotUi/TreePickerNodeComponent.jsx
```diff
@@ -57,7 +57,7 @@ const TreePickerNodeComponent = ({
   const labelCellProps = expanderElement ? { onClick: expandNodeBound } : {};
 
   return (
-    <div className={`${baseClass}`}>
+    <div className={node.isChildNode ? `${baseClass} child-node` : `${baseClass}`}>
```
OndrejCholeva: if I read @omgaz comment below correctly, this should become:
`<div className={node.path || node.ancestors ? `${baseClass} child-node` : `${baseClass}`}>`
without need to have the `isChildNode` prop.

pphminions: If i put path or ancestor, it will effect the ui in unexpected ways. For example, if I set `path` in publisher modal, it will add something which is not there in the expected design 
(please ignore the incomplete path, this is just for example)

I tried this with the current adslot-ui version of the direct-web, might be slightly different in the new version


That is why I need something which is not effecting other properties in any other way. @omgaz is having another node property as `isChildNode` is okay?

OndrejCholeva: I think that Gary meant also to hide the path text in direct-web (if unwanted).
> Then direct-web can choose to style this how it pleases. This way we provide the information about the node, without dictating how we should style it.

src/components/Main.jsx
```diff
@@ -127,8 +127,8 @@ class AppComponent extends React.Component {
       ],
       treePickerPureSubtree: [],
       selectedNodes: [
-        { id: '2', label: 'Norfolk Island', path: [{ id: '11', label: 'AU' }], type: '', isSelectable: false },
-        { id: '3', label: 'Queensland', path: [{ id: '12', label: 'AU' }], type: '' },
+        { id: '2', label: 'Norfolk Island', path: [], type: '', isSelectable: false },
+        { id: '3', label: 'Queensland', path: [{ id: '12', label: 'AU' }], type: '', isChildNode: true },
```
omgaz: We can remove this now.

pphminions:
@omgaz - Incorporated the review comments

omgaz:
Merging 🚀

PR #495, opened 2017-01-17
New: Ability to group nodes in TreePickerGrid component

vinteo:
image

Used in product:

image


src/components/adslotUi/TreePickerGridComponent.jsx
```diff
@@ -69,6 +87,7 @@ TreePickerGridComponent.propTypes = {
 TreePickerGridComponent.defaultProps = {
   disabled: false,
   hideIcon: false,
+  groupFormatter: () => 'Default Group',
```
omgaz: I'm not sure if we should have groups by default. I think if no group is specified it should be flat.

src/components/adslotUi/TreePickerGridComponent.jsx
```diff
@@ -20,34 +23,48 @@ const TreePickerGridComponent = ({
...
+        </div>
+      )}
+      {nodes ?
+        <Empty
+          collection={nodes}
+          hideIcon={hideIcon}
+          svgSymbol={emptySvgSymbol}
+          text={emptyText}
+        /> :
+        null}
```
omgaz: How would we handle empty states by group? Is this a thing. If we have two groups (Adslot/3rd Party) and I search for '20-20 Year olds', nothing matches under Adslot, should I still show the heading with an empty state?

Maynbe a question for Pedros,

src/components/adslotUi/TreePickerGridComponent.jsx
vinteo: Yea it will only show if there is more than one group

src/components/adslotUi/TreePickerGridComponent.jsx
vinteo: AS the search is handled outside the component, the group will not appear if it was empty cause it will only show nodes given to it.

omgaz:
LGTMerging 🍜

PR #500, opened 2017-01-24
Fix: TreePickerNode Spinner on async/large lists expandNode

omgaz:
An issue where loading large subtrees (> 250 nodes) would fail to render the spinner. State was being correctly updated but the view wasn't being redrawn. React was bundling the redraw of the spinner with the redraw of the loading of the new subtree.

 - Add interval to defer rendering the list by 25ms which will also show a spinner for times when the users' machine is running slow (not just async calls)
 - Refactor code slightly (made fixing the bug clearer and component simpler) and noticeably faster (see gif comparison)
 - Moved `getExpander` to it's own component without the need for a func


src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
```diff
@@ -27,73 +28,55 @@ class TreePickerNodeComponent extends React.Component {
...
-    }
-
-    return {};
-  }
+  setLoadingAndExpandNode() {
+    // Using intervals and setState callback to seamlessly and consistently show the spinner despite sync
+    // or async expandNode method. Renders spinner even for slow DOM redraws or large lists.
+    const delayedExpandNode = () => {
+      const intervalToClear = setInterval(() => {
+        this.props.expandNode(this.props.node); clearInterval(intervalToClear);
```
patrick-ivers: Can you put the `clearInterval(intervalToClear);` on its own line?

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
```diff
@@ -27,73 +28,55 @@ class TreePickerNodeComponent extends React.Component {
...
+      },
 
-  // Bind so we can maintain context in async callbacks, and events.
-  bindIncludeNode() {
-    return this.props.includeNode.bind(null, this.props.node);
-  }
+      25);
```
patrick-ivers: Magic number?

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
```diff
@@ -27,73 +28,55 @@ class TreePickerNodeComponent extends React.Component {
...
       itemType,
       node,
+      expandNode,
       nodeRenderer,
       selected,
       valueFormatter,
     } = this.props;
 
-    const pathElement = !(_.isEmpty(node.path) && _.isEmpty(node.ancestors)) ?
+    const isChildNode = !(_.isEmpty(node.path) && _.isEmpty(node.ancestors));
```
patrick-ivers: Yea, I think `_.some(node, ['path', 'ancestors']);` would be better.

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
omgaz: ah yea, cheers for that.

test/components/adslotUi/TreePicker/TreePickerNodeComponentTest.jsx
```diff
@@ -142,71 +143,42 @@ describe('TreePickerNodeComponent', () => {
...
   });
 
-  it('should fire expandNode when clicking on the `expand` cell', () => {
-    const nodes = [];
-    const expandNode = (node) => nodes.push(node);
-    const component = shallow(<TreePickerNodeComponent itemType={itemType} node={cbrNode} expandNode={expandNode} />);
+  it('should fire expandNode when clicking on the label cell', (done) => {
+    const props = {
+      expandNode: (node) => {
+        expect(node).to.deep.equal(cbrNode);
```
patrick-ivers: Is `.to.deep.equal` the same as `.to.eql`?

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
omgaz: Exactly that... magic!

test/components/adslotUi/TreePicker/TreePickerNodeExpanderComponentTest.jsx
```diff
@@ -0,0 +1,21 @@
...
+import { Spinner } from 'alexandria-adslot';
+import { shallow } from 'enzyme';
+
+describe('TreePickerNodeExpanderComponent', () => {
+  it('should render with default isLoading false', (done) => {
+    const component = shallow(<TreePickerNodeExpander onClick={done} />);
+    expect(component.prop('dts')).to.equal('expander');
+    expect(component.children()).to.have.length(1);
+    expect(component.find(Spinner)).to.have.length(0);
+    component.simulate('click');
```
patrick-ivers: That's an unusual way to end the test. Why have you done it like this? Is there some async thing going on?

OndrejCholeva: `onClick={done}` on line 9 picks that I guess.

src/components/adslotUi/TreePicker/TreePickerNodeExpanderComponent.jsx
```diff
@@ -0,0 +1,26 @@
...
+const TreePickerNodeExpander = ({
+  isLoading,
+  onClick,
+}) => (
+  <GridCell onClick={onClick} dts="expander">
+    {
+      isLoading ? <Spinner size="small" />
+    :
+      <div className="treepickernode-component-expander" />
+    }
```
patrick-ivers: This formatting seems a little weird.
I think any of the following would be better:

```javascript
{isLoading ? <Spinner size="small" />
  : <div className="treepickernode-component-expander" />
}
```

```javascript
{isLoading ?
  <Spinner size="small" />
  : <div className="treepickernode-component-expander" />
}
```

```javascript
{isLoading ?
  <Spinner size="small" />
  :
  <div className="treepickernode-component-expander" />
}
```

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
OndrejCholeva: I don't think that `_.some` would work with `node.path=''` as expected.

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
patrick-ivers: Can we give it a name?

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
patrick-ivers: But `_.some(['']);` is false. Isn't that expected?

test/components/adslotUi/TreePicker/TreePickerNodeExpanderComponentTest.jsx
patrick-ivers: Yes, but each statement in this test is blocking. The `done()` is not being called in a `.then()` or callback as you would expect.

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
mdotwills: `magic_number_Lucky = 25`

src/components/adslotUi/TreePicker/TreePickerNodeExpanderComponent.jsx
mdotwills: 1 or 2 do it for me

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
OndrejCholeva: You're right @patrick-ivers !

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
patrick-ivers: I was thinking more along the lines of `INTERVAL` or `DELAY`.

test/components/adslotUi/TreePicker/TreePickerNodeComponentTest.jsx
patrick-ivers: I'm just wondering why we're using `deep.equal` here and `eql` other places.

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
patrick-ivers: Also, why not use `setTimeout()` instead `setInterval()` + `clearInterval()` ?

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
patrick-ivers: Ahh.. fair enough. Maybe change it to `_.some([node.path.length, node.ancestors.length]);` ?

omgaz: I'll leave it as is.

test/components/adslotUi/TreePicker/TreePickerNodeComponentTest.jsx
omgaz: Adslot UI doesn't have `.eql`, it also doesn have `.to.be`.

src/components/adslotUi/TreePicker/TreePickerNodeComponent.jsx
omgaz: setInterval is non-blocking.

test/components/adslotUi/TreePicker/TreePickerNodeExpanderComponentTest.jsx
omgaz: It tests that clicking the component calls the passed `onClick` function. Passing `done`, as it's a dumb component that isn't affected by the click function. If this click doesn't execute the test will fail because `done` is never called.

patrick-ivers: Ok, but as someone looking at that with no context, it seems a little confusing that you're using `done` as a kind of assertion that `onClick` was called.

patrick-ivers: What happens when the test fails? Does it say "timeout"?

test/components/adslotUi/TreePicker/TreePickerNodeComponentTest.jsx
```diff
@@ -142,71 +143,42 @@ describe('TreePickerNodeComponent', () => {
...
   it('should not show the expander element when the node is not expandable', () => {
     const nonExpandableNode = _.defaults({ isExpandable: false }, cbrNode);
-    const nodes = [];
-    const expandNode = (node) => nodes.push(node);
-    const component = shallow(<TreePickerNodeComponent
-      itemType={itemType} node={nonExpandableNode} expandNode={expandNode}
+    const component = shallow(<TreePickerNode
+      itemType={itemType} node={nonExpandableNode} expandNode={_.noop}
     />);
     const rowElement = component.find({ dts: `${_.kebabCase(itemType)}-${cbrNode.id}` });
```
OndrejCholeva: can we make sure expander is not shown?
Above we are using:
```jsx
expect(cellElements).to.have.length(3); // meta data cell, value cell and include button cell
expect(rowElement.find(TreePickerNodeExpander)).to.have.length(1); // expander cell
```

but here we only `expect(cellElements).to.have.length(3); // meta data cell, value cell and include button cell` which is not enough. can we explicitely `expect(rowElement.find(TreePickerNodeExpander)).to.have.length(0); // no expander cell` ?

omgaz: Sure, definitely.

test/components/adslotUi/TreePicker/TreePickerNodeExpanderComponentTest.jsx
omgaz: I disagree. I think there's plenty of context. 'done' is very explicit that the test finishes. Failing to finish the test is a failed test and a failed expectation, I could wrap the done in a function that `expect(true).to.eql(true)`. It throws exceptions like anything, the error message includes timeout waiting for done to be called. The alternative could be:

```es6
expect(typeof component.prop('onClick')).to.equal('function');
```

if that's any clearer.

patrick-ivers: I just don't understand why we're deviating from our practise of using spies to test function calls. When I see `done()` I immediately think you're testing something asynchronous. Isn't that what it's supposed to be used for? I'd prefer a simple callback that increases a click count over using `done`.

patrick-ivers:
```javascript
let clickCount = 0;
const onClick = () => clickCount++;
...
component.simulate('click');
expect(clickCount).to.equal(1);
```

- Tests onClick was called the correct number of times
- Doesn't require sinon
- Doesn't make you think "where's the async jazz happening?"

omgaz: Ok, I'll add sinon.

patrick-ivers:
Merging... 🚀

PR #516, opened 2017-02-07
Fix: SpinnerButton to postition spinner properly

sonhanguyen:
spinner to not be positioned against elements higher in the DOM than the containing button.


omgaz:
LGTM :)

omgaz:
Merging 🍜

PR #537, opened 2017-03-06
Fix: Broken checkboxes

omgaz:
Checkbox alignment has been off and labels have been bold, which is not the desired behavior. Cleanup lots of `icheckbox` and bootstrap overrides. Also update the form styles to provide more breathing space.

screen shot 2017-03-06 at 17 24 54

screen shot 2017-03-06 at 17 16 56


NamrataTrivedi:
Merging 🚀

PR #545, opened 2017-03-17
Add extra chunk

sonhanguyen:
#Put ```react-datepicker``` to an extra chunk, halving the size

We now have 3 bundles
```adslot-ui-main```: full bundle, the name is preserved to not make this a breaking change.
```adslot-ui-core```: load if you don't need the extra stuff.
```adslot-ui-extra```: load if you need them, right now it's the ```DatePicker```, not dependent on core.

this can enable potential optimization like webpack chunking in ```direct-web``` (so pages that don't use extra stuff can load the core bundle and pages that do use can load the full one)

### DIFF
--- a/src/components/distributionEntry/core.js
+++ b/src/components/distributionEntry/core.js
@@ -1,0 +1,1 @@
+export * from '../core';
--- a/src/components/distributionEntry/index.js
+++ b/src/components/distributionEntry/index.js
@@ -1,0 +1,7 @@
+import _ from 'lodash';
+
+// es6 "import * from" is not used for upseting coverage because of unknown reason
+module.exports = _.assign(
+  require('../core'),
+  require('../extra')
+);
--- a/cfg/dist.js
+++ b/cfg/dist.js
@@ -41,7 +41,9 @@ const config = _.merge(baseConfig, {
     },
   },
   entry: {
-    main: path.join(componentsPath, '/distributionEntry.js'),
+    main: path.join(componentsPath, '/distributionEntry'),
+    core: path.join(componentsPath, '/distributionEntry/core'),
+    extra: path.join(componentsPath, '/distributionEntry/extra'),
   },
   output: {
     path: path.resolve(__dirname, '../dist'),
@@ -57,7 +59,7 @@ const config = _.merge(baseConfig, {
         warnings: false,
       },
       sourceMap: false,
-      include: /main\.js$/,
+      include: /\.js$/,
     }),
     new webpack.optimize.OccurenceOrderPlugin(),
     new webpack.optimize.AggressiveMergingPlugin(),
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: src/components/distributionEntry/core.js [hunk 1]
```diff
@@ -0,0 +1 @@
+export * from '../core';
```
sonhanguyen: webpack does not allow an entry to import another, hence this file

sonhanguyen:
[Split distribution to chunks so that things like DatePicker can be imported from adslot-ui/extras](https://trello.com/c/XdiwbPZH/6146-split-distribution-to-chunks-so-that-things-like-datepicker-can-be-imported-from-adslot-ui-extras)

src/components/distributionEntry/index.js
```diff
@@ -0,0 +1,5 @@
+import _ from 'lodash';
+import * as core from '../core';
+import * as extra from '../extra';
+
+module.exports = _.assign(core, extra);
```
omgaz: Can we not use require?

```
module.exports = {
  core: require('../core'),
  extra: require('../extra'),
}
```

And would this not also fix the issue with webpack you mentioned?

> webpack does not allow an entry to import another, hence this file

src/components/distributionEntry/core.js
lteacher: Did we need the `distributionEntry` actually? I mean, could we instead do the change you did already changing the existing `distributionEntry.js` to 'core.js', then also add the `extra.js` then next up instead, just add the `index.js` right there in the components? Then change the entry to:

```javascript
entry: [
  main: componentsPath,
  core: path.join(componentsPath, '/core'),
  extra: path.join(componentsPath, '/extra'),
]
```

Then you don't need these placeholders?

src/components/distributionEntry/index.js
lteacher: Agree I think if export can't be done then just do the require and remove need for lodash. See my comment about the `webpack does not...` issue

src/components/distributionEntry/core.js
sonhanguyen: what you suggested, which is removing the extra directory, does not solve the problem I'm talking about here. Basically what I meant is the 3 entry files listed in webpack config cannot have dependencies on each other. That's why I had to create another core.js and another extra.js to be entry points, otherwise webpack will complain that the main has dependencies on the other two.

src/components/distributionEntry/index.js
sonhanguyen: @omgaz it's not the same thing, the ```exports``` object should look like this
```
{
  Button,
  Select,
  ...more from core
  DatePicker, // from extra
}
```
not like this
```
{
   core: {
     Button,
     Select,
     ...
   },
   extra: {
     DatePicker,
   },
}
```
... which can be done but seem rather more pervasive a change

vinteo:
Merging...

PR #561, opened 2017-05-16
Fix: Missing Help Component Graphic

omgaz:
Was previously not exporting the help icon.
Also fixed panel titles to work with help icon.


src/components/adslotUi/HelpIconPopoverComponent.jsx
```diff
@@ -1,21 +1,19 @@
 import React, { PropTypes } from 'react';
 import OverlayTrigger from 'react-bootstrap/lib/OverlayTrigger';
 import Popover from 'react-bootstrap/lib/Popover';
-import SvgSymbol from 'components/alexandria/SvgSymbolComponent';
 import expandDts from '../../helpers/expandDtsHelper';
 
 require('styles/adslotUi/HelpIconPopover.scss');
```
sonhanguyen: can we change this to `import` for consistency, or it needs to stay like that for some reason?

sonhanguyen:
LGTM

src/components/adslotUi/HelpIconPopoverComponent.jsx
omgaz: All of these are using `require`

patrick-ivers: @sonhanguyen Not sure why, but all the CSS stuff is imported using `require`. We could look at changing this in a later refactor.

patrick-ivers:
Merging...

PR #564, opened 2017-05-18
Add: Text ellipsis component

lightbringer1991:
Add TextEllipsis component
textellipsis

### DIFF
--- a/src/components/Main.jsx
+++ b/src/components/Main.jsx
@@ -43,6 +43,7 @@ import {
   Tab,
   Tabs,
   Tag,
+  TextEllipsis,
   TileGrid,
   Totals,
   TreePickerSimplePure,
@@ -984,6 +985,12 @@ class AppComponent extends React.Component {
             { label: 'Selected', value: 36.80 },
           ]}
         />
+
+        <h2>TextEllipsis</h2>
+
+        <div style={{ width: '100px' }}>
+          <TextEllipsis text="Lorem ipsum dolor sit amet, consectetuer adipiscing elit." />
+        </div>
 
       </div>
     );
--- a/src/components/adslotUi/TextEllipsisComponent.jsx
+++ b/src/components/adslotUi/TextEllipsisComponent.jsx
@@ -1,0 +1,82 @@
+import _ from 'lodash';
+import React, { PropTypes, Component } from 'react';
+import { OverlayTrigger, Popover } from 'react-bootstrap';
+
+require('styles/adslotUi/TextEllipsis.scss');
+
+class TextEllipsisComponent extends Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      truncated: false,
+    };
+  }
+
+  componentDidMount() {
+    this.setTruncate();
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    return ((this.state.truncated !== nextState.truncated) || (this.props.text !== nextProps.text));
+  }
+
+  componentDidUpdate() {
+    this.setTruncate();
+  }
+
+  setTruncate() {
+    this.setState({
+      truncated: this.container.scrollWidth > this.container.clientWidth,
+    });
+  }
+
+  render() {
+    const {
+      text,
+      popoverProps,
+      overlayTriggerProps,
+    } = this.props;
+    const { truncated } = this.state;
+
+    if (truncated) {
+      const tooltip = (
+        <Popover {...popoverProps}>{text}</Popover>
+      );
+
+      return (
+        <OverlayTrigger {...overlayTriggerProps} overlay={tooltip}>
+          <div className="text-ellipsis-component" ref={(ref) => { this.container = ref; }}>{text}</div>
+        </OverlayTrigger>
+      );
+    }
+
+    return (
+      <div
+        className="text-ellipsis-component"
+        ref={(ref) => { this.container = ref; }}
+      >
+        {text}
+      </div>
+    );
+  }
+}
+
+TextEllipsisComponent.propTypes = {
+  text: PropTypes.string.isRequired,
+  overlayTriggerProps: PropTypes.shape(_.omit(OverlayTrigger.propTypes, ['overlay'])).isRequired,
+  popoverProps: PropTypes.shape(Popover.propTypes).isRequired,
+};
+
+TextEllipsisComponent.defaultProps = {
+  overlayTriggerProps: {
+    trigger: ['focus', 'hover'],
+    placement: 'top',
+  },
+  popoverProps: {
+    id: 'popover',
+    placement: 'top',
+  },
+};
+
+export default TextEllipsisComponent;
--- a/test/components/adslotUi/TextEllipsisComponentTest.jsx
+++ b/test/components/adslotUi/TextEllipsisComponentTest.jsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { mount, shallow } from 'enzyme';
+import { OverlayTrigger } from 'react-bootstrap';
+import TextEllipsisComponent from 'components/adslotUi/TextEllipsisComponent';
+
+describe('TextEllipsisComponent', () => {
+  let divContainer = null;
+
+  beforeEach(() => {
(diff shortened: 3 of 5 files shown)


lightbringer1991:
[Add Publishers and Sites modal needs ellipsis](https://trello.com/c/au5NfDzd/6395-add-publishers-and-sites-modal-needs-ellipsis)

src/components/adslotUi/TextEllipsisComponent.jsx
```diff
@@ -0,0 +1,76 @@
...
+      >
+        {text}
+      </div>
+    );
+  }
+}
+
+TextEllipsisComponent.propTypes = {
+  text: PropTypes.string,
+  overlayTriggerProps: PropTypes.object,
```
omgaz: `overlayTriggerProps: OverlayTrigger.propTypes,`

src/components/adslotUi/TextEllipsisComponent.jsx
```diff
@@ -0,0 +1,76 @@
...
+  tooltipProps: PropTypes.object,
```
omgaz: `tooltipProps: Popover.propTypes,`

omgaz: also, all of these are required.

test/components/adslotUi/TextEllipsisComponentTest.jsx
```diff
@@ -0,0 +1,39 @@
...
+  let divContainer = null;
+
+  beforeEach(() => {
+    divContainer = document.createElement('div');
+    document.body.innerHTML = '';
+    document.body.appendChild(divContainer);
+  });
+
+  it('should render with defaults', () => {
+    const component = mount(<TextEllipsisComponent text="Lorem ipsum dolor sit amet, consectetuer adipiscing elit." />);
```
omgaz: Why are you mounting? We should only need to use shallow rendering.

omgaz: The other's are fine, they allow you to use attachTo

lightbringer1991: `shallow` doesn't render default props. I just want to test the default props. If it's not required, I can remove the test altogether.

patrick-ivers: Really? Are you sure shallow doesn't render default props?

lightbringer1991: @patrick-ivers yea it returns undefined for props that should have default values.

patrick-ivers: Instead of using `mount()`, can you try with `shallowWrapper.instance().props()`?
[http://airbnb.io/enzyme/docs/api/ShallowWrapper/props.html](props())
[http://airbnb.io/enzyme/docs/api/ShallowWrapper/instance.html](instance())

README.md
```diff
@@ -107,3 +107,17 @@ window.Perf = Perf;
...
+- Commits have to be squashed into 1 commit per PR
+- Commit message has to be prefixed with:
+  - 'Fix:'
+  - 'Update:'
+  - 'Breaking:'
+  - 'Docs:'
+  - 'Build:'
+  - 'New:'
+  - 'Upgrade:'
+  - 'Release:'
```
omgaz: Seems like this should be in contributing.md and not readme

lightbringer1991: ok I'll move it over

src/components/adslotUi/TextEllipsisComponent.jsx
```diff
@@ -0,0 +1,77 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      truncated: false,
+    };
+  }
+
+  componentDidMount() {
+    // determine if the text is truncated or not
+    if (this.container.scrollWidth > this.container.clientWidth) {
```
xiaofan2406: it wouldn't change when `this.props.text` is change. So in the case if initially `text` is long and needs to truncated, it sets the state correctly. But when the `text` changed from parent to something really short, it will stay truncated.

src/components/Main.jsx
```diff
@@ -982,6 +983,12 @@ class AppComponent extends React.Component {
           ]}
         />
 
+        <h2>TextEllipsis</h2>
```
xiaofan2406: Can we add demo into its seperate example? and imported in the component?

lightbringer1991: the demo is only in `Main.jsx` though, I'm not sure if we need to split demos into multiple files

src/components/adslotUi/TextEllipsisComponent.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+  render() {
+    const {
+      text,
+      popoverProps,
+      overlayTriggerProps,
+    } = this.props;
+    const { truncated } = this.state;
+
+    const tooltip = (
```
xiaofan2406: this should be inside the if statement below

src/components/Main.jsx
xiaofan2406: the `AppComponent` is way too big. We should seperate new demo into its own component as we touch them. So it doesn't rerender everything on one change.

src/components/adslotUi/TextEllipsisComponent.jsx
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    return ((this.state.truncated !== nextState.truncated) || (this.props.text !== nextProps.text));
+  }
+
+  componentDidUpdate() {
+    if (this.container.scrollWidth > this.container.clientWidth) {
+      this.setTruncate(true);
```
xiaofan2406: this.setTruncate(this.container.scrollWidth > this.container.clientWidth)

src/components/adslotUi/TextEllipsisComponent.jsx
```diff
@@ -0,0 +1,89 @@
...
+    } else {
+      this.setTruncate(false);
+    }
+  }
+
+  setTruncate(isTruncated) {
```
xiaofan2406: this should still be binded even though the current usage isn't causing problem

sonhanguyen: can you get rid of the param and put the block `if (this.container.scrollWidth > this.container.clientWidth) {` so that it's not duplicated?

src/components/Main.jsx
patrick-ivers: @xiaofan2406 Let's do this in a separate PR.

patrick-ivers:
@lightbringer1991 Could you please add a description to this PR.

src/components/Main.jsx
xiaofan2406: @patrick-ivers We should probably put new examples in the examples folder as adding them, so it's less work for the refactor later.

src/components/adslotUi/TextEllipsisComponent.jsx
sonhanguyen: @xiaofan2406 to me prototype method is pointless because you end up creating a bound one anyway. I prefer just put everything in a constructor so that you don't have to create two functions for the same thing. I'm not sure if it makes calling constructor more expensive (may not because JIT), but whatever the cost is we gain back from adding those to prototype at load time.

patrick-ivers:
Merging...

src/components/Main.jsx
patrick-ivers: @xiaofan2406 Actually, we want to possibly go a different way with this...

```
components/
├── MyComponent/
│   ├── component.jsx
│   ├── component.spec.jsx
│   ├── component.example.jsx
│   └── styles.scss
```

sonhanguyen: @omgaz @patrick-ivers we can even get rid of `Main.jsx` and let something like `react-storybook` gather them into one place: https://github.com/storybooks/storybook/blob/master/examples/cra-storybook/.storybook/config.js

PR #558, opened 2017-05-12
Fix: FilePicker input ref got destroyed on file select

xiaofan2406:
#### Changes

- Fix `FilePicker` was destroying `fileInput` ref when a file is selected, due to the conditional rendering.

- Add a seperate `FilePicker` demo, make sure it works as `direct-web` requires

PR #565, opened 2017-05-18
Fix: remove examples tests

xiaofan2406:
As discussed in previous PR https://github.com/Adslot/adslot-ui/pull/558, removing tests for `examples` and exclude them from coverage


vinteo:
Merging...

PR #567, opened 2017-05-26
Breaking: change TextEllipsis component to handle generic content

lightbringer1991:
Previously, TextEllipsis can only accepts simple text as the content for popover

Now it can handle generic contents that are parsed in as children

textellipsis

### DIFF
--- a/src/components/Main.jsx
+++ b/src/components/Main.jsx
@@ -43,7 +43,6 @@ import {
   Tab,
   Tabs,
   Tag,
-  TextEllipsis,
   TileGrid,
   Totals,
   TreePickerSimplePure,
@@ -54,6 +53,7 @@ import {
   ExampleForm,
   ExampleSelect,
   FilePickerDemo,
+  TextEllipsisDemo,
 } from '../examples/exampleEntry';
 
 import {
@@ -986,12 +986,7 @@ class AppComponent extends React.Component {
           ]}
         />
 
-        <h2>TextEllipsis</h2>
-
-        <div style={{ width: '100px' }}>
-          <TextEllipsis text="Lorem ipsum dolor sit amet, consectetuer adipiscing elit." />
-        </div>
-
+        <TextEllipsisDemo />
       </div>
     );
   }
--- a/src/components/adslotUi/TextEllipsisComponent.jsx
+++ b/src/components/adslotUi/TextEllipsisComponent.jsx
@@ -17,23 +17,21 @@ class TextEllipsisComponent extends Component {
     this.setTruncate();
   }
 
-  shouldComponentUpdate(nextProps, nextState) {
-    return ((this.state.truncated !== nextState.truncated) || (this.props.text !== nextProps.text));
-  }
-
   componentDidUpdate() {
     this.setTruncate();
   }
 
   setTruncate() {
-    this.setState({
-      truncated: this.container.scrollWidth > this.container.clientWidth,
-    });
+    const nextTruncateState = this.container.scrollWidth > this.container.clientWidth;
+    if (this.state.truncated !== nextTruncateState) {
+      this.setState({
+        truncated: nextTruncateState,
+      });
+    }
   }
 
   render() {
     const {
-      text,
       popoverProps,
       overlayTriggerProps,
     } = this.props;
@@ -41,12 +39,12 @@ class TextEllipsisComponent extends Component {
 
     if (truncated) {
       const tooltip = (
-        <Popover {...popoverProps}>{text}</Popover>
+        <Popover {...popoverProps}>{this.props.children}</Popover>
       );
 
       return (
         <OverlayTrigger {...overlayTriggerProps} overlay={tooltip}>
-          <div className="text-ellipsis-component" ref={(ref) => { this.container = ref; }}>{text}</div>
+          <div className="text-ellipsis-component" ref={(ref) => { this.container = ref; }}>{this.props.children}</div>
         </OverlayTrigger>
       );
     }
@@ -56,14 +54,14 @@ class TextEllipsisComponent extends Component {
         className="text-ellipsis-component"
         ref={(ref) => { this.container = ref; }}
       >
-        {text}
+        {this.props.children}
       </div>
     );
   }
 }
 
 TextEllipsisComponent.propTypes = {
-  text: PropTypes.string.isRequired,
+  children: PropTypes.node.isRequired,
   overlayTriggerProps: PropTypes.shape(_.omit(OverlayTrigger.propTypes, ['overlay'])).isRequired,
   popoverProps: PropTypes.shape(Popover.propTypes).isRequired,
 };
--- a/test/components/adslotUi/TextEllipsisComponentTest.jsx
+++ b/test/components/adslotUi/TextEllipsisComponentTest.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { mount, shallow } from 'enzyme';
 import { OverlayTrigger } from 'react-bootstrap';
-import TextEllipsisComponent from 'components/adslotUi/TextEllipsisComponent';
+import TextEllipsis from 'components/adslotUi/TextEllipsisComponent';
 
 describe('TextEllipsisComponent', () => {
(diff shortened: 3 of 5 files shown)


lightbringer1991:
[Apply ellipsis and hover over feature to long list item names on pickers everywhere](https://trello.com/c/Qa4KBP2i/6450-apply-ellipsis-and-hover-over-feature-to-long-list-item-names-on-pickers-everywhere)

src/components/adslotUi/TextEllipsisComponent.jsx
```diff
@@ -56,14 +55,14 @@ class TextEllipsisComponent extends Component {
...
-        {text}
+        {this.props.children}
       </div>
     );
   }
 }
 
 TextEllipsisComponent.propTypes = {
-  text: PropTypes.string.isRequired,
+  children: PropTypes.arrayOf(PropTypes.node).isRequired,
```
sonhanguyen: just `PropTypes.node`, don't need `arrayOf`

src/components/adslotUi/TextEllipsisComponent.jsx
```diff
@@ -18,7 +18,7 @@ class TextEllipsisComponent extends Component {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
-    return ((this.state.truncated !== nextState.truncated) || (this.props.text !== nextProps.text));
+    return ((this.state.truncated !== nextState.truncated) || (this.props.children !== nextProps.children));
```
sonhanguyen: `this.props.children` always change (as result of `createElement`) so you don't really need to compare them? @xiaofan2406

src/components/adslotUi/TextEllipsisComponent.jsx
lightbringer1991: @sonhanguyen even in the case of multiple children elements?

sonhanguyen: `node` is for a number of elements, `element` is for single

src/components/adslotUi/TextEllipsisComponent.jsx
xiaofan2406: Yes. This is function will be triggered on state or props change. and in the case of props, it is not necessary to do shallow comparison since they are `node`.  If the width of child element changes, it will not captured here. 
So we should remove props comparison here.
In fact, we should remove this function. Because it will block state change all. @sonhanguyen

src/components/Main.jsx
```diff
@@ -989,9 +989,11 @@ class AppComponent extends React.Component {
         <h2>TextEllipsis</h2>
 
         <div style={{ width: '100px' }}>
-          <TextEllipsis text="Lorem ipsum dolor sit amet, consectetuer adipiscing elit." />
+          <TextEllipsis>
+            Lorem ipsum dolor sit amet, consectetuer adipiscing elit.
+            <span style={{ color: 'red' }}>This is English</span>
+          </TextEllipsis>
```
xiaofan2406: Can we enrich this example so that we can toggle the `children` to see how it works? Because the dynamic behaviours is difficult to capture in tests

xiaofan2406: Can we enrich this example so that we can toggle the `children` to see how it works? Because the dynamic behaviours is difficult to capture in tests

src/components/adslotUi/TextEllipsisComponent.jsx
xiaofan2406: should we now call it `ContentEllipsis` or something?

src/components/adslotUi/TextEllipsisComponent.jsx
sonhanguyen: I think the function is still needed, if width changes and computes to the same `truncate` value, it should not rerender.

xiaofan2406:
btw This isn't breaking. the old behaviour still works

src/components/adslotUi/TextEllipsisComponent.jsx
lightbringer1991: we still encourage users to use text-related content, as if they parse an image into it, it doesn't make sense to do that

lightbringer1991:
@xiaofan2406 old component has `props.text`, while the new one doesn't have it, so it will break the current code

src/components/adslotUi/TextEllipsisComponent.jsx
sonhanguyen: can you add a test case to update children in a way that it won't change truncated? I guess like @xiaofan2406 suggested it won't rerender.

xiaofan2406: I was actually suggesting that `shouldComponentUpdate` should be removed.

lightbringer1991: `shouldComponentUpdate()` needs to be there so that the `componentDidUpdate()` doesn't go into an infinite loop (since it's setting the state, which triggers an update event, which triggers `componentDidUpdate()` again)

best practice is not setting state in `componentDidUpdate()` function, but in this case we need to mount the component to DOM for width computation, and I cannot find another way to do this.

sonhanguyen: instead of `setState()` you can do `this.state.truncated = ..; this.forceUpdate()`, that'll avoid the loop. I think `shouldComponentUpdate` and `forceUpdate` has valid use here.

src/components/demo/TextEllipsisDemo.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { TextEllipsis } from '../distributionEntry';
+
+const TextEllipsisDemo = () => (
```
xiaofan2406: I remember you were changing the `children` in the demo. Did your forget the change `git stash`

sonhanguyen:
LGTM

src/components/demo/TextEllipsisDemo.jsx
lightbringer1991: @xiaofan2406  it's kind of hard to manipulate `children` using just raw html string, I compared `this.props.children` before, and it allows children manipulation as string, but since we removed that bit of code, the inline children editing no longer works.

Also tried `html-to-react` package, but I had some weird error. I'll leave it as a personal pursuit later.

src/components/demo/TextEllipsisDemo.jsx
```diff
@@ -0,0 +1,51 @@
+import moment from 'moment';
+import React, { Component } from 'react';
+import { TextEllipsis, Alert } from '../distributionEntry';
+
+class TextEllipsisDemo extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      childrenData: 'Lorem ipsum dolor sit amet, consectetuer adipiscing elit.' +
```
xiaofan2406: you could actually do 
```
childrenData : ['some text', <span>more text</span>, ....]
```
and render `childrenData` directly like:
```
<TextEllipsis>{this.state.childrenData}</TextEllipsis>
```
But this is minor, we should get the component merged, so its not blocking

lightbringer1991: I put back the editing html so you can edit the children html code in the demo

karma.conf.js
```diff
@@ -38,6 +38,7 @@ module.exports = function configureKarma(config) {
           lines: 100,
           excludes: [
             'src/components/Main.jsx',
+            'src/components/demo/*',
```
patrick-ivers: Why are we creating yet another pattern for example code? Can you put this in `src/examples/...`?

test/components/adslotUi/TextEllipsisComponentTest.jsx
```diff
@@ -13,9 +13,8 @@ describe('TextEllipsisComponent', () => {
   });
 
   it('should render with defaults', () => {
-    const component = shallow(<TextEllipsisComponent text="Sample text" />);
```
patrick-ivers: `<TextEllipsisComponent>` ---> `<TextEllipsis>`

PR #569, opened 2017-05-29
Fix: Add throttle to carousel navigation to prevent sliding backwards

patrick-ivers:
#### Changes
- Add a throttle to the "previous" and "next" actions to prevent the carousel from taking the shortest path to a slide (instead moving in direction it was already going)


patrick-ivers:
[(2) Marketplace landing: interface - collection carousel](https://trello.com/c/D7FojRZB/6353-2-marketplace-landing-interface-collection-carousel)

PR #570, opened 2017-05-29
Update: Modularise Alexandria components

patrick-ivers:
#### Background
The current structure of our component library makes it difficult to add, remove, and update components, as their core elements are spread across separate directory trees for JSX, styles, and tests, e.g.
```
src/
├── components/
│   ├── alexandria/
│   │   ├── CardComponent.jsx
│   …   …
│
├── styles/
│   ├── alexandria/
│   │   ├── Card.scss
…   …   …

test/
├── components/
│   ├── alexandria/
│   │   ├── CardComponentTest.jsx
…   …   …
```

One of the advantages to using Webpack is that we can make components self-contained by keeping all files related to a component together, e.g.
```
src/
├── components/
│   ├── alexandria/
│   │   ├── Card/
│   │   │   ├── index.jsx
│   │   │   ├── index.spec.jsx
│   │   │   └── styles.scss
…   …   …
```

This modular design makes components easier to maintain and reason about.

#### Changes
- Update test and linter configuration to allow for new component structure
- "Modularise" all Alexandria components


src/components/adslotUi/AccordionComponent.jsx
```diff
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import React, { PropTypes } from 'react';
 import Panel from 'components/adslotUi/PanelComponent';
-import Card from 'components/alexandria/CardComponent';
+import Card from 'components/alexandria/Card/component';
```
xiaofan2406: we could have a `index.js` in each folder to control the exports.
`import Card from 'components/alexandria/Card'` looks better in my opinion.

Maybe we could also get rid of the `alexandria` and `adslotUi` directory and move all components to directly under `components`?

sonhanguyen: I think may be using `index.*` rather than `component.*`, since `index` is a widely used naming convention and also makes the `imports` cleaner.

patrick-ivers: I was initially against this because it's node-specific syntactic sugar, and not part of the ES6 spec. But it does look a *lot* nicer.

As for moving all components: Yes, that is the direction we want to head but we're starting with just the Alexandria components to see how it goes.

xiaofan2406: @patrick-ivers almost all bundles follows this convention.

patrick-ivers: Alright, after some discussion with @sonhanguyen , we've decided to change it to index.*

README.md
```diff
@@ -74,7 +74,8 @@ __Adslot UI:__
 
 __Alexandria:__
 
-`yo react-webpack:component alexandria/ComponentName` 
+We don't currently have a command for generating Alexandria components.
```
mdotwills: 'We don't have a command *any more* for generating Alexandria components' 🤣

mdotwills:
LGTM so far, and will look even better with the index.* change you're bringing in 👍

README.md
patrick-ivers: Actually, when I have time, I'm gonna write a script to generate a component with JSX, styles, test stubs. So it really is only temporary. ;)

sonhanguyen:
LGTM

src/components/alexandria/Breadcrumb/index.spec.jsx
```diff
@@ -1,7 +1,8 @@
-import BreadcrumbComponent from 'components/alexandria/BreadcrumbComponent';
-import BreadcrumbNodeComponent from 'components/alexandria/BreadcrumbNodeComponent';
-import { shallow } from 'enzyme';
+/* eslint-disable lodash/prefer-lodash-method */
 import React from 'react';
+import { shallow } from 'enzyme';
+import BreadcrumbNodeComponent from 'components/alexandria/Breadcrumb/Node';
```
lightbringer1991: `BreadcrumbNode` ?

src/components/alexandria/Empty/index.spec.jsx
```diff
@@ -1,7 +1,8 @@
-import { shallow } from 'enzyme';
-import EmptyComponent from 'components/alexandria/EmptyComponent';
-import SvgSymbolCircleComponent from 'components/alexandria/SvgSymbolCircleComponent';
+/* eslint-disable lodash/prefer-lodash-method */
 import React from 'react';
+import { shallow } from 'enzyme';
+import SvgSymbolCircleComponent from 'components/alexandria/SvgSymbol/Circle';
```
lightbringer1991: `SvgSymbolCircle`

src/components/alexandria/Search/index.spec.jsx
```diff
@@ -1,8 +1,9 @@
+/* eslint-disable lodash/prefer-lodash-method */
 import _ from 'lodash';
 import { shallow } from 'enzyme';
 import React from 'react';
-import SearchComponent from 'components/alexandria/SearchComponent';
-import SvgSymbolComponent from 'components/alexandria/SvgSymbolComponent';
+import SvgSymbolComponent from 'components/alexandria/SvgSymbol';
```
lightbringer1991: `SvgSymbol`

src/components/alexandria/Totals/index.jsx
```diff
@@ -1,10 +1,10 @@
 import _ from 'lodash';
-import Grid from 'components/alexandria/GridComponent';
-import GridCell from 'components/alexandria/GridCellComponent';
-import GridRow from 'components/alexandria/GridRowComponent';
+import Grid from 'components/alexandria/Grid';
+import GridCell from 'components/alexandria/Grid/Cell';
+import GridRow from 'components/alexandria/Grid/Row';
```
omgaz: This is where co-location has a massive downfall. One of the big benefits with co-location is that you should treat everything as modular, this clearly shows that is not the case.

If I were to delete Grid, I'd still have clean-up across this Totals component. The dependency of location doesn't stack horizontally, only vertically.

I like the idea of co-location but it's wrought with problems that you have to be considerate of.

Also, should I import Cell and Row in this way I feel this is better maybe

`import Grid, { GridCell, GridRow } = 'components/alexandria/Grid';`

This way it's only one line to delete and prevents orphaned includes

patrick-ivers:
@omgaz I think that's kinda splitting hairs. Besides, our "tests" read more like specs anyway. See: http://joshldavis.com/2013/05/27/difference-between-tdd-and-bdd/

src/components/alexandria/TileGrid/index.jsx
```diff
@@ -1,9 +1,8 @@
 import _ from 'lodash';
 import React, { PropTypes } from 'react';
+import './styles.scss';
```
omgaz: We now have `styles.scss (local) and `styles/` (global) through aliasing of directories. If we're adopting the `index` convention shouldn't our styles here be `index.scss`?

Also, would it make sense to alias `@import 'styles/variable';` something like `@import 'alexandria-styles/variable';` or `@import 'global-styles/variable';`

README.md
omgaz: - currently

I don't think we'll ever be adding one.

src/components/alexandria/Totals/index.jsx
sonhanguyen: I think what @patrick-ivers is doing in this PR is just introduce the structure first, for components like this we can definitely change the `imports` inside `index` to achieve what you suggests.

src/components/alexandria/TileGrid/index.jsx
patrick-ivers: Why? The point of index.jsx is so the module loader treats it as the default export. There's no advantage to changing styles.scss to index.scss, and I feel that doing so just obscures the purpose of the file.

Also, the reason I aliased from `/src` instead of `/src/styles` is because I was thinking we might move the global styles to a "Core" or "App" component, and get rid of `/src/styles`, when all the other components have been "modularised".

README.md
patrick-ivers: Of course we will.

src/components/alexandria/TileGrid/index.jsx
sonhanguyen: the reason we thought it might be a good idea to keep the name as `style.scss` was that I saw libraries using css-modules doing it (I know we're not using css-modules but it should be somewhere down the line). By this we can later introduce some convention for the components to be imported as plain or with style:
```
import Grid from '../Grid' // with style
import { Grid } from '../Grid' // plain
import '../order-sensitive-style.css'
import '../Grid/style.scss' // import style individually
```
besides, I think it looks nicer than `import '../Grid/index.scss'`

src/components/alexandria/Totals/index.jsx
patrick-ivers: @omgaz I structured them in this way because we really shouldn't be using a GridRow or GridCell if we're not also using a Grid; They go together.. or at least they should, like the Card.Container and Card.Content components.

I'm happy to look at a refactor of these in a subsequent PR. I just don't wanna load this one up with different objects. #Iterate #AlwaysBeShipping

src/components/alexandria/TileGrid/index.jsx
patrick-ivers: Oh yea, that too. 😆

PR #571, opened 2017-06-05
New: Add AlertInput component

patrick-ivers:
#### Changes
- Add alertable input component that displays a popover when an alert message is set (and optional colour depending on the alert status)

#### Screenshots
alert-input

### DIFF
--- a/src/components/adslotUi/AlertInput/example.jsx
+++ b/src/components/adslotUi/AlertInput/example.jsx
@@ -1,0 +1,72 @@
+import React, { Component } from 'react';
+import AlertInput from '.';
+
+export default class AlertInputExample extends Component {
+  constructor() {
+    super();
+    this.state = {
+      default: 'Initial value',
+      success: '123.45',
+      info: '120,000,000',
+      warning: '123.45',
+      error: 'patrick.ivers@adslot.com',
+    };
+    this.changeValue = this.changeValue.bind(this);
+  }
+
+  changeValue(key, value) {
+    this.setState({ [key]: value });
+  }
+
+  render() {
+    const style = { marginBottom: '10px' };
+
+    return (
+      <div>
+        <h2>Alert Input</h2>
+        <div style={style}>
+          <AlertInput
+            value={this.state.default}
+            onValueChange={(event) => this.changeValue('default', event.target.value)}
+          />
+        </div>
+        <div style={style}>
+          <AlertInput
+            value={this.state.success}
+            alertStatus="success"
+            alertMessage="This is a success message"
+            onValueChange={(event) => this.changeValue('success', event.target.value)}
+          />
+        </div>
+        <div style={style}>
+          <AlertInput
+            value={this.state.info}
+            suffixAddon="impressions"
+            alertStatus="info"
+            alertMessage="This is an info message"
+            onValueChange={(event) => this.changeValue('info', event.target.value)}
+          />
+        </div>
+        <div style={style}>
+          <AlertInput
+            value={this.state.warning}
+            prefixAddon="$"
+            suffixAddon="CPM"
+            alertStatus="warning"
+            alertMessage="This is a warning message"
+            onValueChange={(event) => this.changeValue('warning', event.target.value)}
+          />
+        </div>
+        <div style={style}>
+          <AlertInput
+            value={this.state.error}
+            prefixAddon="@"
+            alertStatus="error"
+            alertMessage="This is an error message"
+            onValueChange={(event) => this.changeValue('error', event.target.value)}
+          />
+        </div>
+      </div>
+    );
+  }
+}
--- a/src/components/adslotUi/AlertInput/index.jsx
+++ b/src/components/adslotUi/AlertInput/index.jsx
@@ -1,0 +1,63 @@
+import React, { PropTypes } from 'react';
+import OverlayTrigger from 'react-bootstrap/lib/OverlayTrigger';
+import Popover from 'react-bootstrap/lib/Popover';
+import './styles.scss';
+
+export const baseClass = 'alert-input-component';
+
+const AlertInput = ({
+  value,
+  prefixAddon,
+  suffixAddon,
+  alertStatus = '',
+  alertMessage,
+  popoverTrigger,
+  onValueChange,
+}) => {
+  const selectAll = (event) => event.target.select();
+  const baseComponent = (
+    <div className={`${baseClass} ${alertStatus}`}>
+      {prefixAddon ? <span className={`${baseClass}-addon`}>{prefixAddon}</span> : null}
+      <input
+        className={`${baseClass}-input`}
+        type="text"
+        value={value}
+        onClick={selectAll}
+        onChange={onValueChange}
+      />
+      {suffixAddon ? <span className={`${baseClass}-addon`}>{suffixAddon}</span> : null}
+    </div>
+  );
+
+  if (alertMessage) {
+    const popover = (
+      <Popover className={`${baseClass}-popover ${alertStatus}`} id="alert-input-popover">
+        <strong>{alertMessage}</strong>
+      </Popover>
+    );
+
+    return (
+      <OverlayTrigger trigger={popoverTrigger} placement="bottom" overlay={popover}>
+        {baseComponent}
+      </OverlayTrigger>
+    );
+  }
+
+  return baseComponent;
+};
+
+AlertInput.propTypes = {
+  value: PropTypes.string,
+  prefixAddon: PropTypes.node,
+  suffixAddon: PropTypes.node,
+  alertStatus: PropTypes.oneOf(['success', 'info', 'warning', 'error']),
+  alertMessage: PropTypes.string,
+  popoverTrigger: OverlayTrigger.propTypes.trigger,
+  onValueChange: PropTypes.func,
+};
+
+AlertInput.defaultProps = {
+  popoverTrigger: ['hover', 'focus'],
+};
+
+export default AlertInput;
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,4 +5,5 @@ coverage:
   ignore:
     - "src/components/Main.js"
     - "src/components/**/*.spec.*"
+    - "src/components/**/example.jsx"
     - "src/examples"
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -39,6 +39,7 @@ module.exports = function configureKarma(config) {
           excludes: [
             'src/components/Main.jsx',
             'src/components/**/*.spec.*',
+            'src/components/**/example.jsx',
             'src/examples/**/*',
           ],
         },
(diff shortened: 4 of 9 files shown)


sonhanguyen:
can you attach the card with the design?

patrick-ivers:
[(3) Vanguard performance: Deprecate impression, rate and cost picker](https://trello.com/c/bmAap0Kp/6460-(3)-vanguard-performance%3A-deprecate-impression%2C-rate-and-cost-picker)

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+  render() {
+    const {
+      value,
+      alert,
+      prefixAddon,
+      suffixAddon,
+      trigger,
+      placement,
+      onValueChange,
+    } = this.props;
```
xiaofan2406: imo, we need `...rest` here and spread it to `input`, so that we don't have to change the `AlertInput` every time if we need to add native input attribute to it.

We don't have to require `onValueChange`. This will force AlertInput to be a controlled input element. Considering this component actually does a lot, I am not sure re-render on every change is a good idea.

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+      );
+    }
+
+    return baseComponent;
+  }
+}
+
+AlertInput.propTypes = {
+  value: PropTypes.string,
+  alert: PropTypes.shape({
```
xiaofan2406: I think seperate this to `alertType`, `alertMessage` is much more intuitive, and allows you to get rid of `PureComponent` too

patrick-ivers: I'm using `PureComponent` because I need to use `ref` on the input to select all content on click. Has nothing to do with the structure of the props.

src/components/adslotUi/AlertInput/example.jsx
```diff
@@ -0,0 +1,80 @@
...
+export default class AlertInputExample extends Component {
+  constructor() {
+    super();
+    this.state = {
+      a: 'Initial value',
+      b: '123.45',
+      c: '120,000,000',
+      d: '123.45',
+      e: 'patrick.ivers@adslot.com',
+    };
```
sharmajyo: instead of using a, b, c.. for vars can we use may be text, success, info.. or anything else?

src/components/adslotUi/AlertInput/index.jsx
xiaofan2406: `Component` is actually faster in general.

If we don't need the select all text on click behaviour, we could remove the ref and let it take an props like `inputRef` so its parent can fully controll it

src/components/adslotUi/AlertInput/index.jsx
patrick-ivers: Destructuring remaining props from an object with the rest operator is ES7 syntax. So we can't do that. 🤔

src/components/adslotUi/AlertInput/index.jsx
sonhanguyen: @patrick-ivers I think you should have a reason to use `PureComponent` and not just use it because it sounds good. `PureComponent` implies that prop updates are slightly more expensive because of the shallow checking. It's not a bad thing or significant but again you just need a class so it's not a reason to be a pure component.
The `shapeOf` actually looks more intuitive to me, and I don't really see how it has something to do with being pure.

sonhanguyen:
LGTM

src/components/adslotUi/AlertInput/index.jsx
xiaofan2406: We should update our babel config to support it. I think it it essential for a third party library.
Imagine later on we want `onBlur` event on the input, and `onFocus` later on... We had to change this component's API every time

patrick-ivers: The AdslotUI components are highly opinionated. We're not trying to build another Bootstrap here. If we have those requirements in the future, we'll add them.

sonhanguyen: or we can replace babel with typescript, keep the jsx extension and call it a day :)

xiaofan2406: @patrick-ivers Those are not additional functionality we have to build though. Even though we are building AdslotUI based on our requirements. I think it should still be portable and reusable

src/components/adslotUi/AlertInput/index.jsx
patrick-ivers: @xiaofan2406 We *do* need the select all text on click behaviour, that's why it's in there.

@sonhanguyen I'm not using `PureComponent` because it "sounds good", I'm using it (as opposed to `Component`) because it is a *faster* Component class, as it only does shallow props checking. The fact that I even need a class in the first place is because of the `ref` used to select all content in the input on click.

xiaofan2406: `PureComponent` is slower because it to do a shallow comparison on `shouldComponentUpdate`. React internally calls `shouldComponentUpdate` more often than you think. Regardless of Pure or not, flat props helps as we do want to re-render when when primitive values change.

for the select all behaviour: 
```
<input onClick={handleClick} />

handleClick(event) {
    event.target.select()
    if (this.props.onClick) {this.props.onClick(event)}
}
```
So we can convert this to a functional component

src/components/adslotUi/AlertInput/index.jsx
patrick-ivers: @omgaz Thoughts?

lteacher: Dont see the issue with adding the plugin (though maybe the `.babelrc` needs love, move to use `env` instead of specifically 2015 which is way behind). Regarding the spread to input, well I dont mind either but since its library for consumption maybe the features we need today are not the feature someone else needs today?

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+      onValueChange,
+    } = this.props;
+
+    const statusClass = _.get(alert, 'status', '');
+
+    const baseComponent = (
+      <div className={`${baseClass} ${statusClass}`}>
+        {prefixAddon ? <span className={`${baseClass}-addon`}>{prefixAddon}</span> : null}
+        <input
+          className={`${baseClass}-input`}
```
omgaz: Do we get any benefit/leverage for supporting bootstrap .form-control? and input-group-addon.

I feel like we're maturing at a decent rate that the more we can distance ourselves from bootstrap (beyond grid/modal etc…) the better… potentially.

src/components/adslotUi/AlertInput/index.jsx
omgaz: I basically just echoed word-for-word what Patrick had said above. The opinions are to enforce consistency. What's the point of propTypes if we disregard what they set out to achieve.

As much as I see benefits in something like bootstrap to allow complete customisation and configurations, a glove to fit any hand, we're not building for that. If we need an onBlur it should be explicitly added, not implicitly included in a 'behind-closed-doors' deal with some third-party library. If we want to future proof then we add all the property options, but I don't see a benefit of building for every eventuality.

If the problem is having to change the component every time then the solution is to make that process less painful, not to give in to laziness for every imaginary situation.

As for ES7, I see no need to go bleeding-edge and off standards, there's a reason for stable channels and I don't think we're brave enough to veer off. We'll stick with the industry on this one.

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+AlertInput.propTypes = {
+  value: PropTypes.string,
+  alert: PropTypes.shape({
+    status: PropTypes.oneOf(['success', 'info', 'warning', 'error']),
+    message: PropTypes.string.isRequired,
+  }),
+  prefixAddon: PropTypes.node,
+  suffixAddon: PropTypes.node,
+  trigger: OverlayTrigger.propTypes.trigger,
+  placement: OverlayTrigger.propTypes.placement,
```
omgaz: maybe popoverTrigger and popoverPlacement?

src/components/adslotUi/AlertInput/index.jsx
patrick-ivers: @xiaofan2406 Your last comment solved all our disagreements. Good job! 👍

src/components/adslotUi/AlertInput/index.jsx
xiaofan2406: @omgaz Object rest spread is nothing bleeding-edge. A good industry standard would be [`create-react-app`](https://github.com/facebookincubator/create-react-app/blob/master/packages/babel-preset-react-app/index.js) IMO.

patrick-ivers: @xiaofan2406 We can't predict how our component will be used if we allow additional props to be passed through to the input. This makes it difficult to upgrade without breaking other people's implementations. It is better to tightly control how our opinionated components work, and add additional functionality when required. That way we can do it properly (think styles).

src/components/adslotUi/AlertInput/index.jsx
lteacher: If we are going to distance ourselves maybe I can ask if `...Addon` is really the best name?

src/components/adslotUi/AlertInput/index.jsx
lteacher: I changed my mind on the spread, wasn't paying close attention to it. Still don't have any problems with not adding extra props to the input.

sonhanguyen: to recap: look like there are 2 things being discussed here:
- Using newer ES spec: case in point: partial spreading, there are other things such as static props, and more but those two are probably most needed (and safe to add?).
- Passing all react props to 3rd libs:
  + On the one hand: less code to maintain, more flexibility immediately.
  + On the other hand: create implicit couple between 3rd and consuming code, potentially put `adslot-ui` into maintenance issues in the middle.
     This also disputes the need of object spreading for this particular purpose.

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+import Popover from 'react-bootstrap/lib/Popover';
+import './styles.scss';
+
+export const baseClass = 'alert-input-component';
+
+const AlertInput = ({
+  value,
+  prefixAddon,
+  suffixAddon,
+  alertStatus = '',
```
mdotwills: default prop here instead?

patrick-ivers: Actually, I can't do that because the props validation requires one of "success", "info", "warning", or "error" to be passed *if* a value is given.

mdotwills:
Looks awesome btw!

mdotwills:
Merging.. 💯

PR #572, opened 2017-06-07
Fix: AlertInput style tweaks

patrick-ivers:
#### Background
There is too much spacing between the inner elements of the AlertInput component when used in direct-web (i.e. doesn't provide enough room for the input value). There are also some direct-web styles applied directly to popovers that are overriding this component's styles.

#### Changes
- Trim spacing to provide enough breathing room for input value in direct-web


sonhanguyen:
LGTM

omgaz:
LGTMerging :ramen:

PR #573, opened 2017-06-08
Fix: AlertInput focus

patrick-ivers:
#### Background
If a change in component props causes the render function to return a different node, any element within that component that was "focused" will lose focus. This is a problem in direct-web where we validate input as it changes (i.e. as the user types) and show a popover if there is a validation message, which blurs the field and prevents the user from continuing to type in the input (requiring them to refocus on it).

#### Changes
- Change AlertInput to always render with the same root component (OverlayTrigger) so the input element doesn't lose focus if the alertMessage prop changes while the user is typing in the input.
- Add background color to component
- Add defaultValue prop for input so component can be used uncontrolled (for better visual performance)

### DIFF
--- a/src/components/adslotUi/AlertInput/index.jsx
+++ b/src/components/adslotUi/AlertInput/index.jsx
@@ -6,6 +6,7 @@ import './styles.scss';
 export const baseClass = 'alert-input-component';
 
 const AlertInput = ({
+  defaultValue,
   value,
   prefixAddon,
   suffixAddon,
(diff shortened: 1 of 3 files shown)


src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -14,39 +15,38 @@ const AlertInput = ({
...
+  const selectAll = (event) => event.target.select();
+
+  return (
+    <OverlayTrigger trigger={popoverTrigger} placement="bottom" overlay={popover}>
+      <div className={`${baseClass} ${alertStatus}`}>
+        {prefixAddon ? <span className={`${baseClass}-addon`}>{prefixAddon}</span> : null}
+        <input
+          className={`${baseClass}-input`}
+          type="text"
+          defaultValue={defaultValue}
```
sonhanguyen: is having to add this is an example of the down side of not spreading object we discussed in the original PR? I know you don't want to create coupling between the component and some internal bootstrap specifics, but for things like `input`, it's reasonable to expect all props supported by react-dom, such as `defaultValue` and `onBlur` or anything that `react-dom`'s adding in the future.
I'd do `{..._.omit(_.keys(AlertInput.propTypes))}`, but it does look nasty I have to admit.

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -14,39 +15,38 @@ const AlertInput = ({
...
-        className={`${baseClass}-input`}
-        type="text"
-        value={value}
-        onClick={selectAll}
-        onChange={onValueChange}
-      />
-      {suffixAddon ? <span className={`${baseClass}-addon`}>{suffixAddon}</span> : null}
-    </div>
-  );
+  let popover = <div />;
```
patrick-ivers: This is required as `<OverlayTrigger>` doesn't accept null or undefined.

src/components/adslotUi/AlertInput/styles.scss
```diff
@@ -4,6 +4,7 @@
 
 .alert-input-component {
   align-items: center;
+  background-color: $color-background;
```
omgaz: Do we want them to be `color-background` surely `color-well` or something?

patrick-ivers: Ah yea.. `$color-well` is probably better.

src/components/adslotUi/AlertInput/index.jsx
sonhanguyen: does this render an empty popover then?

patrick-ivers: No

vinteo:
Merging...

PR #575, opened 2017-06-18
Fix: AlertInput focus

patrick-ivers:
#### Changes
- Add focus border style to alert input
- Select all input content on focus (instead of on click)

### DIFF
--- a/src/components/adslotUi/AlertInput/index.jsx
+++ b/src/components/adslotUi/AlertInput/index.jsx
@@ -1,51 +1,111 @@
-import React, { PropTypes } from 'react';
-import OverlayTrigger from 'react-bootstrap/lib/OverlayTrigger';
+import React, { Component, PropTypes } from 'react';
+import classnames from 'classnames';
+import Overlay from 'react-bootstrap/lib/Overlay';
 import Popover from 'react-bootstrap/lib/Popover';
 import './styles.scss';
 
 export const baseClass = 'alert-input-component';
 
-const AlertInput = ({
-  defaultValue,
-  value,
-  prefixAddon,
-  suffixAddon,
-  alertStatus = '',
-  alertMessage,
-  popoverTrigger,
-  onValueChange,
-  onBlur,
-}) => {
-  let popover = <div />;
-
-  if (alertMessage) {
-    popover = (
-      <Popover className={`${baseClass}-popover ${alertStatus}`} id="alert-input-popover">
-        <strong>{alertMessage}</strong>
-      </Popover>
-    );
+export default class AlertInput extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isFocused: false,
+      isPopoverVisible: false,
+    };
+    this.getRef = this.getRef.bind(this);
+    this.setRef = this.setRef.bind(this);
+    this.handleMouseEnter = this.handleMouseEnter.bind(this);
+    this.handleMouseLeave = this.handleMouseLeave.bind(this);
+    this.handleInputFocus = this.handleInputFocus.bind(this);
+    this.handleInputBlur = this.handleInputBlur.bind(this);
   }
 
-  const selectAll = (event) => event.target.select();
+  getRef() {
+    return this.root;
+  }
 
-  return (
-    <OverlayTrigger trigger={popoverTrigger} placement="bottom" overlay={popover}>
-      <div className={`${baseClass} ${alertStatus}`}>
+  setRef(root) {
+    this.root = root;
+  }
+
+  handleMouseEnter() {
+    if (this.props.alertMessage) {
+      this.setState({ isPopoverVisible: true });
+    }
+  }
+
+  handleMouseLeave() {
+    this.setState({ isPopoverVisible: false });
+  }
+
+  handleInputFocus(event) {
+    event.target.select();
+    this.setState({
+      isFocused: true,
+      isPopoverVisible: Boolean(this.props.alertMessage),
+    });
+  }
+
+  handleInputBlur(event) {
+    this.setState({
+      isFocused: false,
+      isPopoverVisible: false,
+    });
+
+    if (this.props.onBlur) {
+      this.props.onBlur(event);
+    }
+  }
+
+  render() {
+    const {
+      defaultValue,
+      value,
+      prefixAddon,
+      suffixAddon,
+      alertStatus = '',
+      alertMessage,
+      onValueChange,
+    } = this.props;
+
+    const className = classnames({
+      [baseClass]: true,
+      [alertStatus]: true,
+      'is-focused': this.state.isFocused,
+    });
+
+    return (
+      <div
+        className={className}
+        ref={this.setRef}
+        onMouseEnter={this.handleMouseEnter}
+        onMouseLeave={this.handleMouseLeave}
+      >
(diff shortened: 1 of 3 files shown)


src/components/adslotUi/AlertInput/index.spec.jsx
```diff
@@ -2,94 +2,169 @@
...
+        target: {
+          select: _.noop,
+        },
+      };
+      inputElement.simulate('focus', focusEvent);
+      inputElement.simulate('blur');
+      expect(component.state()).to.eql({
+        isFocused: false,
+        isPopoverVisible: false,
+      });
```
sharmajyo: can we have a var for initial state and use that for various expects?
`const initial_state = { isFocused: false, isPopoverVisible: false }`

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -1,51 +1,104 @@
...
+  componentWillMount() {
+    this.setState({
+      isFocused: false,
+      isPopoverVisible: false,
+    });
+  }
+
+  handleMouseEnter() {
+    if (this.props.alertMessage) {
+      this.setState({ isPopoverVisible: true });
```
pphminions: can you do something like `this.setState({ isPopoverVisible: !!this.props.alertMessage })`
Saw a similar code in line 39 :)

NamrataTrivedi:
Will it respond to keyboard tab and return key events?

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -1,51 +1,104 @@
...
+  }
 
-  if (alertMessage) {
-    popover = (
-      <Popover className={`${baseClass}-popover ${alertStatus}`} id="alert-input-popover">
-        <strong>{alertMessage}</strong>
-      </Popover>
-    );
+  componentWillMount() {
+    this.setState({
```
xiaofan2406: maybe move this to constructor and write as initial state?

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -1,51 +1,104 @@
...
+      suffixAddon,
+      alertStatus = '',
+      alertMessage,
+      onValueChange,
+    } = this.props;
+
+    const className = classnames({
+      [baseClass]: true,
+      [alertStatus]: true,
+      'is-focused': this.state.isFocused,
```
xiaofan2406: Can this be done with just CSS `:focused`?

patrick-ivers: No.

src/components/adslotUi/AlertInput/index.jsx
patrick-ivers: I didn't do it here because I wanted to avoid an unnecessary function call.

patrick-ivers:
@NamrataTrivedi It responds to Tab, but not Enter/Return. I wanted to keep it as close to spreadsheet functionality as possible, given it will mainly be used in the grids. Enter/Return in most spreadsheets moves the focus to the cell beneath, which is something not scoped for the current feature.

NamrataTrivedi:
Will be good to add tests for simulating tab key event, importantly tabbing out of input field. Users might do this most often.

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -1,51 +1,101 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      isFocused: false,
+      isPopoverVisible: false,
+    };
+    this.handleMouseEnter = this.handleMouseEnter.bind(this);
+    this.handleMouseLeave = this.handleMouseLeave.bind(this);
+    this.handleInputFocus = this.handleInputFocus.bind(this);
+    this.handleInputBlur = this.handleInputBlur.bind(this);
```
omgaz: We've adopted this pattern before, what do you think:

```
    for (const methodName of [
      'handleMouseEnter',
      'handleMouseLeave',
      'handleInputFocus',
      'handleInputBlur',
    ]) { this[methodName] = this[methodName].bind(this); }
```

patrick-ivers: I saw that being done in the Main.jsx file, where it makes sense given it's such a huge file. I think it's probably overkill here. Plus it adds two extra lines. What do you think?

patrick-ivers: @omgaz What about transform-class-properties babel plugin? @lteacher suggested it.

omgaz: Sounds good, your call :)

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -1,51 +1,101 @@
...
+
+  handleMouseLeave() {
+    this.setState({ isPopoverVisible: false });
+  }
+
+  handleInputFocus(event) {
+    event.target.select();
+    this.setState({
+      isFocused: true,
+      isPopoverVisible: !!this.props.alertMessage,
```
omgaz: I'm not a fan of the double-bang, also if we switch to typescript are we even able to do that?
 Prefer `Boolean(this.props.alertMessage)`, thoughts?

src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -1,51 +1,101 @@
...
+      'is-focused': this.state.isFocused,
+    });
 
-  return (
-    <OverlayTrigger trigger={popoverTrigger} placement="bottom" overlay={popover}>
-      <div className={`${baseClass} ${alertStatus}`}>
+    return (
+      <div
+        className={className}
+        ref={(root) => { this.root = root; }}
```
omgaz: See that article I linked this morning. Not sure if it makes a diff to us, but this will create a new anonymous function on each render. I've seen us do this quite a lot, but should we start moving away from inline funcs in the render? Thoughts?

src/components/adslotUi/AlertInput/index.jsx
omgaz: Is this because of the element itself not being the focus but is just wrapping the focused element?

src/components/adslotUi/AlertInput/index.jsx
xiaofan2406: @omgaz It only makes a difference when that function is passed on to child component. `ref` functions make no difference in this regards. 
But I think we should still avoid inline ref functions to avoid the `null` value being assigned to `ref` when component is first constructed.

patrick-ivers: @xiaofan2406 Why does it only make a difference when the function is passed on to child component? Each time the `AlertInput` render function is called, it will recreate the anonymous function. Will it not?

src/components/adslotUi/AlertInput/index.jsx
patrick-ivers: Yea, it's because the parent div needs to be aware of the focus state of its child input, which can be achieved with CSS alone.

src/components/adslotUi/AlertInput/index.jsx
xiaofan2406: People advise against inline functions is mainly because it can cause unnecessary rendering for each time a new identify is created but the props are actually the same.
For `ref`, it is never passed to child as props, or used as re-render evaluation.

vinteo:
Merging...

PR #577, opened 2017-06-22
Update: Always display Group label in TreePickerGrid component

lightbringer1991:
always display group in TreePickerGrid, if there is no group specified, display Empty element

### DIFF
--- a/src/components/adslotUi/TreePicker/TreePickerGridComponent.jsx
+++ b/src/components/adslotUi/TreePicker/TreePickerGridComponent.jsx
@@ -32,13 +32,11 @@ const TreePickerGridComponent = ({
     <Grid>
       {_.map(nodesByGroupLabel, (groupedNodes, label) =>
         <div className="treepickergrid-component-group" key={_.kebabCase(label)}>
-          {_.size(nodesByGroupLabel) > 1 ?
-            <div className="treepickergrid-component-group-label">
-              <GridRow dts={`group-label-${_.kebabCase(label)}`}>
-                {label}
-              </GridRow>
-            </div> :
-            null}
+          <div className="treepickergrid-component-group-label">
+            <GridRow dts={`group-label-${_.kebabCase(label)}`}>
+              {label}
+            </GridRow>
+          </div>
           {_.map(groupedNodes, (node) =>
             <TreePickerNodeFast
               key={node.id}
--- a/test/components/adslotUi/TreePicker/TreePickerGridComponentTest.jsx
+++ b/test/components/adslotUi/TreePicker/TreePickerGridComponentTest.jsx
@@ -36,10 +36,10 @@ describe('TreePickerGridComponent', () => {
 
     const groupElement = gridElement.find('.treepickergrid-component-group');
     expect(groupElement).to.have.length(1);
-    expect(groupElement.children()).to.have.length(2); // Two nodes
+    expect(groupElement.children()).to.have.length(3); // Group label and two nodes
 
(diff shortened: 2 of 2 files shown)


lightbringer1991:
[Campaign Management: Targeting Picker does not show heading labels when only one segment context is displayed (Audience Tab)](https://trello.com/c/roCt2rFN/6535-campaign-management-targeting-picker-does-not-show-heading-labels-when-only-one-segment-context-is-displayed-audience-tab)

src/components/adslotUi/TreePicker/TreePickerGridComponent.jsx
```diff
@@ -32,7 +32,7 @@ const TreePickerGridComponent = ({
     <Grid>
       {_.map(nodesByGroupLabel, (groupedNodes, label) =>
         <div className="treepickergrid-component-group" key={_.kebabCase(label)}>
-          {_.size(nodesByGroupLabel) > 1 ?
+          {_.size(nodesByGroupLabel) >= 1 ?
```
lightbringer1991: I'm not sure if we really want this to be displayed everywhere. Maybe adding a flag as a prop would be nicer in this case. What do you think @omgaz @patrick-ivers @sonhanguyen

mdotwills: You raise a good point. In the case we just care about the size >= 1, we could simplify further as

    _.size(nodesByGroupLabel) ?

src/components/adslotUi/TreePicker/TreePickerGridComponent.jsx
```diff
@@ -32,13 +32,11 @@ const TreePickerGridComponent = ({
...
       {_.map(nodesByGroupLabel, (groupedNodes, label) =>
         <div className="treepickergrid-component-group" key={_.kebabCase(label)}>
-          {_.size(nodesByGroupLabel) > 1 ?
-            <div className="treepickergrid-component-group-label">
-              <GridRow dts={`group-label-${_.kebabCase(label)}`}>
-                {label}
-              </GridRow>
-            </div> :
-            null}
+          <div className="treepickergrid-component-group-label">
```
mdotwills: nice change. Just since we are always now showing the label, it should always be provided too, otherwise we'll have a blank header.

- We could hide this bank header if that's the case, or
- We make it a required prop. The label looks like it comes from the the `nodes` and `groupFormatter`, is there a way to enforce that the combination of these props actually have a `label`, or is the ` nodes: PropTypes.arrayOf(TreePickerPropTypes.node)` propType declaration sufficient?

lightbringer1991: if the groupFormatter isn't provided, an empty element will be shown, and no data is displayed

lightbringer1991: that's why I think a flag that decides if we want to show the labels or not may be better here, as in some cases we don't want to show label if there is no group

mdotwills: @omgaz thoughts? We could have a prop as you say that dictates this behaviour, and we can set the defaultProps behaviour so something reasonable like 'showGroupHeaders = false`

omgaz: I think we should be consistent and opinionated and try to avoid too many behavioral branches in our components unless necessary. I also like the idea of being optimistic, we should assume we will have data.

patrick-ivers:
Merging...

PR #578, opened 2017-06-30
Update: ListPicker: allow linkButtons to accept nodes

lightbringer1991:


lightbringer1991:
[(3) Create product: tabs & paging](https://trello.com/c/LPb3SmDZ/6526-3-create-product-tabs-paging)

src/components/adslotUi/ListPickerComponent.jsx
```diff
@@ -135,11 +135,11 @@ class ListPickerComponent extends React.Component {
         <Modal.Footer>
           {_.isEmpty(props.linkButtons) ? null :
             <div className="pull-left">
-              {_.map(props.linkButtons, (linkButton) =>
+              {_.map(props.linkButtons, (linkButton) => (React.isValidElement(linkButton) ? linkButton : (
```
omgaz: You don't need to do isValidElement if you use the propType `PropTypes.element`

src/components/adslotUi/ListPickerComponent.jsx
```diff
@@ -163,6 +163,14 @@ const itemProps = PropTypes.shape({
   id: PropTypes.number.isRequired,
 });
 
+const linkButtonsProps = PropTypes.arrayOf(PropTypes.oneOfType([
+  PropTypes.shape({
+    label: PropTypes.string.isRequired,
+    href: PropTypes.string.isRequired,
+  }),
+  PropTypes.node,
```
omgaz: see above

src/components/adslotUi/ListPickerComponent.jsx
lightbringer1991: hmm I actually want it to be either a node, or an object with predefined attributes. I'll change the code to reflect that

src/components/adslotUi/ListPickerComponent.jsx
```diff
@@ -12,6 +12,11 @@ import SvgSymbol from 'components/alexandria/SvgSymbol';
 
 require('styles/adslotUi/ListPicker.scss');
 
+const isSubset = (array, subArray) =>
```
lightbringer1991: I'm not sure if lodash already have a function to check if an array is a subset of another array. I'm happy to switch if it does

tried `_.includes` but it seems to only check for 1 element

omgaz:
LGTM

src/components/adslotUi/ListPickerComponent.jsx
omgaz: Maybe _(subarray).without(array).isEmpty()?

lteacher:
merging

PR #579, opened 2017-07-03
New: add Textarea component to support word count

xiaofan2406:
New create product details are in React and we don't have a component to support textarea word countdown.

### DIFF
--- a/src/components/Main.jsx
+++ b/src/components/Main.jsx
@@ -62,6 +62,7 @@ import {
 } from '../examples/redux/reducers';
 
 import AlertInputExample from './adslotUi/AlertInput/example';
+import TextareaExample from './adslotUi/Textarea/example';
 
 require('styles/App.scss');
 
@@ -465,6 +466,8 @@ class AppComponent extends React.Component {
         </div>
 
         <AlertInputExample />
+
+        <TextareaExample />
 
         <FilePickerDemo />
 
--- a/src/components/adslotUi/Textarea/example.jsx
+++ b/src/components/adslotUi/Textarea/example.jsx
@@ -1,0 +1,35 @@
+import React, { Component } from 'react';
+import { Button } from 'react-bootstrap';
+
+import Textarea from '.';
+
+class TextareaExample extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      maxLength: 5,
+    };
+
+    this.toggleMaxLength = this.toggleMaxLength.bind(this);
+  }
+
+  toggleMaxLength() {
+    this.setState({
+      maxLength: this.state.maxLength === 5 ? null : 5,
+    });
+  }
+
+  render() {
+    return (
+      <div>
+        <h2>Textarea</h2>
+        <Textarea maxLength={this.state.maxLength} statusClass="pull-right" />
+        <br />
+        <Button className="btn-inverse" bsStyle="primary" onClick={this.toggleMaxLength}>Toggle max length</Button>
+      </div>
+    );
+  }
+}
+
+export default TextareaExample;
--- a/src/components/adslotUi/Textarea/index.jsx
+++ b/src/components/adslotUi/Textarea/index.jsx
@@ -1,0 +1,52 @@
+import _ from 'lodash';
+import React, { PropTypes } from 'react';
+import classnames from 'classnames';
+
+class Textarea extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      charCountRemaining: this.props.maxLength,
+    };
+
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(event) {
+    this.setState({
+      charCountRemaining: this.props.maxLength - event.target.value.length,
+    });
+
+    if (this.props.onChange) {
+      this.props.onChange(event);
+    }
+  }
+
+  render() {
+    const { maxLength, statusClass } = this.props;
+    const restProps = _.omit(this.props, ['statusClass']);
+    const classNames = classnames('form-control', restProps.className);
+
+    return _.isNil(maxLength) ? (
+      <textarea {...restProps} className={classNames} />
+    ) : (
+      <div>
+        <textarea {...restProps} className={classNames} onChange={this.handleChange} />
+        <span className={statusClass}>{this.state.charCountRemaining} characters remaining</span>
+      </div>
+    );
+  }
+}
+
+Textarea.propTypes = {
+  maxLength: PropTypes.number,
+  statusClass: PropTypes.string,
+  onChange: PropTypes.func,
+};
+
+Textarea.defaultProps = {
+  statusClass: '',
+};
+
+export default Textarea;
--- a/src/components/adslotUi/Textarea/index.spec.jsx
+++ b/src/components/adslotUi/Textarea/index.spec.jsx
@@ -1,0 +1,55 @@
+/* eslint-disable lodash/prefer-lodash-method */
+import React from 'react';
+import { shallow } from 'enzyme';
+import sinon from 'sinon';
+import Textarea from '.';
+
+describe('Textarea', () => {
+  it('should render a textarea element', () => {
+    const component = shallow(<Textarea />);
+    const textarea = component.find('textarea');
+    expect(textarea).to.have.length(1);
+  });
+
+  it('should render a countdown span when maxLength is specified', () => {
+    const component = shallow(<Textarea maxLength={120} statusClass="someclass" />);
+    expect(component.find('span').text()).to.equal('120 characters remaining');
+    expect(component.find('span').hasClass('someclass')).to.equal(true);
+  });
+
+  it('should give additional className to the textarea', () => {
+    const component = shallow(<Textarea maxLength={120} className="someclass" />);
+    expect(component.find('textarea').hasClass('form-control someclass')).to.equal(true);
+  });
+
+  it('should trigger onChange handler', () => {
+    const props = { onChange: sinon.spy(), maxLength: 120 };
+    const component = shallow(<Textarea {...props} />);
+    component.find('textarea').simulate('change', { target: { value: 'abcde' } });
+    expect(props.onChange.callCount).to.equal(1);
+  });
+
+  it('should update the remaining character count on textarea change', () => {
+    const setStateSpy = sinon.spy(Textarea.prototype, 'setState');
+    const component = shallow(<Textarea maxLength={120} />);
+    component.find('textarea').simulate('change', { target: { value: 'abcde' } });
+    expect(setStateSpy.callCount).to.equal(1);
+    expect(component.state('charCountRemaining')).to.equal(115);
+    setStateSpy.restore();
+  });
+
+  it('should not update the remaining character count when maxLength is not specified', () => {
+    const setStateSpy = sinon.spy(Textarea.prototype, 'setState');
+    const component = shallow(<Textarea />);
+    component.find('textarea').simulate('change', { target: { value: 'abcde' } });
+    expect(setStateSpy.callCount).to.equal(0);
+    setStateSpy.restore();
+  });
+
+  it('should pass on additional props to textarea element', () => {
+    const component = shallow(<Textarea placeholder="hello" maxLength={120} />);
+    const textarea = component.find('textarea');
+    expect(textarea.prop('maxLength')).to.equal(120);
+    expect(textarea.prop('placeholder')).to.equal('hello');
+  });
+});


src/components/adslotUi/Textarea/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+  render() {
+    const { children, maxLength, statusClass } = this.props;
+
+    const restProps = {};
+    _.forEach(this.props, (prop, key) => {
+      if (key !== 'children') {
+        restProps[key] = this.props[key];
+      }
+    });
```
xiaofan2406: I mentioned this pattern in @patrick-ivers `AlertInput` PR, not sure how everyone feels about for this component.

As right now we dont know whats needed for the textarea, it is difficult to specify props beforehand.

sonhanguyen: can you just use `omit`?

lightbringer1991: `_.omit(this.props, ['children'])`

src/components/adslotUi/Textarea/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+    const classNames = classnames('form-control', restProps.className);
+
+    return _.isNil(maxLength) ? (
+      <textarea {...restProps} className={classNames} >{children}</textarea>
+    ) : (
+      <div>
+        <textarea {...restProps} className={classNames} onChange={this.handleChange}>{children}</textarea>
```
sonhanguyen: is it possible to throttle here? (and have props for throttling)

src/components/adslotUi/Textarea/index.jsx
xiaofan2406: dem lodash. will change over

src/components/adslotUi/Textarea/index.jsx
xiaofan2406: I was thinking about add a throttle too but couldn't decide the limit. 
But I dont think having a prop to change it is necessary.

src/components/Main.jsx
```diff
@@ -62,6 +62,7 @@ import {
 } from '../examples/redux/reducers';
 
 import AlertInputExample from './adslotUi/AlertInput/example';
+import TextareaExample from './adslotUi/Textarea/example';
```
mdotwills: `TextAreaExample`. If `Input` is capitalised on the line above, then `Area` should too be capitalised

src/components/adslotUi/Textarea/example.jsx
```diff
@@ -0,0 +1,37 @@
+
+import React, { Component } from 'react';
+import { Button } from 'react-bootstrap';
+
+import Textarea from './';
```
mdotwills: `TextArea`

src/components/adslotUi/Textarea/index.jsx
sonhanguyen: just give it a default, any value is better than unthrottled.

src/components/adslotUi/Textarea/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+import React, { PropTypes } from 'react';
+import classnames from 'classnames';
+
+
+class Textarea extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      textRemaining: this.props.maxLength,
```
mdotwills: `charCountRemaining`

src/components/Main.jsx
xiaofan2406: I always thought textarea is one word...

src/components/adslotUi/Textarea/index.jsx
mdotwills: i don't think it needs to be throttled, we're responding to button press onChange events, like we're about to do for controlled inputs (and it makes no sense to throttle that)

src/components/adslotUi/Textarea/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+      <div>
+        <textarea {...restProps} className={classNames} onChange={this.handleChange}>{children}</textarea>
+        <span className={statusClass}>{this.state.textRemaining} characters remaining</span>
+      </div>
+    );
+  }
+}
+
+Textarea.propTypes = {
+  children: PropTypes.node,
```
lightbringer1991: hmm for a textarea, shouldn't the children be just a string? 

Ref https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

src/components/adslotUi/Textarea/index.jsx
sonhanguyen: I don't get why it makes no sense. If you think rerender every keystroke is bad then do throttle.
did you think this was `onBlur`?

xiaofan2406: It is not a controlled component actually.

src/components/adslotUi/Textarea/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+  maxLength: PropTypes.number,
+  statusClass: PropTypes.string,
+  onChange: PropTypes.func,
+};
+
+Textarea.defaultProps = {
+  children: null,
+  maxLength: null,
+  statusClass: '',
+  onChange: null,
```
lightbringer1991: `_.noop` I think, so it is still a function

xiaofan2406: but I want it to be a falsy value so that it is not called at all

src/components/adslotUi/Textarea/index.jsx
sonhanguyen: then how come this wraps `children`, do you mean to use `defaultValue` instead?

mdotwills: sorry you're right @xiaofan2406 it is not a controlled component, it would controlled if the `value` was set to the prop change.

Throttle is not required, we want to update the remaining char count on keypress as this the natural update mechanism that the user is expecting w.r.t. the remaining char count update, and the burden to the DOM is negligible.

xiaofan2406: @sonhanguyen It is a `textarea` it doesn't has `value` or `defaultValue`
https://www.w3schools.com/tags/tag_textarea.asp

sonhanguyen: https://facebook.github.io/react/docs/uncontrolled-components.html
it's a React thing

xiaofan2406: @sonhanguyen oh didnt know that react adds `value` and `defaultValue` to `textarea`. I can get rid of this all together

src/components/Main.jsx
mdotwills: Yep

omgaz:
Do we not want to have the component dictate status class based on whether or not it's over the maxLength?

src/components/adslotUi/Textarea/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+    const { maxLength, statusClass } = this.props;
+    const restProps = _.omit(this.props, ['statusClass']);
+    const classNames = classnames('form-control', restProps.className);
+
+    return _.isNil(maxLength) ? (
+      <textarea {...restProps} className={classNames} />
+    ) : (
+      <div>
+        <textarea {...restProps} className={classNames} onChange={this.handleChange} />
+        <span className={statusClass}>{this.state.charCountRemaining} characters remaining</span>
```
omgaz: not sure if `characters` is very conversational. Is it even needed?

xiaofan2406: It is in our current design pdf. we could make this a prop later when a different design comes up. what do you think? @omgaz

xiaofan2406:
@omgaz the `maxLength` prop on `textarea` would prevent any more characters being entered. So if specified, it will simply not update its content anymore

sonhanguyen:
Actually I forgot: do we need `defaultValue` and `value` listed in `propTypes`?

xiaofan2406:
@sonhanguyen no. Just as how you would use a normal `textarea` they are not required, and you dont have to provide either of them

sonhanguyen:
Not required is different than not having proptype isn't it?

xiaofan2406:
@sonhanguyen thats why they should be left off. otherwise we would have to include all the props that native textarea support and put them into the propTyps (required or not).

sonhanguyen:
How come we list `onChange` and earlier `children` but not `value`, what's the reasoning for treating these props differently?

the idea is that we should know what props are supported. Anyone who uses this component should be able to look at `propTypes` and know exactly what is supported. It's should not be "this component is called textarea something so I **guess** it's used the same way" and even have to read Dom API docs because the symbol `value` is nowhere to be found in the code.

Ideally if we are commiting into making this compatible with the dom element we can use https://github.com/tyv/pick-react-known-prop BUT I'm just suggesting to make explicit at least the fact that this component takes `defaultValue` or `value`

sonhanguyen:
nvm I guess we can frame it "this component is intended to be used without `defaultValue`"

xiaofan2406:
@sonhanguyen `children` was there because I didnt know React adds `value` to `textarea`.
`onChange` is here, because it is used internally, and I want any additional `onChange` is still called as usual. It is in one of the unit test.
This is a common pattern for creating components that just add one or two behaviours to the native dom element. You certainly can copy all the textarea attributes and put them here, as all of them are supported. But that just felt a little bit off for me. 
Just like when you creating a higher order component, you don't care what the actual props are, you just passing them along.

It is not intended to be used without `defautValue`. It can be used with or without it. As equivalent to you can write an html `<textarea></textarea>` or `<textarea>content</textarea>`

src/components/adslotUi/Textarea/example.jsx
patrick-ivers: Change to `import Textarea from '.';`

src/components/adslotUi/Textarea/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+Textarea.propTypes = {
+  maxLength: PropTypes.number,
+  statusClass: PropTypes.string,
+  onChange: PropTypes.func,
+};
+
+Textarea.defaultProps = {
+  maxLength: null,
+  statusClass: '',
+  onChange: null,
```
patrick-ivers: Why are we setting default to null? Can't we just leave as undefined?

src/components/adslotUi/Textarea/index.spec.jsx
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable lodash/prefer-lodash-method */
+import React from 'react';
+import { shallow } from 'enzyme';
+import sinon from 'sinon';
+import Textarea from './';
```
patrick-ivers: `import Textarea from '.';`

src/components/adslotUi/Textarea/index.spec.jsx
```diff
@@ -0,0 +1,55 @@
...
+    expect(component.find('span').text()).to.equal('120 characters remaining');
+    expect(component.find('span').hasClass('someclass')).to.equal(true);
+  });
+
+  it('should give additional className to the textarea', () => {
+    const component = shallow(<Textarea maxLength={120} className="someclass" />);
+    expect(component.find('textarea').hasClass('form-control someclass')).to.equal(true);
+  });
+
+  it('shoud trigger onChange handler', () => {
```
patrick-ivers: *should

src/components/adslotUi/Textarea/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+  render() {
+    const { maxLength, statusClass } = this.props;
+    const restProps = _.omit(this.props, ['statusClass']);
+    const classNames = classnames('form-control', restProps.className);
+
+    return _.isNil(maxLength) ? (
+      <textarea {...restProps} className={classNames} />
+    ) : (
+      <div>
+        <textarea {...restProps} className={classNames} onChange={this.handleChange} />
```
patrick-ivers: `{...restProps}` --> `maxlength={maxLength}`

src/components/adslotUi/Textarea/index.jsx
omgaz: That's fine.

src/components/adslotUi/Textarea/index.jsx
xiaofan2406: null is a falsy value, undefined can potentially invoke runtime error.
 But i guess they are not expecting object in this case, I will change

src/components/adslotUi/Textarea/index.jsx
xiaofan2406: hum... I intend to pass on all the props not just `maxLength`

patrick-ivers: But there are no other props.

xiaofan2406: So that you can do something like `<Textarea value={somevalue} onFocus={somefunc} />` as if it is a native dom textarea

omgaz:
Aside from Patrick's comments, and a few design related issues that may need to be revised with Pedros, this PR looks good.

src/components/adslotUi/Textarea/index.jsx
patrick-ivers: How does undefined invoke a runtime error?

src/components/adslotUi/Textarea/index.jsx
omgaz: And <textarea> is HTML and not JSX at this point so we should be explicit about what we pass through to protect against side effects and un-supported attributes being appended without cause.

xiaofan2406: But react camel case all props.

There are other props (all textarea attributes). It is just that this component doenst care about them.

src/components/Main.jsx
patrick-ivers: Are 'textarea' and 'checkbox' really single words though? I know the HTML tags are written all lowercase, but I wouldn't take that as meaning they're single words. To me "Textarea" looks weirder than "TextArea". 😛

src/components/adslotUi/Textarea/index.jsx
patrick-ivers: @xiaofan2406 But those props are not defined on the propTypes, so you can't just pass them through. We had this conversation a few weeks ago: It's better to be explicit about what props we're receiving and passing through, than to just pass everything. We're not building another bootstrap; our requirements of these components are very specific.

xiaofan2406: but again, you cant know how user intends to use this Textarea as it should support all native textarea attributes. Like I discussed with @sonhanguyen , I don't mind adding all the textarea attributes in. (Linter wont be happy though). But it looks really off for me, as this component doesnt care about those props at all. Just as any higher order component, it simply add some functionality to a component, and pass rest of the props along.

src/components/adslotUi/Textarea/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+Textarea.propTypes = {
+  maxLength: PropTypes.number,
+  statusClass: PropTypes.string,
+  onChange: PropTypes.func,
+};
+
+Textarea.defaultProps = {
+  maxLength: undefined,
+  statusClass: '',
+  onChange: undefined,
```
patrick-ivers: So now that they're undefined, you don't need to specify them as defaultProps, yea?

src/components/adslotUi/Textarea/index.jsx
patrick-ivers: Hrmm.. I see what you're saying. It really is only a super lightweight wrapper around a textarea. @omgaz Thoughts?

xiaofan2406:
@sonhanguyen in your `asModal`, are you specifying all the propTypes from `ProductForm`?

`<textarea></textarea> and <textarea>content</textarea>` was just to demonstrate that children are optional in html, as in this component `value` is optional. We don't know for sure that this is used with `value` and `defaultValue`.

src/components/adslotUi/Textarea/index.jsx
xiaofan2406: will check. I thought we had that rule non-required-props-must-have-a-default-prop-value enabled.

sonhanguyen:
no and my point is that this is not a higher order component so let's not focus on that.
I'm not objecting you passing everything down, just suggesting you to add **one** prop type. Note that we already have onChange and maxLength, whether or not we intercept them, the naming means to adhere to dom API, not some contract we invented, therefore by your logic those shouldn't be there either.

src/components/adslotUi/Textarea/index.jsx
patrick-ivers: @xiaofan2406 After speaking with @omgaz and @sonhanguyen , I'm happy for you to use this pattern as this is not a complex component, just a wrapper.

src/components/adslotUi/Textarea/index.jsx
patrick-ivers: ahh.. maybe we do. i'll have a look

xiaofan2406:
@sonhanguyen You are right. According to what I said, I shouldn't put `onChange` or 'maxLength' in `propTypes`. They are simply there to make linter happy. I am happy to add `eslint-disable-line`.

sonhanguyen:
fair enough, I accept the linter argument.

omgaz:
Some good discussions, @sonhanguyen any objections to merging this?

sonhanguyen:
LGTM, already approved

mdotwills:
Merging.. 🚀

PR #581, opened 2017-07-10
Update: add `displayGroupHeader` props to always display group header when needed

lightbringer1991:
Sometimes we don't want to display group header because there is none (e.g. list of Countries, which is obvious enough to not have a header). More examples in the bug card

Screenshots:
- Default (`displayGroupHeader = true`):

  - with only 1 group
screen shot 2017-07-10 at 3 55 48 pm

   - with multiple groups
screen shot 2017-07-10 at 3 57 18 pm

- `displayGroupHeader = false`
screen shot 2017-07-10 at 3 50 46 pm

### DIFF
--- a/src/components/Main.jsx
+++ b/src/components/Main.jsx
@@ -641,6 +641,7 @@ class AppComponent extends React.Component {
           includeNode={_.noop}
           removeNode={_.noop}
           nodeRenderer={nodeRenderer}
+          displayGroupHeader
         />
 
         <h2>ListPicker</h2>
--- a/src/components/adslotUi/TreePicker/TreePickerGridComponent.jsx
+++ b/src/components/adslotUi/TreePicker/TreePickerGridComponent.jsx
@@ -26,17 +26,21 @@ const TreePickerGridComponent = ({
   selected,
   valueFormatter,
   emptyText,
+  displayGroupHeader,
 }) => {
   const nodesByGroupLabel = _.groupBy(nodes, groupFormatter);
   return (
     <Grid>
       {_.map(nodesByGroupLabel, (groupedNodes, label) =>
         <div className="treepickergrid-component-group" key={_.kebabCase(label)}>
-          <div className="treepickergrid-component-group-label">
-            <GridRow dts={`group-label-${_.kebabCase(label)}`}>
-              {label}
-            </GridRow>
-          </div>
+          {(displayGroupHeader) ?
+            <div className="treepickergrid-component-group-label">
+              <GridRow dts={`group-label-${_.kebabCase(label)}`}>
+                {label}
+              </GridRow>
+            </div> :
+            null
+          }
           {_.map(groupedNodes, (node) =>
             <TreePickerNodeFast
               key={node.id}
@@ -83,12 +87,14 @@ TreePickerGridComponent.propTypes = {
   removeNode: PropTypes.func,
   selected: PropTypes.bool.isRequired,
   valueFormatter: PropTypes.func,
+  displayGroupHeader: PropTypes.bool,
 };
 
 TreePickerGridComponent.defaultProps = {
   disabled: false,
   hideIcon: false,
   groupFormatter: () => 'Default Group',
+  displayGroupHeader: true,
 };
 
 export default TreePickerGridComponent;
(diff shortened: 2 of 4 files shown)


src/components/adslotUi/TreePicker/TreePickerGridComponent.jsx
```diff
@@ -26,17 +26,21 @@ const TreePickerGridComponent = ({
...
   return (
     <Grid>
       {_.map(nodesByGroupLabel, (groupedNodes, label) =>
         <div className="treepickergrid-component-group" key={_.kebabCase(label)}>
-          <div className="treepickergrid-component-group-label">
-            <GridRow dts={`group-label-${_.kebabCase(label)}`}>
-              {label}
-            </GridRow>
-          </div>
+          {alwaysDisplayGroupHeader || _.size(nodesByGroupLabel) > 1 ?
```
lightbringer1991: return to the previous behaviour, but now we have a props to dictate the display

lightbringer1991:
[Publisher and sites and other modals where there is only one group shows a Default group label](https://trello.com/c/unO5m924/6571-publisher-and-sites-and-other-modals-where-there-is-only-one-group-shows-a-default-group-label)

src/components/adslotUi/TreePicker/TreePickerSimplePureComponent.jsx
```diff
@@ -129,11 +132,13 @@ TreePickerSimplePureComponent.propTypes = {
   subtree: PropTypes.arrayOf(TreePickerPropTypes.node.isRequired),
   svgSymbolCancel: PropTypes.shape(SvgSymbol.propTypes),
   svgSymbolSearch: PropTypes.shape(SvgSymbol.propTypes),
+  alwaysDisplayGroupHeader: PropTypes.bool,
 };
 
 TreePickerSimplePureComponent.defaultProps = {
   disabled: false,
   itemType: 'node',
+  alwaysDisplayGroupHeader: false,
```
xiaofan2406: `displayGroupHeader`?

lightbringer1991: default behaviour is displaying group headers if there are more than 1 group

so `displayGroupHeader` can be misleading, as it already displays group headers.

mdotwills: I also don't think `alwaysDisplayGroupHeader` is a good approach since it biases what could be a generic interpretation of the component. It should not matter how we intend to use it, it's actually an isolated component to begin with. `displayGroupHeader` with default props make sense to me.

lteacher: Cant we just not do the magical display if size > 1 || whatever logic because the name is terrible. The component should just be on off without wizard logic. A user of this library should wrap or manage whether or not they need to display the group header.

Is group header the term here? what are other props called. I can see `node` used everywhere so is it not a NodeHeader e.g. `[show|display]NodeHeaders` actually also not convinced.

src/components/Main.jsx
```diff
@@ -641,6 +641,7 @@ class AppComponent extends React.Component {
           includeNode={_.noop}
           removeNode={_.noop}
           nodeRenderer={nodeRenderer}
+          alwaysDisplayGroupHeader
```
mdotwills: I think this prop also needs to be updated?

src/components/adslotUi/TreePicker/TreePickerGridComponent.jsx
```diff
@@ -83,12 +87,14 @@ TreePickerGridComponent.propTypes = {
   removeNode: PropTypes.func,
   selected: PropTypes.bool.isRequired,
   valueFormatter: PropTypes.func,
+  displayGroupHeader: PropTypes.bool,
```
mdotwills: 👍

test/components/adslotUi/TreePicker/TreePickerGridComponentTest.jsx
```diff
@@ -53,45 +54,66 @@ describe('TreePickerGridComponent', () => {
...
+        displayGroupHeader: false,
+      });
+      const component = shallow(<TreePickerGrid {...props} />);
+      const gridElement = component.find(Grid);
+      expect(gridElement).to.have.length(1);
+      expect(gridElement.children()).to.have.length(2); // One group node and an Empty.
+      const groupElement = gridElement.find('.treepickergrid-component-group');
+      expect(groupElement).to.have.length(1);
+      expect(groupElement.children()).to.have.length(1);
+    });
```
mdotwills: good work on these tests

mdotwills:
Merging.. 🥇

PR #583, opened 2017-07-12
Update: AlertInput: accept type, min, placeholder as props

xiaofan2406:
- make `AlertInput` accpet `type` as a prop, and pass it on the the internal `input` html element.

### context
- need to use a `number` input in the create product form

### DIFF
--- a/src/components/adslotUi/AlertInput/index.jsx
+++ b/src/components/adslotUi/AlertInput/index.jsx
@@ -62,6 +62,9 @@ export default class AlertInput extends Component {
     const {
       defaultValue,
       value,
+      type,
+      min,
+      placeholder,
       prefixAddon,
       suffixAddon,
       alertStatus = '',
@@ -85,9 +88,11 @@ export default class AlertInput extends Component {
         {prefixAddon ? <span className={`${baseClass}-addon`}>{prefixAddon}</span> : null}
         <input
           className={`${baseClass}-input`}
-          type="text"
+          type={type}
           defaultValue={defaultValue}
           value={value}
+          min={min}
+          placeholder={placeholder}
           onChange={onValueChange}
           onFocus={this.handleInputFocus}
           onBlur={this.handleInputBlur}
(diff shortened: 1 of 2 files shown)


src/components/adslotUi/AlertInput/index.jsx
```diff
@@ -110,10 +111,16 @@ export default class AlertInput extends Component {
 AlertInput.propTypes = {
   defaultValue: PropTypes.string,
   value: PropTypes.string,
+  type: PropTypes.string,
```
patrick-ivers: change to `PropTypes.oneOf(...)`, and put the types you want to allow in there

src/components/adslotUi/AlertInput/index.spec.jsx
```diff
@@ -122,6 +122,11 @@ describe('AlertInput', () => {
       expect(overlayElement.prop('placement')).to.equal('bottom');
     });
 
+    it('should render different types of input', () => {
```
mdotwills: it should (also) render default props

src/components/adslotUi/AlertInput/index.spec.jsx
```diff
@@ -122,6 +122,11 @@ describe('AlertInput', () => {
...
+      expect(shallow(<AlertInput />).find('input').prop('type')).to.equal('text');
```
mdotwills: then in this test case, set the input type to something else other than `text`

src/components/adslotUi/AlertInput/index.jsx
xiaofan2406: for now the usage are only `text` and `number`. Should I just add these two?

patrick-ivers: Yep.

xiaofan2406:
@mdotwills @patrick-ivers incorporated reviews.

patrick-ivers:
Merging...

PR #584, opened 2017-07-13
New: add Tabs component

lightbringer1991:
adslottabs

### DIFF
--- a/src/components/adslotUi/AdslotTabs/index.jsx
+++ b/src/components/adslotUi/AdslotTabs/index.jsx
@@ -1,0 +1,35 @@
+import _ from 'lodash';
+import React, { PropTypes } from 'react';
+import { Tabs as BsTabs } from 'react-bootstrap';
+import './styles.scss';
+
+const Tabs = (props) => {
+  const { adslotStyle } = props;
+  const bsTabsProps = _.omit(props, ['adslotStyle']);
+
+  switch (adslotStyle) {
+    case 'chevron': {
+      const className = _([props.className, 'chevron-tabs-component']).compact().join(' ');
+      return (<BsTabs {...bsTabsProps} className={className}>
+        {React.Children.map(props.children, (tab) => { // eslint-disable-line
+          const tabClassName = _([tab.props.tabClassName, 'chevron-tab-component']).compact().join(' ');
+          return React.cloneElement(tab, { tabClassName });
+        })}
+      </BsTabs>);
+    }
+
+    default: {
+      return (<BsTabs {...bsTabsProps}>{props.children}</BsTabs>);
+    }
+  }
+};
+
+Tabs.propTypes = _.assign({}, BsTabs.propTypes, {
+  adslotStyle: PropTypes.oneOf(['default', 'chevron']),
+});
+
+Tabs.defaultProps = {
+  adslotStyle: 'default',
+};
+
+export default Tabs;
--- a/src/components/Main.jsx
+++ b/src/components/Main.jsx
@@ -40,8 +40,6 @@ import {
   Statistic,
   SvgSymbol,
   SvgSymbolCircle,
-  Tab,
-  Tabs,
   Tag,
   TileGrid,
   Totals,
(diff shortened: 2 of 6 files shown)


lightbringer1991:
[Create product: create Chevron tabs](https://trello.com/c/ld74Yelw/6608-create-product-create-chevron-tabs)

src/components/adslotUi/AdslotTabs/index.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+Tabs.propTypes = _.assign({}, BsTabs.propTypes, {
+  adslotStyle: PropTypes.oneOf(['default', 'chevron']),
+});
+
+Tabs.defaultProps = {
+  adslotStyle: 'default',
+};
+
+export default Tabs;
```
sonhanguyen: still I think you should re-export `Tab` here, but both this and `Tab` are consumed via `adslot-ui` anyway so it matters less I guess.

omgaz: @TuanNguyen I thought these chevron tabs had been descoped. They were already scoped as an optional 'nice-to-have'.

Similarly I have a lot of problems with naming used here.

1. AdslotTabs should just be Tabs. We're already in AdslotUi, we're exporting as Tabs, we shouldn't care where things have come from. Everything is Adslot.

2. The naming of adslotStyle, same adslot is redundant - and style leads me to believe we're applying a style tag (which isn't the case). Type would have been better. And on that note, we have default and chevron…

3. … default is default and is applied as a default prop… Can't default just be the default... why do we need to default to default as a string? If the prop isn't provided we want default behaviour, just return bootstrap Tabs.

4. Which gives us just the one oneOf ['chevron']… so we should probably just have a prop of chevron that's a boolean value… much cleaner. If we plan to support further types we could consider keeping an array but it's managing a lot of extra complexity now.

5. Chevron is also a bad name… similar to the discussion in the other PR. Chevron (like underline) is describing implementation, not intention. Intention shows meaning, context and responsibility. The chevron is a visual representation of the intent of showing progress, transition, steps. Can you consider, and discuss, a better name something like directional, progression, steps... and if it's a boolean name appropriately with is/has nomenclature or something with meaning like 'withProgress', 'isDirectional', 'isProgressive', 'hasOrder'. You'd also not need to do `_.omit(props, ['adslotStyle']);` if it's a boolean value.

src/components/adslotUi/AdslotTabs/index.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+const Tabs = (props) => {
+  const { adslotStyle } = props;
+  const bsTabsProps = _.omit(props, ['adslotStyle']);
+
+  switch (adslotStyle) {
+    case 'chevron': {
+      const className = _([props.className, 'chevron-tabs-component']).compact().join(' ');
+      return (<BsTabs {...bsTabsProps} className={className}>
+        {React.Children.map(props.children, (tab) => { // eslint-disable-line
```
omgaz: Why are we disabling the linter? What's the linter protecting against. Clearly it thinks something bas is being done on this line. If we need to disable it for this line it's fine. If we need to, and should, disable it globally (because it makes sense to) that's fine too.

src/components/adslotUi/AdslotTabs/index.jsx
```diff
@@ -0,0 +1,35 @@
+import _ from 'lodash';
+import React, { PropTypes } from 'react';
+import { Tabs as BsTabs } from 'react-bootstrap';
+import './styles.scss';
+
+const Tabs = (props) => {
+  const { adslotStyle } = props;
+  const bsTabsProps = _.omit(props, ['adslotStyle']);
```
omgaz: Redundant (see comment below)

src/components/adslotUi/AdslotTabs/index.jsx
sonhanguyen: @omgaz linter suggests `React.Children.map` should be replaced by _.map, which is not what we want.
The directive can be more specific though.

omgaz: cool 👍

lightbringer1991:
@omgaz it is descoped but not removed, we're still having it as a nice to have
anyway I'll come back to this one later when I have time, this is a nice tab to have imo

PR #585, opened 2017-07-19
Update: add `TreePickerSimplePureDemo` component

lightbringer1991:
treepicker

### DIFF
--- a/src/styles/adslotUi/TreePickerNav.scss
+++ b/src/styles/adslotUi/TreePickerNav.scss
@@ -3,7 +3,6 @@
 .treepickernav-component {
   border-left: $border-lighter;
   border-right: $border-lighter;
-  min-height: 66px; // Fix Chrome rendering bug when the node list is longer than one page.
   padding: 10px;
 
   > div + div {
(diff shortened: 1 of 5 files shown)


src/styles/adslotUi/TreePickerNav.scss
```diff
@@ -3,7 +3,6 @@
 .treepickernav-component {
   border-left: $border-lighter;
   border-right: $border-lighter;
-  min-height: 66px; // Fix Chrome rendering bug when the node list is longer than one page.
```
lightbringer1991: I couldn't see any bug when rendering a long node list

src/styles/adslotUi/TreePickerSimplePure.scss
```diff
@@ -12,7 +12,6 @@
   .treepickernav-component {
     background-color: $color-background;
     border-top: $border-lighter;
-    min-height: 71px; // Align the rows after this item with the other pane's rows.
```
lightbringer1991: I don't know if this is required anymore

omgaz:
I don't think this is the right approach. The tree picker should always have a search. It aids usability and provides navigation, I'm not sure I see the value in creating a split in the implementation of this component.

If this is for the survey modal we're only hiding it so we can merge and add as a phase 2, in the time to write this the search probably could have been implemented :/

omgaz:
As for the example (demo) that's a great improvement that can be merged.

lightbringer1991:
@omgaz ok I'll remove the props, but will keep the demo so we have a better example for the picker

omgaz:
LGTMerging 🍜

PR #588, opened 2017-07-30
Upgrade: NodeJS v6 and minor dependencies

omgaz:


.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: node_js
 node_js:
 - '6'
-- '4'
```
bchew: If the node engine in the `package.json` specifies >=6, we should be adding 7, 8 as well?

omgaz: fixed to only support LTS

bchew:
Merging :rocket:

PR #590, opened 2017-08-02
Update: TreePickerSimplePureComponent to support action on enter key

Maxmus1989:
TreePickerSimplePureComponent
- Change `onChange: PropTypes.func` from `searchOnChange: PropTypes.func`.
- Change `onCancel: PropTypes.func` from `searchOnClear: PropTypes.func`.
- Change `searchOnChange: PropTypes.bool` from to determine `onSearch` mode (on value change).
- Add `searchOnEnterKey: PropTypes.bool` to determine `onSearch` mode (on enter key).
- Add `debounceInterval: PropTypes.number` to support debounce on `onSearch` when `searchOnChange` mode.

out

out
#


RyanKempe:
looks good. please make sure we can choose between the different modes of searching for each node
... as we will adopt both modes of searching in this tree picker implementation

__2017c5__create_dfp_product_from_custom_line_item_on_design_sprint___trello

RyanKempe:
(I need to investigate which style of searching we want to use for key values and connection)

vinteo:
Does it have to be a separate function? Do we have to manually pass in a noop for `onChange` if we don't want it to trigger on change? Would it be better to pass in a `changeOnEnter` falg and reuse the `onChange` function?

src/components/alexandria/Search/index.jsx
```diff
@@ -16,6 +17,14 @@ const Search = ({
   const searchClassSuffixes = !disabled ? svgSymbolSearch.classSuffixes : ['color-disabled'];
   const cancelClassSuffixes = !disabled ? svgSymbolCancel.classSuffixes : ['color-disabled'];
   const onChangeBound = !disabled ? (event) => onChange(_.get(event, 'target.value')) : null;
+  const onKeyPress = !disabled ? (event) => {
+    switch (event.key) {
+      case 'Enter':
```
xiaofan2406: I am not sure if this is the most compatible approach. I have read somewhere to use `event.which`.

src/components/alexandria/Search/index.jsx
```diff
@@ -16,6 +17,14 @@ const Search = ({
...
+        return onEnterKey ? onEnterKey() : null;
```
xiaofan2406: maybe for now `if (event.which === 13 && onEnterKey) { onEnterKey(); }`

src/components/adslotUi/TreePicker/TreePickerNavComponent.jsx
```diff
@@ -22,6 +23,7 @@ const TreePickerNavComponent = ({
       disabled={disabled}
       onChange={searchOnChange}
       onClear={searchOnClear}
+      onEnterKey={searchOnEnterKey}
```
xiaofan2406: Theres is more to do in this component.

I think we want the ability to `not to search onChange` when `onEnterKey` handler is provided. Otherwise the request will still be send off multiple times anyway.

lightbringer1991:
I would like to propose new props for the search component
```jsx
TreePickerNavComponent.propTypes = {
	breadcrumbNodes: PropTypes.arrayOf(TreePickerPropTypes.breadCrumbNode),
	breadcrumbOnClick: PropTypes.func,
	disabled: PropTypes.bool,
	onSubmit: PropTypes.func.isRequired,
	searchOnEnter: PropTypes.bool,
	searchValue: PropTypes.string,
	svgSymbolCancel: PropTypes.shape(SvgSymbol.propTypes),
	sgvSymbolSearch: PropTypes.shape(SvgSymbol.propTypes),
};

TreePickerNavComponent.defaultProps = {
	disabled: false,
	searchOnChange: true,
	searchOnEnter: false,
};
```

if `searchOnEnter == true`, `searchOnChange` is ignored
This way, we decouple the submission action away from onChange props, and therefore we can bind submit action into enter key

If the proposal is passed, this is a breaking change to all treepickers
@vinteo @Maxmus1989 @xiaofan2406

Maxmus1989:
Few questions to consider:
- Do we still want to pass in `searchValue: PropTypes.string` and the `input` value will always referring to parent/wrapper component flux cycle value?
   - If yes, we need to pass in `onChange: PropTypes.func` (currently named as `searchOnChange`) to update the `searchValue`.
   - If no, the value will not be control from parent/wrapper component. We can just call `onSubmit ` when `onChange` or `Enter key`.

vinteo:
yea I like it better this way. I think its better to keep it stateless as well so we shouldn't need to keep the value in redux. Unless there is a need for it, but I can't think of any use of it.

src/components/adslotUi/TreePicker/TreePickerNavComponent.jsx
```diff
@@ -39,8 +47,12 @@ TreePickerNavComponent.propTypes = {
   breadcrumbNodes: PropTypes.arrayOf(TreePickerPropTypes.breadCrumbNode),
   breadcrumbOnClick: PropTypes.func,
   disabled: PropTypes.bool,
-  searchOnChange: PropTypes.func,
-  searchOnClear: PropTypes.func,
+  onChange: PropTypes.func,
+  onClear: PropTypes.func,
+  onSearch: PropTypes.func,
+  debounceOnChangeSearchMS: PropTypes.number,
+  searchOnChange: PropTypes.bool,
```
lightbringer1991: maybe add `defaultProps` for the newly added props

src/components/adslotUi/TreePicker/TreePickerSimplePureComponent.jsx
```diff
@@ -15,6 +15,7 @@ const TreePickerSimplePureComponent = ({
   additionalClassNames,
   breadcrumbNodes,
   breadcrumbOnClick,
+  debounceOnChangeSearchMS,
```
lightbringer1991: just wondering what does `MS` stand for

Maxmus1989: millisecond

src/components/adslotUi/TreePicker/TreePickerNavComponent.jsx
Maxmus1989: done

src/components/adslotUi/TreePicker/TreePickerSimplePureComponent.jsx
lightbringer1991: can't we include the logic inside `onChange` or `onSearch` logic, since we can also delay the logic inside that function, therefore we don't need an extra props. And it allows flexibility (like delaying the search until another Promise is resolved, etc.)

e.g. 
```js
const onSearch = (event) -> {
    // do stuff
};
const onSearchDebounce = _.debounce(onSearch, 1000);


<TreePickerSimplePureComponent {...props} onSearch={onSearchDebounce} />
```

just wondering if the above solution works

Maxmus1989: Yes that could work. I can remove debounce related props.

src/components/alexandria/Search/index.jsx
```diff
@@ -1,63 +1,109 @@
...
+  onChange: PropTypes.func,
+  onSearch: PropTypes.func,
   placeholder: PropTypes.string.isRequired,
   svgSymbolCancel: PropTypes.shape(SvgSymbol.propTypes),
   svgSymbolSearch: PropTypes.shape(SvgSymbol.propTypes),
-  value: PropTypes.string.isRequired,
+  value: PropTypes.string,
+  searchOnChange: PropTypes.bool,
+  searchOnEnterKey: PropTypes.bool,
+  debounceMS: PropTypes.number,
```
vinteo: our camel case style is `debounceMs`

src/components/adslotUi/TreePicker/TreePickerSimplePureComponent.jsx
vinteo: I guess it would be nice to have a default debounce that can be overwritten.

src/components/adslotUi/TreePicker/TreePickerNavComponent.jsx
```diff
@@ -39,15 +47,21 @@ TreePickerNavComponent.propTypes = {
   breadcrumbNodes: PropTypes.arrayOf(TreePickerPropTypes.breadCrumbNode),
   breadcrumbOnClick: PropTypes.func,
   disabled: PropTypes.bool,
-  searchOnChange: PropTypes.func,
-  searchOnClear: PropTypes.func,
+  onChange: PropTypes.func,
+  onClear: PropTypes.func,
+  onSearch: PropTypes.func,
+  debounceOnChangeSearchMS: PropTypes.number,
```
vinteo: I think `debounceOnSearchMs` is enough

src/components/adslotUi/TreePicker/TreePickerSimplePureComponent.jsx
lightbringer1991: I guess we can have the props as optional, so we don't need to utilize it most of the time.

lightbringer1991: also I would expect to have no delay when user presses enter key. It is logical to have delay in onChangeSearch, but I think it's not very user friendly when we have delay on enter.

Maxmus1989: Yup, the current behaviour for `debounce` only happen for `searchOnChange` mode. when switch to `searchOnEnterKey` it won't `debounce anymore. Also when user click on `cancel/clear` key it won't debounce as well.

Maxmus1989: @lightbringer1991 `_.debounce` 😃 https://lodash.com/docs/4.17.4#debounce

src/components/alexandria/Search/index.jsx
```diff
@@ -1,63 +1,107 @@
...
+  }
+
+  onChangeBound(event) {
+    const { disabled, searchOnChange, onChange } = this.props;
+    if (disabled) return;
+
+    const value = _.get(event, 'target.value');
+
+    this.setState({ value });
+    onChange(value);
```
lightbringer1991: I assume that the current `onChangeSearch` function takes the value as the parameter and not the event itself?

src/components/alexandria/Search/index.jsx
```diff
@@ -1,63 +1,107 @@
...
+        {_.isEmpty(this.state.value)
+          ? <SvgSymbol href={svgSymbolSearch.href} classSuffixes={searchClassSuffixes} />
+          : <SvgSymbol href={svgSymbolCancel.href} classSuffixes={cancelClassSuffixes} onClick={this.onClearBound} />
+        }
+      </div>
+    );
+  }
+}
 
 Search.displayName = 'AlexandriaSearchComponent';
```
lightbringer1991: I think we should change all `Alexandria` footprint in this file to just `AdslotUI`, or something that we use in other components

src/components/alexandria/Search/index.jsx
```diff
@@ -1,63 +1,107 @@
...
+  onChange: PropTypes.func,
+  onSearch: PropTypes.func,
   placeholder: PropTypes.string.isRequired,
   svgSymbolCancel: PropTypes.shape(SvgSymbol.propTypes),
   svgSymbolSearch: PropTypes.shape(SvgSymbol.propTypes),
-  value: PropTypes.string.isRequired,
+  value: PropTypes.string,
+  searchOnChange: PropTypes.bool,
+  searchOnEnterKey: PropTypes.bool,
+  debounceMs: PropTypes.number,
```
lightbringer1991: I'm still convinced that the `Ms` does not need to be here, maybe `debounceTime` or `debounceInterval` would be better?

src/components/alexandria/Search/index.jsx
```diff
@@ -1,63 +1,107 @@
...
 };
 
 Search.defaultProps = {
   disabled: false,
-
-  onChange: () => { throw new Error('Alexandria Search needs an onChange handler'); },
-
-  onClear: () => { throw new Error('Alexandria Search needs an onClear handler'); },
-
+  onChange: _.noop,
```
lightbringer1991: default props on `onClear` as well I think

src/components/alexandria/Search/index.jsx
Maxmus1989: yup

src/components/alexandria/Search/index.jsx
Maxmus1989: `debounceInterval` sound good, will update that.

src/components/adslotUi/TreePicker/TreePickerNavComponent.jsx
```diff
@@ -39,15 +47,21 @@ TreePickerNavComponent.propTypes = {
   breadcrumbNodes: PropTypes.arrayOf(TreePickerPropTypes.breadCrumbNode),
   breadcrumbOnClick: PropTypes.func,
   disabled: PropTypes.bool,
-  searchOnChange: PropTypes.func,
-  searchOnClear: PropTypes.func,
+  onChange: PropTypes.func,
+  onClear: PropTypes.func,
+  onSearch: PropTypes.func,
+  debounceOnSearchMs: PropTypes.number,
```
lightbringer1991: maybe change this to `debounceInterval` too?

src/examples/components/TreePickerDebounceDemo.jsx
```diff
@@ -0,0 +1,11 @@
+import TreePickerSimplePureDemo from './TreePickerSimplePureDemo';
+
+class TreePickerDebounceDemo extends TreePickerSimplePureDemo {
+  constructor(props) {
+    super(props);
+    this.state.debounceOnSearchMs = 500;
```
lightbringer1991: probably change the demo state variable to the corresponding name as well

src/components/adslotUi/Search/index.jsx
```diff
@@ -0,0 +1,116 @@
...
+    if (searchOnChange) this.debounceOnSearch(value);
+  }
+
+  onKeyPress(event) {
+    const { disabled, searchOnEnterKey, onSearch, value } = this.props;
+    if (disabled) return;
+
+    switch (true) {
+      case (searchOnEnterKey && event.which === 13):
+        onSearch(value);
```
xiaofan2406: ```onSearch(this.state.value)```

src/components/adslotUi/Search/index.jsx
```diff
@@ -0,0 +1,116 @@
...
+    onChange(value);
+
+    if (searchOnChange) this.debounceOnSearch(value);
+  }
+
+  onKeyPress(event) {
+    const { disabled, searchOnEnterKey, onSearch, value } = this.props;
+    if (disabled) return;
+
+    switch (true) {
```
xiaofan2406: Is it possible to use `if`, since we only care about `13` for now.

`switch(true)` is kinda hacky for me.

src/components/adslotUi/Search/index.jsx
```diff
@@ -0,0 +1,116 @@
...
+
+  onClearBound() {
+    const { disabled, onChange, onClear, onSearch } = this.props;
+    if (disabled) return;
+
+    const value = '';
+
+    this.setState({ value });
+    onChange(value);
+    onSearch(value);
```
xiaofan2406: should `onSearch` be fired when clear? Its a bit counter intuitive for me.

Maxmus1989: yea it should.. since when clear we are actually setting the value to empty string

xiaofan2406: But then you are assuming `searchOnChange` is always `true`. I guess whats missing is wrapping it in the `if` and use `this.debounceOnSearch` instead of `onSearch` directly?

Maxmus1989: @xiaofan2406 The reason why we should not debounce `onSearch` at here is simply because once we click on clear button it should call immediately. Don't think anyone are expecting debounce on this button behaviour, no matter they are in `searchOnChange` or `searchOnEnterKey` mode.

src/components/adslotUi/Search/index.jsx
```diff
@@ -0,0 +1,111 @@
...
+    const value = '';
+
+    this.setState({ value });
+    onChange(value);
+    onClear(value);
+  }
+
+  render() {
+    const { disabled, placeholder, svgSymbolCancel, svgSymbolSearch } = this.props;
+    const searchClassSuffixes = !disabled ? svgSymbolSearch.classSuffixes : ['color-disabled'];
```
sonhanguyen: can we remove the `!` and have the ternary other way around?

src/components/adslotUi/Search/index.jsx
```diff
@@ -0,0 +1,111 @@
...
+  constructor(props) {
+    super(props);
+    this.state = { value: props.value };
+    this.debounceOnSearch = _.debounce(props.onSearch, props.debounceInterval);
+    this.onChangeBound = this.onChangeBound.bind(this);
+    this.onKeyPress = this.onKeyPress.bind(this);
+    this.onClearBound = this.onClearBound.bind(this);
+  }
+
+  onChangeBound(event) {
```
sonhanguyen: I don't quite get the naming here, technically this is `onChange`, only when you call `onChange.bind(this)` you get a bound function of which the name `onChangeBound` makes sense.
I myself sometimes confused not knowing what convention to follow, between `__change`, `onChange`, `change` and `changeBound`. I've tried naming the function just by the verb, then use `on*Click` for the bound one, but then it can be confusing because now we have 2 name for one function and we really only care about the bound one.
I could be overthinking this, may be @xiaofan2406 can share some thought.

Maxmus1989: Hmm.. In this case, I think just `onChange` make more sense.

mdotwills:
Merging.. 🚢

PR #591, opened 2017-08-04
Update: tree picker options to hide search on root

Maxmus1989:
out


lightbringer1991:
@omgaz could you please take a look at this one. Looks like what we're trying to achieve in the audience survey segment picker

omgaz:
Can you give some more detail? Which other picker and what is the goal? Sorry, been away a bit and forget.

lightbringer1991:
@omgaz the hide search on the survey request one

omgaz:
LGTM. Pedros and I need to go through our component library when i get back and solidify the UX patterns we want to adopt. I feel like this should always just be the default behaviour - but maybe cases where not.

vinteo:
Merging...

PR #592, opened 2017-08-07
New: Introducing flat button

pphminions:
Introducing a new flat button without any borders, even when hover over the button. This is useful for close buttons. 

 
screen shot 2017-08-07 at 11 30 42 am

can use for buttons like - close button
screen shot 2017-08-07 at 11 31 42 am

can't use borderless - focus or hover will add a border for the borderless button
screen shot 2017-08-07 at 11 33 52 am

### DIFF
--- a/src/components/Main.jsx
+++ b/src/components/Main.jsx
@@ -391,6 +391,9 @@ class AppComponent extends React.Component {
           <Button className="btn-borderless" disabled>
             Borderless Disabled
           </Button>
+          <Button className="btn-flat">
+            Flat Button
+          </Button>
         </div>
 
         <div className="btn-panel">
--- a/src/styles/bootstrapOverrides/Button.scss
+++ b/src/styles/bootstrapOverrides/Button.scss
@@ -135,6 +135,19 @@
   }
 }
 
+.btn-flat {
+  &:not(:disabled) {
+    @include button-variant($color-text-light, $color-inverse, $color-white);
+    border: 0;
+    box-shadow: none;
+
+    &:hover,
+    &:focus {
+      background-color: $color-white;
+    }
+  }
+}
+
 .btn-xs,
 .btn-group-xs > .btn {
   font-size: $font-size-tiny;


pphminions:
Can not use borderless button - since when hover or focus on the button, the button border will be visible

src/components/Main.jsx
```diff
@@ -391,6 +391,9 @@ class AppComponent extends React.Component {
           <Button className="btn-borderless" disabled>
             Borderless Disabled
           </Button>
+          <Button className="btn-flat">
```
lightbringer1991: will the flat button be the same with our existing close buttons on `TagComponent`?

if it's the same, maybe the tags that we're trying to do is just a different style of `TagComponent`

pphminions:
@lightbringer1991 u mean the inverse tag component?
screen shot 2017-08-07 at 11 41 15 am

src/components/Main.jsx
pphminions: @lightbringer1991 u mean the inverse tag component?
screen shot 2017-08-07 at 11 41 15 am

vinteo:
does the font colour change on hover? For example the current `X` buttons changes colour on hover

pphminions:
@vinteo - even though if I changed to change the colour for the text when hover or focus, it won't work for the close button. Close button icon is a svg icon
I will make the icon colour bit darker as per the design. 
this is from the design
screen shot 2017-08-07 at 11 59 25 am

lightbringer1991:
@pphminions maybe check with Pedros if he really wants the close button to be darker

pphminions:
@lightbringer1991 - the button should be darker when the user hover over the button. Same as the modal close button.
It should be a css change from the direct-web side

pphminions:
@vinteo - the svg image colour changed from the direct-web and it will look like this. (ignore the border since it is still a borderless-btn)
close button

pphminions:
@vinteo, @sonhanguyen, @sharmajyo, @lightbringer1991 - please merge

vinteo:
Merging...

PR #593, opened 2017-08-10
Major updates

omgaz:
### Background Context

Starting the arduous task of prepping for webpack 3 / react 15

Big tasks ahead in updating our linters

### Changes

Upgrade: Sinon
Upgrade: Seamless Immutable
Upgrade: Chai
Upgrade: Autoprefixer
Upgrade: ESLint plugin lodash
Upgrade: Datepicker
Fix: Datepicker month font-size and style


src/components/Main.jsx
```diff
@@ -634,7 +634,7 @@ class AppComponent extends React.Component {
         <h2>TreePickerSimplePure without initial state</h2>
 
         <TreePickerSimplePure
-          groupFormatter={(node) => `${node.label.split(' ').length} words`}
+          groupFormatter={(node) => `${_.split(node.label, ' ').length} words`}
```
omgaz: prefer lodash by default, we can choose to ignore for these methods.

src/components/adslotUi/Search/index.spec.jsx
```diff
@@ -109,6 +109,7 @@ describe('Search', () => {
     beforeEach(() => {
       props.searchOnEnterKey = true;
       props.searchOnChange = false;
+      props.disabled = false;
```
omgaz: sinon stub can't stub an undefined prop

src/components/Main.jsx
lightbringer1991: I would prefer lodash in this case

.eslintrc
```diff
@@ -48,9 +48,8 @@
     "lodash/no-commit": 2,
     "lodash/no-double-unwrap": 2,
     "lodash/no-extra-args": 2,
-    "lodash/no-single-chain": 2,
+    "lodash/chaining": [2, "always"],
```
lightbringer1991: hmm maybe 3, as some chainings call `values()` at the end of the chain. I'm not sure if it counts as a chain. But if it does, then with 2 we can only do one thing, which is basically equivalent to a normal lodash function

lightbringer1991:
merging 🚀

PR #594, opened 2017-08-21
Fix: Alert Input User-Agent padding

omgaz:
Fixes Adslot Product input field growth in line items.

 - Remove user-agent styling of 1px from alert input field to prevent un-expected container heights.
 - Add 1px more left/right padding for breathability of inputs.

screen shot 2017-08-21 at 11 01 52

### DIFF
--- a/src/components/adslotUi/AlertInput/styles.scss
+++ b/src/components/adslotUi/AlertInput/styles.scss
@@ -8,7 +8,7 @@
   border: $border-lighter;
   border-radius: $border-radius;
   display: flex;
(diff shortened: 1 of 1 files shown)


omgaz:
[Bug: Line item height is inconsistent between read-only/edit mode and site/product/publisher names cling to the bottom of the cell](https://trello.com/c/RboKAxx0/6767-bug-line-item-height-is-inconsistent-between-read-only-edit-mode-and-site-product-publisher-names-cling-to-the-bottom-of-the-cel)

PR #597, opened 2017-09-01
Build: Deprecate PhantomJS in favour of Headless Chrome

omgaz:
Massive speed increase when running tests.

- Remove accidental commit of package-lock.json
- Remove phantomjs
- Add karma chrome and gonzales

Should speed up npm installs as phantom was huge.


omgaz:
FYI Chrome 60 was released a month ago which supports this on Windows. Will test in a VM.

omgaz:
Works a treat on Windows.

PR #598, opened 2017-09-03
Breaking: New Component Library Layout

omgaz:
Separation of docs from src and added navigation and logo.

Phase 2 will involve updating examples with prop types and notes (PR to follow)

screen shot 2017-09-05 at 16 53 36

screen shot 2017-09-20 at 17 08 54

#### Changes

 - Added side navigation
 - Switch `<h2>`'s to `<PageTitle>` components (dog-food)
 - Migrate example/demo code to `/docs` directory
 - Delete `MainTest`, this has come up multiple times and thought we'd trash it now that it's independent.
 - Added `id` prop to PageTitle component (allows page anchors to navigate to)
 - Co location of tests, styles and mocks
 - Added yeoman-esque component scaffolder that uses a more visually aware and customisable template directory
 - deprecated yeoman

Folder structure of components

```
/src
  /components
    /adslot-ui
      /MyComponent
        index.jsx
        index.spec.jsx
        styles.scss
```

Components are aliased through `adslot-ui`.

import via through folder name will automatically load `index`.


lightbringer1991:
I'm thinking about having a sample code section right under the component itself
so it's easy for user to implement and they don't have to inspect the code

xiaofan2406:
I feel these internal `Nav` `Grid` are not exported individually. It may be too complex for them to have their own folder

```
/TreePicker
  Grid.jsx
  Grid.scss
  Nav.jsx
  Nav.scss
  Node.jsx
  Node.scss
```
@omgaz whats your thought on this?

omgaz:
@xiaofan2406 my thoughts were complicated on this one.

Something like a TreePicker is made up of Nav, Grids and Nodes. A grid has Nodes and Nodes have NodeExpanders.

You'd never use a Node outside the context of TreePicker and was worried about mis-use and so wanted the index of the TreePicker to only export the treepicker and the nested behaviour of the component, with their own exports, still allow for use of composing new TreePicker components.

It also keeps each folder predictable (always index, spec and style).

Most of this PR was grep/find and replace etc.. but am open to changing.

omgaz:
> Is this a good opportunity to remove alexandria and adslot-ui folder? have all components directly under components folder?

Yea seems like a good idea. @Adslot/frontend any problem with this?

README.md
```diff
@@ -14,10 +14,9 @@ This includes Alexandria and third-party restyled components, e.g. Bootstrap.
 Technology:
 
 - [React](http://facebook.github.io/react/)
-- ES2015 (using [Babel](http://babeljs.io)\)
+- ES201x (using [Babel](http://babeljs.io)\)
```
sonhanguyen: 2020 is fast approaching

docs/examples/components/selects.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Select } from 'components/distributionEntry';
+import Select from 'react-select';
```
sonhanguyen: why is this an example of `react-select` and not `adslot-ui`'s select

sonhanguyen:
I think someone mentioned it when moved the styles, can we turn on `css-modules` & consider it an opt-in thing?

README.md
omgaz: :P

omgaz:
@sonhanguyen sure will look at css modules in another pr. Maybe fix-up friday

docs/examples/components/selects.jsx
omgaz: It's because adslot-ui doesn't have a select component and the distribution entry just exports react-select.

I made this decision so we can see that these are not adslot-ui components even though we export them. I think it will make it more obvious that these components should be brought internally (either with a wrapper/hoc or written ourselves).

PR #600, opened 2017-09-11
Build: Upgrade to Webpack 3

omgaz:
Upgrade build to Webpack 3.
Clean-up and optimise configs.
Add documentation on config.
Upgrade related build dependencies including react-hot-loader.


.babelrc
```diff
@@ -1,9 +1,23 @@
 {
   "presets": [
-    "es2015",
-    "react"
+    ["env", {
```
omgaz: > At the moment, several presets let you determine what features Babel should support, babel-preset-env lets you specify an environment and only transpiles features that are missing in that environment.

But maybe we can leave it as es2015 (or do we upgrade to 2016).

.babelrc
```diff
@@ -1,9 +1,23 @@
...
+          "last 2 versions",
+          "> 5%",
+          "ie >= 8",
+          "not and_chr > 0",
+          "not and_uc > 0",
+          "not android > 0",
+          "not ie_mob > 0",
+          "not ios_saf > 0",
+          "not op_mini > 0"
+        ]
```
omgaz: This was just the default list we were already using in autoprefixer. We can't export/import the same list into the config so we need it to be separate. This should probably be re-addressed at some point.

Do we know what browser support we wish to set as a baseline.

config/common.js
```diff
@@ -0,0 +1,185 @@
+// shared config (dev and prod)
+const { resolve } = require('path');
+const StyleLintPlugin = require('stylelint-webpack-plugin');
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
+
+const srcPath = resolve(__dirname, '../src');
+
+module.exports = {
+  resolve: {
+    extensions: ['.webpack.js', '.web.js', '.js', '.jsx'],
```
omgaz: remove `.webpack.js` and `.web.js`

config/common.js
```diff
@@ -0,0 +1,185 @@
...
+      commonjs2: 'redux',
+      commonjs: 'redux',
+      amd: 'redux',
+    },
+  },
+  performance: {
+    hints: false,
+  },
+};
+
```
omgaz: BELOW: remove all this, I just left it here as it was an easy reference whilst refactoring.

src/components/adslotUi/AlertInput/styles.scss
```diff
@@ -1,6 +1,6 @@
-@import 'styles/border';
-@import 'styles/color';
-@import 'styles/font-size';
+@import '~styles/border';
```
vinteo: what does this do?

config/common.js
```diff
@@ -0,0 +1,185 @@
...
+      commonjs2: 'react-redux',
+      commonjs: 'react-redux',
+      amd: 'react-redux',
+    },
+    'redux': {
+      root: 'Redux',
+      commonjs2: 'redux',
+      commonjs: 'redux',
+      amd: 'redux',
+    },
```
omgaz: Maybe moment should be in here as well.

src/components/adslotUi/AlertInput/styles.scss
omgaz: Required by webpack 3 and sass-loader to tell the loader that this is an aliased import and not a relative file path.

The alternative would be to use absolute/relative paths and do away with aliases altogether (which might not be a bad idea)

webpack.config.js
```diff
@@ -1,10 +1,8 @@
-'use strict';
-
 const path = require('path');
 const args = require('minimist')(process.argv.slice(2));
 
 // List of allowed environments
-const allowedEnvs = ['dev', 'dev-cold', 'dist', 'test'];
+const allowedEnvs = ['dev', 'dist', 'test'];
```
omgaz: dev-cold wasn't being used; correct me if I'm wrong.

package.json
```diff
@@ -8,7 +8,7 @@
     "check-commit": "node scripts/commit-msg",
     "clean": "rimraf dist/*",
     "codecov": "cat coverage/coverage-final.json | codecov",
-    "dist": "webpack --bail --env=dist",
+    "dist": "webpack --bail --env=dist --progress -p --hide-modules",
```
omgaz: --progress gives us a more information and a % indicator.

-p tells webpack this is a production build and keeps the bundle size down

--hide-modules just keeps the logs cleaner

config/dist.js
```diff
@@ -0,0 +1,123 @@
...
+        warnings: false,
+      },
+      output: {
+        comments: false,
+      },
+      sourceMap: false,
+      include: /\.js$/,
+      parallel: {
+        cache: true,
+        workers: 2,
```
omgaz: use 2 cpu cores (maybe runs a little quicker) @bchew @vinteo could this cause ci problems?

webpack.config.js
vinteo: I was using it cause I didn't want the hot loading (auto browser refresh) because  i was not testing on localhost (through virtual machine) not sure if that still applies

config/dist.js
bchew: Based on the travis-ci docs, that should be fine: https://docs.travis-ci.com/user/reference/overview/

config/common.js
```diff
@@ -0,0 +1,83 @@
...
+        test: /\.(js|jsx)$/,
+        use: ['babel-loader'],
+        exclude: /node_modules/,
+      },
+      {
+        test: /\.css$/,
+        use: [
+          'style-loader',
+          'css-loader',
+          'postcss-loader',
```
sonhanguyen: I notice this wasn't in before, what is the need?

config/common.js
```diff
@@ -0,0 +1,83 @@
...
+      {
+        test: /\.(js|jsx)$/,
+        use: ['babel-loader'],
+        exclude: /node_modules/,
+      },
+      {
+        test: /\.css$/,
+        use: [
+          'style-loader',
+          'css-loader',
```
sonhanguyen: can we get `css-modules` too?

package.json
```diff
@@ -60,47 +60,50 @@
...
-    "open": "^0.0.5",
-    "postcss-loader": "^0.9.1",
+    "open-browser-webpack-plugin": "0.0.5",
+    "postcss-loader": "^2.0.6",
     "prop-types": "^15.5.10",
     "react-bootstrap": "^0.31.1",
-    "react-datepicker": "^0.53.0",
-    "react-hot-loader": "^1.3.1",
+    "react-datepicker": "^0.54.0",
+    "react-hot-loader": "^3.0.0-beta.7",
```
xiaofan2406: v3 has a different API for hot reloading.
The `Main.jsx` should be updated according

config/postcss.config.js
```diff
@@ -0,0 +1,19 @@
+const autoprefixer = require('autoprefixer');
+
+const SUPPORTED_BROWSERS = [
```
xiaofan2406: I think you forgot to use `supported-browsers.js`

omgaz: Yup. Thanks, I split it out so when we upgrade browsers the git diff and file names is relevant and independent of postcss config changes.

package.json
omgaz: Will read through https://github.com/gaearon/react-hot-loader/tree/master/docs#migration-to-30

config/common.js
omgaz: :)

webpack.config.js
omgaz: Will leave as is.

package.json
omgaz: done

src/components/run.jsx
```diff
@@ -1,7 +1,25 @@
...
+  <AppContainer>
+    <App />
+  </AppContainer>,
+  document.getElementById('app')
+);
+
+// Hot Module Replacement API
+if (module.hot) {
+  module.hot.accept('./Main', () => {
+    const NextApp = require('./Main').default; // eslint-disable-line
```
xiaofan2406: The require here is not need anymore, because we are using webpack 3. Its tree shaking knows to resolve modules bindings.

You can do smt like this:
```
 const render = () => {
    ReactDOM.render(
      <AppContainer>
        <App />
      </AppContainer>,
      rootElement
    );
  };

  render();

  if (module.hot) {
    module.hot.accept('./Main', render);
  }
```

.babelrc
sonhanguyen: I think this is the better way as we care more about supporting browsers and can update on that basis. The downside though is that it's not clear from this config what version of the spec are we at.

xiaofan2406: It only adds three addition features that are already finished proposal. I don't see why we shouldn't be taking advantage of them. 
@sonhanguyen the spec is always http://babeljs.io/docs/plugins/preset-latest/

omgaz: Hugo will enjoy his Async Functions.

https://medium.com/komenco/what-to-expect-from-javascript-es2017-the-async-edition-618e28819711

src/index.html
```diff
@@ -11,7 +11,8 @@
 <body>
   <div id="app">APPLICATION CONTENT</div>
 
-  <script>__REACT_DEVTOOLS_GLOBAL_HOOK__ = parent.__REACT_DEVTOOLS_GLOBAL_HOOK__</script>
+  <script src="https://unpkg.com/react@15/dist/react.min.js"></script>
+  <script src="https://unpkg.com/react-dom@15/dist/react-dom.min.js"></script>
```
xiaofan2406: Since this is for development, this should include the `react.js` file, so that we get more warnings and better errors

omgaz: Ah thanks, I think I'll even try to load it locally anyhow.

xiaofan2406:
Merging 🚀

src/components/adslotUi/AlertInput/styles.scss
mdotwills: I actually like this distinction

PR #602, opened 2017-09-18
Update: update search component to allow value to be controlled by props

lightbringer1991:
## Current behaviour
`searchValue` is completely controlled internally via state, and therefore we cannot clear value on demand. The only way to clear value is triggering `onClear` function.
This causes problems when applying `Search` into `Treepicker`, since we would like to remove searchString when expanding nodes, breadcrumb clicks, etc.

## Changes
- remove internal state of `Search` component
- add `defaultValue` into `input`
- add Search demo component

### DIFF
--- a/src/components/adslotUi/Search/example.jsx
+++ b/src/components/adslotUi/Search/example.jsx
@@ -1,0 +1,42 @@
+import React, { Component } from 'react';
+import Search from './';
+
+export default class SearchExample extends Component {
+  constructor() {
+    super();
+    this.state = {
+      value: '',
+    };
+
+    this.onChange = this.onChange.bind(this);
+    this.onClear = this.onClear.bind(this);
+  }
+
+  onChange(value) {
+    this.setState({ value });
+  }
+
+  onClear() {
+    this.setState({ value: '' });
+  }
+
+  render() {
+    const props = {
+      placeholder: 'Cities',
+      onChange: this.onChange,
+      onClear: this.onClear,
+      value: this.state.value,
+    };
+
+    return (
+      <div>
+        <h2>Search</h2>
+        <div className="row">
+          <div className="col-xs-12">
+            <Search {...props} />
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
--- a/src/components/adslotUi/Search/index.jsx
+++ b/src/components/adslotUi/Search/index.jsx
@@ -8,8 +8,9 @@ import './styles.scss';
 export default class Search extends Component {
   constructor(props) {
     super(props);
-    this.state = { value: props.value };
     this.debounceOnSearch = _.debounce(props.onSearch, props.debounceInterval);
+    this.defaultValue = props.value;
+
     this.onChange = this.onChange.bind(this);
     this.onClear = this.onClear.bind(this);
     this.onKeyPress = this.onKeyPress.bind(this);
@@ -20,10 +21,7 @@ export default class Search extends Component {
     if (disabled) return;
 
     const value = _.get(event, 'target.value');
-
-    this.setState({ value });
     onChange(value);
-
     if (searchOnChange) this.debounceOnSearch(value);
   }
 
@@ -32,7 +30,8 @@ export default class Search extends Component {
     if (disabled) return;
 
     if (searchOnEnterKey && event.which === 13) {
-      onSearch(this.state.value);
+      const value = _.get(event, 'target.value');
+      onSearch(value);
     }
   }
 
@@ -42,7 +41,6 @@ export default class Search extends Component {
 
     const value = '';
 
-    this.setState({ value });
     onChange(value);
     if (searchOnChange) {
       onSearch(value);
@@ -51,7 +49,7 @@ export default class Search extends Component {
   }
 
   render() {
-    const { disabled, isLoading, placeholder, svgSymbolCancel, svgSymbolSearch } = this.props;
+    const { disabled, isLoading, placeholder, svgSymbolCancel, svgSymbolSearch, value } = this.props;
     const searchClassSuffixes = disabled ? ['color-disabled'] : svgSymbolSearch.classSuffixes;
     const cancelClassSuffixes = disabled ? ['color-disabled'] : svgSymbolCancel.classSuffixes;
 
@@ -66,10 +64,11 @@ export default class Search extends Component {
           onKeyPress={this.onKeyPress}
           placeholder={`Search ${placeholder}`}
           type="search"
-          value={this.state.value}
+          value={value}
+          defaultValue={this.defaultValue}
         />
         {isLoading ? <Spinner size="small" /> : null}
-        {_.isEmpty(this.state.value)
+        {_.isEmpty(value)
           ? <SvgSymbol href={svgSymbolSearch.href} classSuffixes={searchClassSuffixes} />
           : <SvgSymbol href={svgSymbolCancel.href} classSuffixes={cancelClassSuffixes} onClick={this.onClear} />
         }
--- a/src/components/adslotUi/Search/index.spec.jsx
+++ b/src/components/adslotUi/Search/index.spec.jsx
@@ -5,10 +5,9 @@ import { shallow } from 'enzyme';
 import React from 'react';
 import SvgSymbol from 'components/alexandria/SvgSymbol/index';
 import Spinner from 'components/alexandria/Spinner/index';
-import Search from '.';
+import Search from './';
 
 describe('Search', () => {
-  const testFunction = _.noop;
   let sandbox = null;
   let props = {};
 
@@ -17,7 +16,7 @@ describe('Search', () => {
   });
 
   beforeEach(() => {
-    props = { value: 'some value', onClear: testFunction, onChange: testFunction, onSearch: testFunction };
+    props = { onClear: _.noop, onChange: _.noop, onSearch: _.noop };
     sandbox.spy(props, 'onClear');
     sandbox.spy(props, 'onChange');
     sandbox.spy(props, 'onSearch');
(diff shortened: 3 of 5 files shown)


src/components/adslotUi/Search/index.jsx
```diff
@@ -93,6 +94,8 @@ Search.propTypes = {
   searchOnChange: PropTypes.bool,
   searchOnEnterKey: PropTypes.bool,
   debounceInterval: PropTypes.number,
+  // if true, allow search value to be controlled internally, otherwise value is controlled via props
+  selfControl: PropTypes.bool,
```
lightbringer1991: AI uprising 🤖

lightbringer1991:
[(3) Audience Survey: picker search](https://trello.com/c/56O1Qp4R/6836-3-audience-survey-picker-search)

src/components/adslotUi/Search/index.jsx
sonhanguyen: can we just make it uncontrolled when `value` is present?

mdotwills: `isControlled` is better imo

mdotwills: @sonhanguyen maybe a default prop for `isControlled` based on `value`?

lightbringer1991: @sonhanguyen was thinking about it before but for some reason I discarded it
I'll give it another try

sonhanguyen: `selfControl` feels too "creative" to me, hb `isControlled` as how this kind of comp is referred to in react docs.

src/components/adslotUi/Search/example.jsx
```diff
@@ -0,0 +1,68 @@
...
+        <div className="row">
+          <div className="col-xs-6">
+            <Search {...props} ref={(component) => { this.searchState = component; }} />
+          </div>
+          <div className="col-xs-6">
+            <Button bsStyle="primary" onClick={this.clearPropValue}>
+              Clear Value on Props
+            </Button>
+            <div>Prop value: {this.state.value}</div>
+            <div>Search internal state: {JSON.stringify(this.searchState.state)}</div>
```
omgaz: Can we remove this.

src/components/adslotUi/Search/example.jsx
```diff
@@ -0,0 +1,68 @@
...
+        </div>
+        <div className="row">
+          <div className="col-xs-6">
+            <Search {...props} ref={(component) => { this.searchState = component; }} />
+          </div>
+          <div className="col-xs-6">
+            <Button bsStyle="primary" onClick={this.clearPropValue}>
+              Clear Value on Props
+            </Button>
+            <div>Prop value: {this.state.value}</div>
```
omgaz: Can we remove this.

src/components/adslotUi/Search/example.jsx
```diff
@@ -0,0 +1,68 @@
...
+          </Checkbox>
+        </div>
+        <div className="row">
+          <div className="col-xs-6">
+            <Search {...props} ref={(component) => { this.searchState = component; }} />
+          </div>
+          <div className="col-xs-6">
+            <Button bsStyle="primary" onClick={this.clearPropValue}>
+              Clear Value on Props
+            </Button>
```
omgaz: We'd not use a button like this to clear the value. What's the use case, can we show a better example of when and why this would be used to show intent. Trying to get some control on adslot ui to show meaning and intent. Would this be a `Reset Form` or `Reset Search` button. It'd also be an inverse button most likely.

src/components/adslotUi/Search/example.jsx
omgaz: Again, moving Adslot UI examples to not show anything we'd not do ourselves and to have the examples not look like a developer sandbox. Things like this are great during feature development but we're in the process of making this a living and rich example of how to do things.

src/components/adslotUi/Search/example.jsx
lightbringer1991: the intention is to show when the search field is uncontrolled, then clearing `value` prop does not remove the text displayed in the search input (since it's controlled by the internal state). So it's not really a clear/reset button.

I'm not sure what to change into. Could you please advise.

src/components/adslotUi/Search/index.jsx
```diff
@@ -51,9 +51,10 @@ export default class Search extends Component {
   }
 
   render() {
-    const { disabled, isLoading, placeholder, svgSymbolCancel, svgSymbolSearch } = this.props;
+    const { disabled, isLoading, placeholder, svgSymbolCancel, svgSymbolSearch, value } = this.props;
     const searchClassSuffixes = disabled ? ['color-disabled'] : svgSymbolSearch.classSuffixes;
     const cancelClassSuffixes = disabled ? ['color-disabled'] : svgSymbolCancel.classSuffixes;
+    const searchValue = _.isNil(value) ? this.state.value : value;
```
omgaz: I feel this could confuse things. Would it not make sense, if `value` changes, in `componentWillReceiveProps` to update the internal state using setState.

src/components/adslotUi/Search/example.jsx
omgaz: Sure thing, not a criticism we're moving things in a different direction and you're just the first ~~guinea pig~~ volunteer :)

src/components/adslotUi/Search/index.jsx
```diff
@@ -8,13 +8,19 @@ import './styles.scss';
...
+    this.state = { value: '' };
     this.debounceOnSearch = _.debounce(props.onSearch, props.debounceInterval);
     this.onChange = this.onChange.bind(this);
     this.onClear = this.onClear.bind(this);
     this.onKeyPress = this.onKeyPress.bind(this);
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (!_.isNull(nextProps.value)) {
+      this.setState({ value: nextProps.value });
```
xiaofan2406: You are essentially proxy props to state. 

I dont quite understand the purpose of this PR. I thought it already does what you are trying to do according to your description

src/components/adslotUi/Search/index.spec.jsx
```diff
@@ -105,6 +105,29 @@ describe('Search', () => {
...
+      const inputEl = component.find('input');
+      expect(inputEl.prop('value')).to.equal('needle');
+
+      const svgSymbolEl = component.find(SvgSymbol);
+      expect(svgSymbolEl.prop('href')).to.equal('/assets/svg-symbols.svg#cancel');
+      expect(svgSymbolEl.prop('classSuffixes')).to.deep.equal(['gray-darker']);
+    });
+
+    it('should allow props to control value string', () => {
+      const component = shallow(<Search {...props} />, { lifecycleExperimental: true });
```
omgaz: Just had to google this one :)

```
setProps() will not trigger the lifecycle method componentDidUpdate when using shallow() unless you set lifecycleExperimental to true.
```

src/components/adslotUi/Search/index.jsx
lightbringer1991: you cannot change the text value outside of `Search` component at the moment, so to be able to change value, you have to implement/trigger `onChange` or `onClear` function.

In our tree picker we want to clear out the text when expanding nodes, and there is currently no option to do so, since the value is internal managed.

src/components/adslotUi/Search/index.spec.jsx
lightbringer1991: I think it's a nice touch to test certain lifecycle methods without the need of using `mount`

src/components/adslotUi/Search/index.jsx
sonhanguyen: @xiaofan2406 this props to state thing was suggested in previous discussion because the component needs to support both controlled and uncontrolled way, I guess it's simpler to only deal with the state internally, as opposed to do `props.value || state.value` everywhere.

@lightbringer1991 instead of `isNull` which I suppose is your way to determine 'controlled' or 'uncontrolled', may be simply check if the prop is different from the  current prop. That way value can be cleared from outside.

lightbringer1991: @sonhanguyen I would like to have the prop controls the value whenever it's present, that's why the `_.isNull` is there. It is similar to what you suggested.

omgaz: A positive point towards HOC and having every component stateless with stateful wrappers. If you want to control the state you use the stateless component otherwise the wrapper. @sonhanguyen do you think this is a use case for this pattern?

omgaz:
LGTM besides the discussion, do we have any problems with the component handling both?

src/components/adslotUi/Search/index.jsx
sonhanguyen: yeah that'll be good separation of concern but I would try to be careful not to over-engineer stuff like this.
About your comment on being both controlled and uncontrolled. Imo, it's not necessary a bad thing, react-dom implements all `input` like that.

xiaofan2406: When prop changes, the component would have already be re-rendered with the new prop, whats the value of setting it into state again?

The problem here isn't the prop part. It is the state. We dont need state here at all. I mentioned it previously on @Maxmus1989 's PR when this component was touched. But that PR was too big to change so much.
If you look at all the places it reads `this.state.value`, and they can be all replaced with `event.target.value`.

@lightbringer1991 With the current `Search` implementation, eg if you render `<Search value={someVar} />`, you can change the value of the input by changing `someVar`. Did I misunderstood something?

sonhanguyen: I think @xiaofan2406 got a point here, may be if possible looking into simplify this and remove the state altogether?

omgaz: ^good point. I'd agree that we can trash the state completely.

lightbringer1991: ok I'll remove the internal state, this component will no longer be able to self-controlled after that though.

sonhanguyen: it will, if you don't pass a `value`, the `input` will maintain its own state. Should add `defaultValue` for that case like suggested.

src/components/adslotUi/Search/index.jsx
```diff
@@ -20,19 +19,16 @@ export default class Search extends Component {
...
-
-    this.setState({ value });
     onChange(value);
-
     if (searchOnChange) this.debounceOnSearch(value);
   }
 
   onKeyPress(event) {
-    const { disabled, searchOnEnterKey, onSearch } = this.props;
+    const { disabled, searchOnEnterKey, onSearch, value } = this.props;
```
sonhanguyen: can you read value from `target` instead. Just to have a consistent source of truth. it'll work both in controlled and uncontrolled manner.

src/components/adslotUi/Search/index.jsx
```diff
@@ -66,10 +61,10 @@ export default class Search extends Component {
           onKeyPress={this.onKeyPress}
           placeholder={`Search ${placeholder}`}
           type="search"
-          value={this.state.value}
+          value={value}
```
sonhanguyen: `defaultValue` as well? depending on your need but I think it's nice to have.

src/components/adslotUi/Search/index.jsx
lightbringer1991: this event is only for the enter key, so the value to be searched will be the one from props anyway.

src/components/adslotUi/Search/index.jsx
lightbringer1991: I have added the default value

src/components/adslotUi/Search/index.jsx
sonhanguyen: do you mean enter key doesn't have target, or its target is not `input`?
what should happen if you do:
```
    <Search {...provideNoValue} />
```
type some stuff and enter, will search be triggered with empty query then, since `value` is undefined?

src/components/adslotUi/Search/example.jsx
```diff
@@ -0,0 +1,42 @@
+import React, { Component } from 'react';
+import Search from './';
+
+export default class AlertInputExample extends Component {
```
omgaz: !`AlertInputExample`

hehe

omgaz:
Merging :ramen:

PR #612, opened 2017-09-26
Breaking: Create an individual css asset file per chunk

omgaz:
ExtractTextPlugin seems to override each chunk with the next one so we
were only exporting the extras css under the guise of main.

Fixed, but introducing a potential breaking change.

- adslot-ui.css is not adslot-ui-main.css


lightbringer1991:
merging...

PR #616, opened 2017-09-27
Fix: firefox input overflow on flex

lightbringer1991:
Flex and Input does not play well with Firefox, see screenshots

Firefox: 
screen shot 2017-09-27 at 4 39 07 pm

Chrome:
screen shot 2017-09-27 at 4 39 11 pm

Fix: wrap input inside a wrapper element

### DIFF
--- a/src/components/adslot-ui/AlertInput/index.jsx
+++ b/src/components/adslot-ui/AlertInput/index.jsx
@@ -87,17 +87,19 @@ export default class AlertInput extends Component {
         onMouseLeave={this.handleMouseLeave}
       >
         {prefixAddon ? <span className={`${baseClass}-addon`}>{prefixAddon}</span> : null}
-        <input
-          className={`${baseClass}-input`}
-          type={type}
-          defaultValue={defaultValue}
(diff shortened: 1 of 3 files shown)


omgaz:
LGTM!

src/components/adslot-ui/AlertInput/index.jsx
```diff
@@ -87,17 +87,19 @@ export default class AlertInput extends Component {
...
-          type={type}
-          defaultValue={defaultValue}
-          value={value}
-          min={min}
-          placeholder={placeholder}
-          onChange={onValueChange}
-          onFocus={this.handleInputFocus}
-          onBlur={this.handleInputBlur}
-        />
+        <span className={`${baseClass}-flex-wrapper`}>
```
omgaz: Any reason we can't make the outside `<div` also the wrapper?

lightbringer1991: the outside div is the flex container. The issue happens when the input flex item is not rendered correctly in Firefox

PR #621, opened 2017-10-02
Update: Search component now supports self-control behaviour

lightbringer1991:
Search component now works if `value` is not passed in as prop.

Tested on the following scenarios:
- `value` is passed from props. Internal state is updated to prop value accordingly.
- `value` is NOT passed from props. Internal state controls the input value.

### DIFF
--- a/src/components/adslot-ui/Search/index.jsx
+++ b/src/components/adslot-ui/Search/index.jsx
@@ -11,10 +11,19 @@ class Search extends Component {
   constructor(props) {
     super(props);
     this.debounceOnSearch = _.debounce(props.onSearch, props.debounceInterval);
+    this.state = {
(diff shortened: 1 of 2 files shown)


src/components/adslot-ui/Search/index.jsx
```diff
@@ -11,17 +11,27 @@ class Search extends Component {
   constructor(props) {
     super(props);
     this.debounceOnSearch = _.debounce(props.onSearch, props.debounceInterval);
+    this.state = {
```
sonhanguyen: didn't we agree that state was not needed in a recent [PR](https://github.com/Adslot/adslot-ui/pull/602), and that this behaviour was already achieved ?

lightbringer1991:
close as this has been discussed before

PR #622, opened 2017-10-03
Fix: Remove deprecated svg xlinkhref

lteacher:
Apparently the attr is deprecated. I noticed some issues with svg partial rendering and changing to `href` seemed to fix those issues so figured perhaps we should just upgrade this.

[Here is a link to the MDN page with the big red thingo saying its deprecated](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use)

image
image

### DIFF
--- a/src/components/alexandria/SvgSymbol/index.jsx
+++ b/src/components/alexandria/SvgSymbol/index.jsx
@@ -11,7 +11,7 @@ const SvgSymbol = (props) => {
 
   return (
     <svg className={`${componentClass}${classesList}`} onClick={onClick}>
-      <use xlinkHref={href} />
+      <use href={href} />
     </svg>
   );
 };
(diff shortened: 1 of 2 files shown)


vinteo:
did you wait for October to do this @lteacher ?

lteacher:
Lolz, actually it just worked out like that :shirt: honest...

Was looking at why my little package icons are all messed up
image

:hand: :shirt: :cocktail:

omgaz:
Was waiting on a fix on SVG4Everybody (which is a little behind). Someone fixed this  last November https://github.com/jonathantneal/svg4everybody/pull/79/files

omgaz:
LGTM

omgaz:
Merging 🍜

sonhanguyen:
@lteacher https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/href
looks like `href` is still not supported in safari

lteacher:
@sonhanguyen :-1: Safari, the new IE

PR #623, opened 2017-10-03
Fix: Add missing babel plugin

lteacher:
This plugin is missing locally and causes errors which means project doesn't build.

Not sure how some people are actually running this without the dep, maybe its installed globally for people?

Heres a little snippettz
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "babel": "^6.23.0",
     "babel-core": "^6.26.0",
     "babel-loader": "^7.1.2",
+    "babel-plugin-transform-react-jsx": "^6.24.1",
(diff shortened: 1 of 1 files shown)


omgaz:
Hmm that is odd. The amount of times I've trashed my node modules and it's been fine. I've not had any errors besides hot-reloading seems to to not work. I wonder if this will fix it.

I guess I got a little trigger-happy on my NPM package cleanup.

omgaz:
LGTM 🍥

lteacher:
Hmm yeah, well I set it as `Fix` I guess it doesnt actually affect any versioning but mleh.. :shirt: :dash: :man_cartwheeling:

omgaz:
Merging 🍜

Issue #625, opened 2017-10-06
Replace react/require-extension rule


lteacher:
Apparently the above rule is deprecated per the message below

_'The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.'_

PR #627, opened 2017-10-06
Upgrade: upgrade all eslint packages

lightbringer1991:
Upgrade the eslint packages to the highest possible version (due to peer dependencies)
```bash
$ npm info "eslint-config-airbnb@latest" peerDependencies
 
{ eslint: '^4.9.0',
  'eslint-plugin-import': '^2.7.0',
  'eslint-plugin-jsx-a11y': '^6.0.2',
  'eslint-plugin-react': '^7.4.0' }
```

This upgrade will also fix https://github.com/Adslot/adslot-ui/issues/625


src/components/adslot-ui/ListPicker/index.jsx
```diff
@@ -20,14 +20,16 @@ const isSubset = (array, subArray) =>
...
       'applyAction',
       'cancelAction',
       'deselectItem',
       'getApplyButtonState',
       'loadData',
       'selectItem',
-    ]) { this[methodName] = this[methodName].bind(this); }
+    ].forEach((methodName) => {
+      this[methodName] = this[methodName].bind(this);
+    });
```
lightbringer1991: http://airbnb.io/javascript/#iterators--nope

src/components/adslot-ui/Panel/index.jsx
```diff
@@ -29,7 +29,7 @@ PanelComponent.propTypes = {
   title: PropTypes.node.isRequired,
   isCollapsed: PropTypes.bool,
   onClick: PropTypes.func.isRequired,
-  children: PropTypes.any,
+  children: PropTypes.node,
```
lightbringer1991: as per PropTypes docs https://reactjs.org/docs/typechecking-with-proptypes.html

> // Anything that can be rendered: numbers, strings, elements or an array
  // (or fragment) containing these types.

src/components/adslot-ui/SpinnerButton/index.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/prop-types */
```
lightbringer1991: disable this rule for this file since we combine Button props with this component props using `_.assign`. Since `eslint` cannot evaluate the results, errors are thrown

src/components/alexandria/Avatar/index.jsx
```diff
@@ -13,7 +13,7 @@ const Avatar = ({ color, givenName, tooltip, image, surname }) => (
     {image ? <img className={`${baseClass}-image`} src={image} role="presentation" /> : null}
 
     <div className="avatar-component-initials">
-      {`${_.first(givenName) || ''}${_.first(surname) || ''}`}
+      {`${_.head(givenName) || ''}${_.head(surname) || ''}`}
```
lightbringer1991: https://github.com/wix/eslint-plugin-lodash/blob/master/docs/rules/preferred-alias.md

src/components/adslot-ui/TreePicker/Node/index.spec.jsx
```diff
@@ -112,9 +112,15 @@ describe('TreePickerNodeComponent', () => {
     let fireCount = 0;
     const testFunction = () => { fireCount += 1; };
 
-    const component = mount(<TreePickerNode
-      itemType={itemType} node={cbrNode} removeNode={testFunction} selected disabled
-    />);
+    const component = mount(
+      <TreePickerNode
+        itemType={itemType}
```
lightbringer1991: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-max-props-per-line.md

src/components/adslot-ui/Panel/index.jsx
sonhanguyen: just want to note that people seem to be enjoying the children as a function pattern recently

lightbringer1991: would it be listed under `elements`? since ReactComponent can be a function

sonhanguyen: element is an object that has a `type` referring to a component, which can be a function. Anw, It was a passing by thought, not that there is anything wrong with your change.

package.json
```diff
@@ -55,14 +55,14 @@
     "codecov.io": "^0.1.6",
     "css-loader": "^0.28.4",
     "enzyme": "^2.9.1",
-    "eslint": "^3.19.0",
-    "eslint-config-airbnb": "^10.0.1",
+    "eslint": "^4.3.0",
+    "eslint-config-airbnb": "^15.1.0",
```
xiaofan2406: @lightbringer1991 bad timing. It is 16 now.

src/components/adslot-ui/ListPickerPure/index.jsx
```diff
@@ -52,14 +52,15 @@ const ListPickerPureComponent = ({
...
+              : null
+            }
           </GridRow>
         </Grid> :
         null
       }
       <div className="listpickerpure-component-items">
         <Grid>
-          {_.map(items, (item) =>
+          {_.map(items, (item) => (
```
vinteo: is there a rule for this change? Not sure I think this is better

xiaofan2406: jsx-wrap-multilines

lightbringer1991: I'll probably just disable this rule. Was a bit reluctant to disable this one since it seems to be logical.

package.json
lightbringer1991: hmm I'll update then
release over the weekend lol

src/components/adslot-ui/ConfirmModal/index.jsx
```diff
@@ -38,20 +38,19 @@ const ConfirmModalComponent = ({
 ConfirmModalComponent.displayName = 'AdslotUiConfirmModalComponent';
 
 ConfirmModalComponent.propTypes = {
-  buttonCancelLabel: PropTypes.string.isRequired,
-  buttonConfirmLabel: PropTypes.string.isRequired,
-  modalApply: PropTypes.func.isRequired,
+  buttonCancelLabel: PropTypes.string,
```
lightbringer1991: @omgaz @sonhanguyen @nimishjha `eslint` throws error saying that the required props are having default value
should we enforce the rule so if a prop is required, then it must not have a default value? 

The default value is useless for required props since user needs to pass in the props anyway.

nimishjha: Yeah, that makes sense. Although I'd phrase it as if a prop has a default value, it doesn't need to be ```required```.

.eslintrc
```diff
@@ -18,7 +20,7 @@
   ],
   "settings": {
     "lodash": {
-      "version": 3
+      "version": 4
```
omgaz: 👍

.eslintrc
```diff
@@ -68,6 +72,20 @@
...
+    "function-paren-newline": 0,
+    "no-plusplus": 0,
+    "prefer-destructuring": 0,
+    "comma-dangle": ["error", {
+      "arrays": "only-multiline",
+      "objects": "only-multiline",
+      "imports": "only-multiline",
+      "exports": "only-multiline",
+      "functions": "never"
+    }]
```
omgaz: 👍

src/components/adslot-ui/ConfirmModal/index.jsx
omgaz: I've always been in two minds about this, but happy to go with `If default prop provided then property is optional`

src/components/alexandria/Breadcrumb/index.jsx
```diff
@@ -23,7 +23,7 @@ const Breadcrumb = ({ nodes, onClick, disabled }) => {
       {
         _.map(nodes, (node, index) =>
           <span className={`${baseClass}-node`} key={node.id}>
-            <span className={`${baseClass}-node-divider`}> > </span>
+            <span className={`${baseClass}-node-divider`}> &gt; </span>
```
lightbringer1991: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md

src/components/adslot-ui/ListPicker/index.jsx
xiaofan2406: I think this is the worst rule that airbnb decided to enforce.

nimishjha: Yeah, I love my for loops

sonhanguyen: I actually like this. Only downside is performance which hopefully will be addressed by adding some fp specific optimisation into V8 at some point in the future. Having multiple ways of doing the same thing sucks (for, while, comprehension, monads), if i have to pick one I pick monads too, it's just a more powerful construct and the flow is clearer.

mdotwills: I prefer this rule actually also

omgaz:
LGTMerging 🍜 🍥 

Can we create a ticket this for alignment with adslot product?

PR #628, opened 2017-10-06
Update: option to run single test & disable coverage

sonhanguyen:
To run a specific test file:
`$ npm run test[:watch] --file=<absolute path or path relative to project root>`
watch mode has coverage off by default, to turn it on:
`$ npm run test:watch --coverage`

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,4 +1,17 @@
 const webpackConfig = require('./webpack.config');
+const webpack = require('webpack');
+const _ = require('lodash');
+
+const env = { ADSLOT_TEST_FILE: '' };
+
+webpackConfig.plugins.push(
+  new webpack.DefinePlugin(
+    _(env)
+      .assign(_.pick(process.env, _.keys(env)))
+      .mapValues(JSON.stringify)
+      .value()
+  )
+);
 
 module.exports = function configureKarma(config) {
   config.set({
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -33,7 +33,10 @@ module.exports = function configureKarma(config) {
       mocha: {},
     },
     singleRun: true,
-    reporters: ['coverage', 'mocha'],
+    reporters: _.compact([
+      process.env.npm_config_coverage ? 'coverage' : null,
+      'mocha'
+    ]),
     preprocessors: {
       'config/loadtests.js': ['webpack'],
     },
--- a/scripts/test
+++ b/scripts/test
@@ -2,38 +2,30 @@
 
 const path = require('path');
 const { execSync } = require('child_process');
+let {
+  coverage, file = process.env.npm_config_file, watch
+} = require('minimist')(process.argv.slice(2));
+if (coverage) process.env.npm_config_coverage = coverage;
 
-const TASK = 2;
-const argv = require('minimist')(process.argv.slice(TASK + 1));
-let { file = argv._.pop() } = argv;
 if (file) {
-  if (!path.isAbsolute(file)) {
-    file = execSync('npm root').toString().replace('node_modules', file).trim()
-  }
+  const projectRoot = execSync('npm root').toString().replace('node_modules', '').trim()
+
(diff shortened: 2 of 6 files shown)


scripts/test
```diff
@@ -0,0 +1,23 @@
...
+
+let env = '';
+const { f, file = f || argv[0] } = argv;
+if (file) env = `ADSLOT_TEST_FILE=${file} `;
+
+/* has to be done this way as karma removed the cli option to override config
+ https://stackoverflow.com/questions/29150998/karma-running-a-single-test-file-from-command-line
+ */
+
+ // npm run test [watch] -- [--file=] <absolute path>
```
sonhanguyen: not sure how to get rid of the first double dash

karma.conf.js
```diff
@@ -1,4 +1,17 @@
...
+const webpack = require('webpack');
+const _ = require('lodash');
+
+const env = { ADSLOT_TEST_FILE: '' };
+
+webpackConfig.plugins.push(
+  new webpack.DefinePlugin(
+    _(env)
+      .assign(_.pick(process.env, _.keys(env)))
+      .mapValues(JSON.stringify)
```
sonhanguyen: the value will be inlined so it has to be defined like `{ ADSLOT_TEST_FILE: '"/some/path"' }`

scripts/test
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
+const TASK = 2;
+const argv = require('minimist')(process.argv.slice(TASK));
+console.log(argv)
```
vinteo: remove?

scripts/test
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
+const TASK = 2;
+const argv = require('minimist')(process.argv.slice(TASK));
```
vinteo: does this need to be added as a dev dependency?

sonhanguyen: it's already there

lightbringer1991: interesting reads
https://npmcompare.com/compare/commander,minimist,nomnom,optimist,yargs

sonhanguyen: there is also a lib called `vorpal` for interactive shell

scripts/test
```diff
@@ -0,0 +1,33 @@
...
+const path = require('path');
+const argv = require('minimist')(process.argv.slice(TASK));
+const { execSync } = require('child_process');
+
+let { file = argv.f || argv._.pop() } = argv;
+if (file) {
+  if (!path.isAbsolute(file)) {
+    file = execSync('npm root').toString().replace('node_modules', file).trim()
+  }
+  console.info('\x1b[4mRunning test:\x1b[0m\x1b[1m ', file, '\x1b[0m');
```
omgaz:
```
const colorGreen = '\x1b[4m';
console.info(`${colorGreen}Running test:\x1b[0m\x1b[1m ', file, '\x1b[0m`); 
```

lightbringer1991:
looking great.
Maybe include an instruction on how to run a single test and/or no coverage

sonhanguyen:
because of the way npm passes argument, it's annoying to use named argument like `npm run test -- --file=path`, which is why I made the `--file` part to be optional.

I settle with it but for reference, there is another approach which is #3fb38b7 which has a drawback of making the script dependent on npm to parse argument which means you can just run `node scripts/test --argument`.

sonhanguyen:
@lightbringer1991 there is a separate branch for coverage off but I'm not sure it's really needed.

lteacher:
I usually just use `only` with watch on, but lgtm :shirt:

lightbringer1991:
@lteacher `.only` doesn't work for me, not sure why. That's why I proposed this change.

omgaz:
LGTM

sonhanguyen:
per feedback, added coverage flag

omgaz:
LGTMerging 🍜

PR #629, opened 2017-10-06
Change pure functions to pure components

### LINKED ISSUE #610: Make all stateless functional components PureComponents
- Stateless (functional) components are no faster than the stateful (class)
- Rendering in React 15 is roughly 25% faster compared to 0.14
- Pure components are the fastest. Use shouldComponentUpdate
- Rendering in development mode is 2-8x slower than rendering in production
- Rendering in development with React 15 is about 2x slower than 0.14

https://moduscreate.com/react_component_rendering_performance/

Also, props are read-only in stateless functional components. We should be using React Dev Tools more. See https://github.com/facebook/react-devtools/issues/314#issuecomment-327843862

lteacher:
Changes to use PureComponent

Resolves #610


src/components/adslot-ui/ConfirmModal/index.jsx
```diff
@@ -1,39 +1,47 @@
...
-      <Modal.Footer>
-        {modalClose ? <Button className="btn-inverse" onClick={cancelAction} data-test-selector="confirm-modal-cancel">
-          {buttonCancelLabel}</Button> : null}
-        <Button bsStyle="primary" onClick={applyAction} data-test-selector="confirm-modal-confirm">
-          {buttonConfirmLabel}</Button>
-      </Modal.Footer>
-    </Modal>
-  );
-};
+  render() {
```
lightbringer1991: seems a little bit inconsistent here yea?
maybe `render = () => {`

lightbringer1991:
I can't make the es6 syntax works with instrumenter either
is it possible to just stick with the old method now and not using arrow function?

The plugin you're using doesn't generate coverage report, so I think that's not what we want

https://github.com/istanbuljs/babel-plugin-istanbul
> Note: This plugin does not generate any report or save any data to any file; it only adds instrumenting code to your JavaScript source code. To integrate with testing tools, please see the Integrations section.

lteacher:
The report is generated through karma

src/components/adslot-ui/ConfirmModal/index.jsx
lteacher: Hmm yeah didn't think of it had just used traditional es6 method and did use short form if didnt need a body but... maybe could use that form instead for consistency?

.babelrc
```diff
@@ -19,6 +19,17 @@
   "plugins": [
     "transform-strict-mode",
     "transform-react-jsx",
+    "transform-class-properties",
```
lteacher: I added this to allow for class properties which removes the need to use `.bind` or anon delegate style etc...

[Here is a link to babel docs for the proposal](https://babeljs.io/docs/plugins/transform-class-properties/)

.babelrc
```diff
@@ -19,6 +19,17 @@
...
     "react-hot-loader/babel"
-  ]
+  ],
+  "env": {
+    "test": {
+      "plugins": [["istanbul", {
+        "exclude": [
+          "src/**/*.spec**",
+          "src/lib/**"
```
lteacher: This stuff is for the istanbul instrumenter via babel (`babel-istanbul`)

.eslintrc
```diff
@@ -1,4 +1,5 @@
 {
+  "parser": "babel-eslint",
```
lteacher: Allows eslint to understand fully babelified codez

.eslintrc
```diff
@@ -38,6 +39,7 @@
       "component": true,
       "html": false
     }],
+    "react/prefer-stateless-function": 0,
```
lteacher: Obviously we need to switch this off now.

config/test.js
```diff
@@ -30,16 +30,6 @@ module.exports = merge(commonConfig, {
...
-
-      {
-        test: jsRegEx,
-        include: srcPath,
-        exclude: /src\/lib/,
-        use: {
-          loader: 'istanbul-instrumenter-loader',
-          options: { esModules: true },
-        },
-      },
```
lteacher: Removed the instrumenter loader it was replaced by `babel-istanbul`

package.json
```diff
@@ -22,7 +22,7 @@
     "start": "npm run start:dev",
     "start:dev": "webpack-dev-server --config=./config/dev.js",
     "start:cold": "node server.js --env=dev-cold",
-    "test": "karma start",
+    "test": "NODE_ENV=test karma start",
```
lteacher: Test env is here to hit the env setting in the `.babelrc`

xiaofan2406:
Personally, I dont think we should do this. PureComponent just adds a shallow equality check for all props. It can essentially be much more slower.

src/components/adslot-ui/ConfirmModal/index.jsx
xiaofan2406: render and lifecycles by default had `this` binded already.

src/components/adslot-ui/Carousel/index.jsx
```diff
@@ -47,16 +47,18 @@ export const getNextDecorator = () => {
...
 
-  return (
-    <Carousel decorators={decorators} {...props} className={classNames(baseClass, className)}>
-      {children}
-    </Carousel>
-  );
-};
+    return (
+      <Carousel decorators={decorators} {...this.props} className={classNames(baseClass, className)}>
+        {children}
```
nutboltu: If we declare the parent component as Pure component and one of the children has a state which is updated by another state(e.g: local state), can it be triggered as well?
https://60devs.com/pure-component-in-react.html
This article suggests not to put Pure component in the parent.

src/components/adslot-ui/ConfirmModal/index.jsx
sonhanguyen: @lightbringer1991 that's a good point, I blame javascript, in coffee they are all arrows

src/components/adslot-ui/Carousel/index.jsx
sonhanguyen: just to be clear
> When a pure component has children, all child components that depend on a change in this.context will not re-render when the context is changed, unless the parent PureComponent declares contextTypes

if we do use `context` then this should be taken in account (we don't in `direct-web`). Although from the gh issue linked this seems like a mother instance of the `context` API incomplete rather than `PureComponent` being broken.

sonhanguyen:
just adding my 2 cents, not to invalid the effort but the article in the original issue @omgaz created dated 2016, since then quite a few people have written about `PureComponent` being premature optimisation that can turn out not as performant. Also the benchmark in the article seems to be with a dummy `shouldComponentUpdate`, not with the official `PureComponent`

src/components/adslot-ui/Carousel/index.jsx
lteacher: Yeah I went through the issue a bit as well and it seems like maybe should remove the wrap for components with children I guess just in case some are context aware for now...

lteacher:
Hmm I didn't look into it too much but just went from the issues for Friday whirlwind... Most seem to reference [this random tweet](https://twitter.com/dan_abramov/status/755343749983657986). Obviously tweets are the source of truth for all things, but I couldn't quickly scan through the react source to see if anyone is familiar with where such logic occurs?

Do you some updated resources to reference for the below quote so I can take a look (maybe can link onto the issue instead for the discussion)
> _...since then quite a few people have written about PureComponent being premature optimisation that can turn out not as performant..._

lightbringer1991:
To me, it seems logical that no checks < checks

sonhanguyen:
@lteacher don't really have any concrete benchmark, just this medium article which is slightly more legitimate than a tweet I guess https://blog.shakacode.com/react-purecomponent-pitfalls-d057882f4b6e

sonhanguyen:
@lightbringer1991 depend on the actual component. The idea is that `shouldComponentUpdate` can save unnecessary `render` but it gets called more than `render` and sometimes not cheap enough to offset the cost of the wasted render cycles it blocks.

omgaz:
A lot of my reasoning for PureComponents within AdslotUI seems different from the direction this conversation has taken which I feel more adequately suited to product implementation and discussion in Adslot. Remembering this is a component library to enforce UX principles and consistency in the UI. I don't intend for this to change nor influence integration of these components or implementation of other components outside this component library.

Having full React Component classes supported gives us a consistent way of declaring components (always extends React.Component whether Pure or otherwise). Having supported React classes that are part of the core will mean support and improvements from the core.

Being able to access improved debugging in React DevTools (especially the much noted prop type inspections and modifications) is a massive advantage for the component library, front end, and ux team that is mostly unseen by the full-stacks. Being able to adjust and modify props on the fly, especially during design review, product updates, planning meetings and research is highly valuable.

It's re-assuring to see such concern for performance but it's a lot of speculation on case by case basis. There's clear proof that SFC provide performance gains and RPCs can add unnecessary, and potentially false-positive, shallow equals.

If React is giving us a 30% increase over AngularJS, and PureComponents over functional costs us even 1% then we still have a vast gain in improved performance.

I would rather switch to stateless functional components on a case-by-case basis than the other way round and for developers to be mindful of the potential hazards of each. There are times, such as treepicker nodes, where we need the power of fastStatelessWrapper to do an explicit prop check. There will be places where we know props will never change because we deem them static read-only props. I'd feel as though these are the cases to be mindful of when, and if, we encounter performance issues.

lteacher:
@omgaz Is this still desired? Not sure there, there was a bunch of discussion around it so let me know. Its becoming pretty out of date now.

omgaz:
I think so. Bit as Xiaofan pointed out we should probably review (as a follow up) if any of them should be downgraded to `Component` and and manually add a `shouldComponentUpdate`

lteacher:
Out of date

Issue #613, opened 2017-09-26
Make OpenBrowser optional when starting dev environment


lightbringer1991:
In Linux server version, there is no graphic library and therefore browsers cannot be installed.
`npm start` will throw error when it tries to open the default browser. Error details below:

```bash
Error: Command failed: /var/www/nodejs/adslot-ui/node_modules/open/vendor/xdg-open "http://localhost:8888/webpack-dev-server/"
xdg-open: no method available for opening 'http://localhost:8888/webpack-dev-server/'

    at ChildProcess.exithandler (child_process.js:198:12)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at maybeClose (internal/child_process.js:920:16)
    at Socket.<anonymous> (internal/child_process.js:351:11)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at Pipe._handle.close [as _onclose] (net.js:497:12)
```

Proposal: add an option to NOT open the browser if needed, so by default it will still do what it does now
e.g. `OPEN_BROWSER=false npm start`

vinteo:
you can use `npm run start:cold`

lightbringer1991:
@vinteo I think `start:cold` is broken atm, since the actual command is `"start:cold": "node server.js --env=dev-cold"`, and there is no `server.js`

I think it should be changed to 
`"start:cold": "webpack-dev-server --config=./config/dev-cold.js"`
(with some changes in `dev-cold.js`, since it's not working atm)

Also it doesn't provide hot reload, which I guess why it's called `cold`. However I would like to still have hot reload.

This is what I setup for devevelopment:
- Ubuntu 16.04 server in VMWare (no GUI)
- Webstorm in Windows host, editing the code via SMB/samba

so basically all code editing/troubleshooting is from Windows, and interact with the node server remotely(ish). That's why I'd like hot reload as well, just don't open a browser when starting dev env.

However I think this can be fixed via the deprecation of `OpenBrowserPlugin`, as stated in https://github.com/Adslot/adslot-ui/issues/614

Issue #614, opened 2017-09-26
Deprecate `OpenBrowserPlugin`


lightbringer1991:
As per https://webpack.github.io/docs/webpack-dev-server.html

> --open: opens the url in default browser (for webpack-dev-server versions > 2.0).

We only need to add `open` into `webpack-dev-server` config to open the browser

If this one is approved and fixed then https://github.com/Adslot/adslot-ui/issues/613 is also fixed, since it auto resolves if browser cannot be opened
```bash
Unable to open browser: http://localhost:8888/webpack-dev-server/. If you are running in a headless environment, please do not use the open flag.
```

lightbringer1991:
I can't make the es6 syntax works with instrumenter either
is it possible to just stick with the old method now and not using arrow function?

The plugin you're using doesn't generate coverage report, so I think that's not what we want

https://github.com/istanbuljs/babel-plugin-istanbul
> Note: This plugin does not generate any report or save any data to any file; it only adds instrumenting code to your JavaScript source code. To integrate with testing tools, please see the Integrations section.

PR #630, opened 2017-10-08
Update: deprecate OpenBrowserPlugin

lightbringer1991:
Fix https://github.com/Adslot/adslot-ui/issues/613 and https://github.com/Adslot/adslot-ui/issues/614

As per https://webpack.github.io/docs/webpack-dev-server.html
> --open: opens the url in default browser (for webpack-dev-server versions > 2.0).

We don't need this plugin anymore.


omgaz:
Merging 🍜

PR #636, opened 2017-10-18
New: AlertInput onFocus prop addition

adityasabnis:
Add `onFocus` propType function to `AlertInput`

Resolves: https://github.com/Adslot/adslot-ui/issues/637

capture

### DIFF
--- a/src/components/adslot-ui/AlertInput/index.jsx
+++ b/src/components/adslot-ui/AlertInput/index.jsx
@@ -46,6 +46,10 @@ export default class AlertInput extends Component {
       isFocused: true,
       isPopoverVisible: Boolean(this.props.alertMessage),
     });
+
+    if (this.props.onFocus) {
+      this.props.onFocus(event);
+    }
   }
 
   handleInputBlur(event) {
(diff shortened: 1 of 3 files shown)


src/components/adslot-ui/AlertInput/index.jsx
```diff
@@ -46,6 +46,10 @@ export default class AlertInput extends Component {
       isFocused: true,
       isPopoverVisible: Boolean(this.props.alertMessage),
     });
+
+    if (this.props.onFocus) {
+      this.props.onFocus(event);
```
lightbringer1991: probably just default to `_.noop` so you don't have to make this check

omgaz: I disagree, I think this check is valid and preferable to a noop for consistency with other types and for providing explicitness with the optional parameters.

adityasabnis:
Discussion outcome for moving to noop:  Stick to existing.

mdotwills:
Merged.. ⏰

PR #639, opened 2017-10-20
New: Add hover dropdown component

lightbringer1991:
after

### DIFF
--- a/docs/components/Layout/index.jsx
+++ b/docs/components/Layout/index.jsx
@@ -46,6 +46,7 @@ import TreePickerExample from '../../examples/TreePickerExample';
 import UserListPickerExample from '../../examples/UserListPickerExample';
 import InformationBoxExample from '../../examples/InformationBoxExample';
 import SplitPaneExample from '../../examples/SplitPaneExample';
+import HoverDropdownMenuExample from '../../examples/HoverDropdownMenuExample';
 
 import { PageTitle } from '../../../src/dist-entry';
 
@@ -87,6 +88,7 @@ const componentsBySection = {
   navigation: [
     'breadcrumb',
     'tab',
+    'hover-dropdown-menu',
   ],
   'feedback-and-states': [
     'alert',
@@ -175,6 +177,7 @@ class PageLayout extends React.Component {
             <PageTitle title="Navigation" />
             <BreadcrumbExample />
             <TabExample />
+            <HoverDropdownMenuExample />
 
             <PageTitle title="Feedback and States" />
             <AlertExample />
--- a/docs/examples/HoverDropdownMenuExample.jsx
+++ b/docs/examples/HoverDropdownMenuExample.jsx
@@ -1,0 +1,113 @@
+import _ from 'lodash';
+import React from 'react';
+import Example from '../components/Example';
+import {
+  Avatar,
+  HoverDropdownMenu,
+} from '../../src/dist-entry';
+
+
+class HoverDropdownMenuExample extends React.PureComponent {
+  render() {
+    const links = [
+      {
+        title: 'Adslot.com',
+        url: 'http://www.adslot.com',
+        target: '_self',
+        isEnabled: true,
+        onClick: _.noop,
+      },
+      {
+        title: 'Logout',
+        url: '#',
+        target: '_modal',
+        isEnabled: true,
+        onClick: _.noop,
+      },
+    ];
+
+    const props = {
+      arrowPosition: 'left',
+      headerText: 'Hello, John Smith',
+      hoverComponent: (<Avatar givenName="John" surname="Smith" />),
+    };
+
+    return (
+      <div className="hover-dropdown-example">
+        <HoverDropdownMenu {...props}>
+          {_.map(links, (link) => (
+            <HoverDropdownMenu.Item key={link.title} {...link} />
+          ))}
+        </HoverDropdownMenu>
+      </div>
+    );
+  }
+}
+
+
+const exampleProps = {
+  componentName: 'Hover Dropdown Menu',
+  notes: '',
+  exampleCodeSnippet: `
+const links = [
+  {
+    title: 'Adslot.com',
+    url: 'http://www.adslot.com',
+    target: '_self',
+    isEnabled: true,
+    onClick: _.noop,
+  },
+  {
+    title: 'Logout',
+    url: '#',
+    target: '_modal',
+    isEnabled: true,
+    onClick: _.noop,
+  },
+];
+
+const props = {
+  arrowPosition: 'left',
+  headerText: 'Hello, John Smith',
+  hoverComponent: (<Avatar givenName="John" surname="Smith" />),
+};
+
+<div className="hover-dropdown-example">
+  <HoverDropdownMenu {...props}>
+    {_.map(links, (link) => (
+      <HoverDropdownMenu.Item key={link.title} {...link} />
+    ))}
+  </HoverDropdownMenu>
+</div>
+`,
+  propTypes: [
+    {
+      propType: 'arrowPosition',
+      type: "oneOf ['left', 'right']",
+      defaultValue: 'left',
+      note: 'determine the placement of the popover',
+    }, {
+      propType: 'headerText',
+      type: 'string',
+      defaultValue: '',
+      note: 'If set to empty string, header will not be rendered.',
+    }, {
+      propType: 'links',
+      type: "arrayOf {oneOf ['_self', '_modal']: target, string: title, string: url, bool: isEnabled}",
+      defaultValue: '[ ]',
+      note: 'Each link will be used to render dropdown item',
+    }, {
+      propType: 'hoverComponent',
+      type: 'node',
+      note: 'displayed element to be hovered on, e.g. Avatar component.',
+    }, {
+      propType: 'onLinkClick',
+      type: 'func',
+      note: 'onLinkClick(link), callback when user clicks on a dropdown item; link is an object of `links` prop',
+      defaultValue: '_.noop',
+    },
+  ],
+};
+
+
+export default () => <Example {...exampleProps}><HoverDropdownMenuExample /></Example>;
--- a/src/components/adslot-ui/HoverDropdownMenu/PopoverLinkItem/index.jsx
+++ b/src/components/adslot-ui/HoverDropdownMenu/PopoverLinkItem/index.jsx
@@ -1,0 +1,49 @@
+/* eslint-disable react/prop-types */
+import _ from 'lodash';
+import React, { PropTypes } from 'react';
+import { Button } from 'react-bootstrap';
+import './styles.scss';
+
+
+class PopoverLinkItemComponent extends React.PureComponent {
+  render() {
+    const { target, title, url, isEnabled, onClick } = this.props;
+
+    const buttonProps = {
+      disabled: !isEnabled,
+      onClick,
+      bsStyle: 'link',
+    };
+
+    if (target !== '_modal') {
+      _.assign(buttonProps, { href: url });
+    }
+
+    if (target === '_blank') {
+      _.assign(buttonProps, { rel: 'noopener noreferrer' });
+    }
+
+    return (
+      <li className="popover-link-item">
+        <Button {...buttonProps}>{title}</Button>
+      </li>
+    );
+  }
+}
+
+export const LINK_PROPS = {
+  target: PropTypes.oneOf(['_blank', '_self', '_modal']),
+  title: PropTypes.string.isRequired,
+  url: PropTypes.string,
+  isEnabled: PropTypes.bool,
+};
+
+PopoverLinkItemComponent.propTypes = _.assign({ onClick: PropTypes.func }, LINK_PROPS);
+
+PopoverLinkItemComponent.defaultProps = {
+  target: '_self',
+  isEnabled: true,
+  onClick: _.noop,
+};
+
+export default PopoverLinkItemComponent;
--- a/src/components/adslot-ui/HoverDropdownMenu/index.jsx
+++ b/src/components/adslot-ui/HoverDropdownMenu/index.jsx
@@ -1,0 +1,127 @@
+import _ from 'lodash';
+import React, { PropTypes } from 'react';
+import { Overlay, Popover } from 'react-bootstrap';
+import PopoverLinkItem from './PopoverLinkItem';
+import './styles.scss';
+
+// if change this, change the width in styles.scss too
+const POPOVER_WIDTH = 160;
+
+/* eslint-disable react/prop-types */
+export const renderPopoverComponent = (arrowPosition) => (props) => {
+  // default offset is 50%, therefore to adjust to align with the arrow, we need to move back/forth 30%
+  const popoverOffset = POPOVER_WIDTH * 0.3;
+  const popoverProps = _.assign({}, props, {
+    arrowOffsetLeft: arrowPosition === 'left' ? '20%' : '80%',
+    style: {
+      left: arrowPosition === 'left' ? props.style.left + popoverOffset : props.style.left - popoverOffset,
+    },
+  });
+
+  return (<Popover {...popoverProps} />);
+};
+/* eslint-enable react/prop-types */
+
+
+export class HoverDropdownMenuComponent extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isOpen: false,
+      target: null,
+      mouseInPopover: false,
+    };
+
+    this.closeMenu = _.debounce(() => {
+      if (!this.state.mouseInPopover) {
+        this.setState({
+          isOpen: false,
+        });
+      }
+    }, 100);
+
+    this.openMenu = this.openMenu.bind(this);
+    this.popoverEnterHandler = this.popoverEnterHandler.bind(this);
+    this.popoverLeaveHandler = this.popoverLeaveHandler.bind(this);
+  }
+
+  componentDidMount() {
+    // prevent default title popup if exists, assuming the first child is the hoverComponent
+    this.element.childNodes[0].removeAttribute('title');
+  }
+
+  openMenu(event) {
+    this.setState({
+      isOpen: true,
+      target: event.target,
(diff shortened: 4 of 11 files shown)


src/components/adslot-ui/HoverDropdown/PopoverLinkItem/index.jsx
```diff
@@ -0,0 +1,47 @@
...
+    if (link.target !== '_modal') {
+      _.assign(listGroupItemProps, { href: link.url });
+    }
+
+    return (<ListGroupItem {...listGroupItemProps}>{link.title}</ListGroupItem>);
+  }
+}
+
+PopoverLinkItem.propTypes = {
+  link: PropTypes.shape(DROPDOWN_MENU_ITEM_PROPTYPES).isRequired,
```
lightbringer1991: I'm tempted to flatten this prop, but just afraid that symphony needs to change the code quite a bit
thoughts @omgaz ?

docs/examples/HoverDropdownExample.jsx
```diff
@@ -0,0 +1,182 @@
...
+    this.state = {
+      isModalOpen: false,
+    };
+
+    this.closeModal = this.closeModal.bind(this);
+  }
+
+  closeModal() {
+    this.setState({ isModalOpen: false });
+  }
```
omgaz: I don't think this modal adds anything to the example. An example use-case would more likely be page navigation. A `noop` will suffice.

docs/examples/HoverDropdownExample.jsx
```diff
@@ -0,0 +1,182 @@
...
+        {
+          title: 'Open Modal',
+          url: 'http://www.google.com',
+          target: '_modal',
+          group: '',
+          isEnabled: true,
+          childNavigationLinks: [],
+        },
+      ],
+    };
```
omgaz: Just one example.

docs/examples/HoverDropdownExample.jsx
omgaz: The modal also muddies the example and detracts from the context of what this component is responsible for.

docs/examples/HoverDropdownExample.jsx
lightbringer1991: ok I'll remove the second set
was putting it there to view the rendered popover left and right orientation

docs/examples/HoverDropdownExample.jsx
```diff
@@ -0,0 +1,182 @@
...
+      note: 'If set to empty string, header will not be rendered.',
+    }, {
+      propType: 'links',
+      type: "arrayOf {oneOf ['_self', '_modal']: target, string: title, string: url, bool: isEnabled}",
+      defaultValue: '[ ]',
+      note: 'Each link will be used to render dropdown item',
+    }, {
+      propType: 'hoverComponent',
+      type: 'node',
+      note: 'displayed element to be hovered on',
```
omgaz: `… e.g. Avatar component.`

src/dist-entry/core.js
```diff
@@ -127,4 +128,5 @@ export {
   TreePickerNode,
   TreePickerSimplePure,
   UserListPicker,
+  HoverDropdown,
```
omgaz: Maybe `HoverDropdownMenu` is more intutive that links and actions are to be expected.

src/components/adslot-ui/HoverDropdown/styles.scss
```diff
@@ -0,0 +1,21 @@
+@import '~styles/variable';
+
+$dropdown-width: 160px;
+
+.hover-dropdown {
+  display: inline-block;
+}
+
+.hover-dropdown-popover {
```
omgaz: Can we nest this as to only have one root per file e.g.

```scss
.hover-dropdown {
  display: inline-block;

  &-dropdown-popover {
    …
  }
}
```

src/components/adslot-ui/HoverDropdown/propTypes.js
```diff
@@ -0,0 +1,10 @@
+import { PropTypes } from 'react';
+
+const DROPDOWN_MENU_ITEM_PROPTYPES = {
+  target: PropTypes.oneOf(['_self', '_modal']),  // not sure if there are any other target type
```
omgaz: _blank

also if `_blank` then we should be adding `rel="noopener noreferrer"` to the `<a>`

src/components/adslot-ui/HoverDropdown/propTypes.js
```diff
@@ -0,0 +1,10 @@
...
+  title: PropTypes.string,
+  url: PropTypes.string,
```
omgaz: Maybe optionally a url or an onClick (there may be case for changing page for a single page app) in which case we probably shouldn't be using `<a>` but `<Button bsStyle='link'>`; which would also, conveniently, save us some extra css with spacing / padding etc...

src/components/adslot-ui/HoverDropdown/PopoverLinkItem/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+      onClick: this.onLinkItemClick,
+    };
+
+    if (link.target !== '_modal') {
+      _.assign(anchorProps, { href: link.url });
+    }
+
+    return (
+      <li className="popover-link-item">
+        <a {...anchorProps}>{link.title}</a>
```
omgaz: See notes on bootstrap styling

src/components/adslot-ui/HoverDropdown/PopoverLinkItem/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+    const { link, onClick } = this.props;
+    onClick(link);
+  }
+
+  render() {
+    const { link } = this.props;
+
+    const anchorProps = {
+      disabled: !link.isEnabled,
+      onClick: this.onLinkItemClick,
```
omgaz: onClick shouldn't be set on an anchor tag, use a `<Button bsStyle='link'>`

src/components/adslot-ui/HoverDropdown/PopoverLinkItem/styles.scss
```diff
@@ -0,0 +1,20 @@
+@import '~styles/color';
+
+.popover-link-item {
+  list-style-type: none;
```
omgaz: Bootstrap should do a lot of this for us with `list-unstyled` on the `<ul>`

src/components/adslot-ui/HoverDropdown/PopoverLinkItem/styles.scss
```diff
@@ -0,0 +1,20 @@
...
+    text-decoration: none;
+    color: inherit;
+
+    &:hover {
+      text-decoration: none;
+    }
+
+    &:active {
+      text-decoration: none;
+    }
```
omgaz: Bootstrap should do a lot of this for us with `bsStyle='link' / className='btn btn-link'` on the `<button> / <a>`

src/components/adslot-ui/HoverDropdown/PopoverLinkItem/styles.scss
```diff
@@ -0,0 +1,20 @@
...
+  }
+}
```
omgaz: I don't think any of this file is needed with the bootstrap replacements.

src/components/adslot-ui/HoverDropdown/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+            <Popover
+              id="hover-dropdown-popover"
+              className="hover-dropdown-popover"
+              onMouseEnter={this.popoverEnterHandler}
+              onMouseLeave={this.popoverLeaveHandler}
+              title={headerText}
+              arrowOffsetLeft={arrowPosition === 'left' ? '20%' : '80%'}
+              positionLeft="20%"
+            >
+              <ListGroup>
```
omgaz: What does this do? Can we not just use a <ul className="list-unstyled">?

src/components/adslot-ui/HoverDropdown/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+            placement="bottom"
+            onEntering={this.onEntering}
+          >
+            <Popover
+              id="hover-dropdown-popover"
+              className="hover-dropdown-popover"
+              onMouseEnter={this.popoverEnterHandler}
+              onMouseLeave={this.popoverLeaveHandler}
+              title={headerText}
+              arrowOffsetLeft={arrowPosition === 'left' ? '20%' : '80%'}
```
omgaz: nice :)

src/components/adslot-ui/HoverDropdown/PopoverLinkItem/index.jsx
omgaz: We'll get them to review it as well. I think flattening it and we should be ok.

docs/examples/HoverDropdownExample.jsx
omgaz: Yea, it's good to sandbox this stuff and visually verify that everything's good! Thanks for thinking about both use cases.

src/components/adslot-ui/HoverDropdown/PopoverLinkItem/index.jsx
lightbringer1991: it's reused in the parent component, since it uses an array of `link` as a prop

src/components/adslot-ui/HoverDropdown/index.jsx
lightbringer1991: this actually doesn't work with `Overlay`, so I need to implement custom solution

docs/examples/HoverDropdownMenuExample.jsx
```diff
@@ -0,0 +1,100 @@
...
+  Avatar,
+  HoverDropdownMenu,
+} from '../../src/dist-entry';
+
+
+class HoverDropdownMenuExample extends React.PureComponent {
+  render() {
+    const props = {
+      arrowPosition: 'left',
+      headerText: 'header',
```
omgaz: Can this be changed to match the example: `Hello, John Smith`

docs/examples/HoverDropdownMenuExample.jsx
```diff
@@ -0,0 +1,100 @@
...
+      onLinkClick: _.noop,
+      hoverComponent: (<Avatar givenName="John" surname="Smith" tooltip="test tooltip" />),
```
omgaz: Can the tooltip be removed, with this example it's not needed.

docs/examples/HoverDropdownMenuExample.jsx
```diff
@@ -0,0 +1,100 @@
...
+      links: [
+        {
+          title: 'Adslot.com',
+          url: 'http://www.adslot.com',
+          target: '_self',
+          isEnabled: true,
+        },
+        {
+          title: 'Open Modal',
+          url: 'http://www.google.com',
```
omgaz: Opening google in a modal doesn't really make sense. Maybe change this to be logout with a _.noop

src/components/adslot-ui/HoverDropdownMenu/styles.scss
```diff
@@ -0,0 +1,17 @@
+@import '~styles/variable';
+
+$dropdown-width: 160px;
```
omgaz: Global vars are exported. If we don't mean for this scope it inside `.hover-dropdown`

omgaz:
The header should also have consistent spacing to the list items.

See github's example:

screen shot 2017-11-03 at 10 29 52

src/components/adslot-ui/HoverDropdownMenu/styles.scss
lightbringer1991: I think we can scope it for now, I don't see any other use case for this variable yet.

src/components/adslot-ui/HoverDropdownMenu/index.jsx
```diff
@@ -0,0 +1,143 @@
...
+
+      </div>
+    );
+  }
+}
+
+HoverDropdownMenuComponent.propTypes = {
+  arrowPosition: PropTypes.oneOf(['left', 'right']),
+  headerText: PropTypes.string,
+  links: PropTypes.arrayOf(PropTypes.shape(LINK_PROPS)),
```
lightbringer1991: I'd like to have the `onClick` handler inside the link item itself, right now it uses a global `onLinkClick` function

what do you think @omgaz

src/components/adslot-ui/HoverDropdownMenu/PopoverLinkItem/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+    return (
+      <li className="popover-link-item">
+        <Button {...buttonProps}>{title}</Button>
+      </li>
+    );
+  }
+}
+
+export const LINK_PROPS = {
```
sonhanguyen: need a consensus on constant naming convention.
I think it's suitable for primitives and enums, not so for objects because they are mutable (and also for some reason it's not a convention in the js community)

src/components/adslot-ui/HoverDropdownMenu/index.jsx
sonhanguyen: how about instead of passing some object that this menu component needs to know the shape of, you can simply pass children that are element of the component that already define the shape and already has `onClick` in it?
```
// ...
<ul className="list-unstyled">
    {children.map(element =>
        React.cloneElement(element, {
           ...element.props,
           onClick() {
              element.props.onClick()
              this.onLinkClick()
           }
        })
    )} 
// ...
children: PropTypes.arrayOf(PropTypes.shape(
   type: PopoverLinkItem
   props: PropTypes.shape({
       onClick: PropTypes.func
       // do not care about the rest of the props
   })
))
```

sonhanguyen: since the elements already have `onClick`, you won't need to expose `onLinkClick` as a prop anymore

xiaofan2406:
Just curious, whats the context of this PR?

I feel this component isn't necessary. we are passing massive json objects to specify what to render, rather than laying out the components themselves.

It gets very difficult to maintain as well. If say later, we want to render a `input` inside the popover menu, for the current implementation, , we need to change `link_props`, and the `onLinkClick` etc and the API for this components would get even bigger.

Maybe we need
- a `Popover` component that can knows what to render inside the popover and handles the hover action to show its content
- a `Menu` component that can knows how to render its children. Basically a very dump component purely include styles. Ofc we can add shortcut to simplify the `MenuItem` components such as `onItemClick` if all the item has the same handler.
- a `MenuItem` component that deals with the action of a menu item. It could be anything, a link, a button, a text or a input that knows what to do with itself.

and then to achieve this component's usage, we would just render a bunch of `MenuItem`s in the `Menu` and then pass in `Menu` as the content of the `Popover`

src/components/adslot-ui/HoverDropdownMenu/PopoverLinkItem/index.jsx
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable react/prop-types */
+import _ from 'lodash';
+import React, { PropTypes } from 'react';
+import { Button } from 'react-bootstrap';
+import './styles.scss';
+
+
+class PopoverLinkItemComponent extends React.PureComponent {
```
xiaofan2406: imagine some place in our angular code we have `scope.buttonProps` and that gets passed in to a `PopoverLinkItemComponent `. and then we do `scope.buttonProps.disable = true` on some condition. wouldn't this component fail to re-render?

lightbringer1991: `buttonProps` is not a prop for this component, it is composed using various props, so I don't think it's gonna be confused even when you pass in the same variable name manually.
Besides, this component is not exported to be used.

docs/components/Layout/index.jsx
```diff
@@ -87,6 +88,7 @@ const componentsBySection = {
   navigation: [
     'breadcrumb',
     'tab',
+    'hover-dropdown',
```
omgaz: hover-dropdown-menu

docs/examples/HoverDropdownMenuExample.jsx
```diff
@@ -0,0 +1,100 @@
...
+      <div className="hover-dropdown-example">
+        <HoverDropdownMenu {...props} />
+      </div>
+    );
+  }
+}
+
+
+const exampleProps = {
+  componentName: 'HoverDropdownMenu',
```
omgaz: Hover Dropdown Menu

src/components/adslot-ui/HoverDropdownMenu/styles.scss
```diff
@@ -0,0 +1,27 @@
...
+      ul {
+        margin: 0;
+      }
+
+      .btn {
+        padding: 0;
+      }
+    }
+  }
+}
```
omgaz: Add `$popover-title-bg: $color-background` to `src/styles/variable.scss` under the popover section. We want to maintain white backgrounds for headers.

src/components/adslot-ui/HoverDropdownMenu/index.jsx
```diff
@@ -0,0 +1,143 @@
...
+    const newLeft = this.props.arrowPosition === 'left' ? left + offset : left - offset;
+
+    _.assign(element.style, {
+      left: `${newLeft}px`,
+    });
+
+    // alter arrow position to left-oriented or right-oriented
+    _.assign(element.children[0].style, {
+      left: this.props.arrowPosition === 'left' ? '20%' : '80%',
+    });
```
omgaz: Why aren't we using the supported `arrowOffsetLeft` on the popoover?

https://react-bootstrap.github.io/components.html#popover-props

omgaz:
Also, wiggle the mouse left and right over the item and it bugs out.

bugs

omgaz:
And each line needs a background hover.

Something like:

```
  li:hover {
    background-color: $color-gray-white;
  }
```

PR #640, opened 2017-10-22
Update: 3rd Party Button Component with AdslotUI Wrapper

omgaz:
Starting with the `button` component this PR demonstrates the preferred and proposed way of wrapping and exporting supported third-party components.

It allows us to properly document, test and support features and opinions as part of adslot ui directly - such as data-test-selectors.

### Changes
Create a wrapper component for Bootstrap's Button Component
Allows DTS
Supports 'inverse' buttons
Support disabled buttons to have a 'reason'

### Screenos
Support for disabled reasons
screen shot 2017-12-08 at 17 26 39

### DIFF
--- a/src/components/third-party/bootstrap/Button/index.jsx
+++ b/src/components/third-party/bootstrap/Button/index.jsx
@@ -1,0 +1,62 @@
+/* eslint-disable react/prop-types */
+// disable proptypes check because it doens't take into consideration extended types
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import BootstrapButton from 'react-bootstrap/lib/Button';
+import BootstrapPopover from 'react-bootstrap/lib/Popover';
+import BootstrapOverlayTrigger from 'react-bootstrap/lib/OverlayTrigger';
+import { expandDts } from 'lib/utils';
+
+
+class Button extends React.PureComponent {
+  renderWithReason() {
+    const popover = (<BootstrapPopover
+      id="btn-reason"
+      className="btn-popover-reason"
+    >{this.props.reason}</BootstrapPopover>);
+    return (
+      <BootstrapOverlayTrigger trigger={['focus', 'hover']} placement="bottom" overlay={popover}>
+        {this.renderButton()}
+      </BootstrapOverlayTrigger>
+    );
+  }
+
+  renderButton() {
+    const {
+      inverse,
+      children,
+      dts,
+      className,
+    } = this.props;
+
+    return (
+      <BootstrapButton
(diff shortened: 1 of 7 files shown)


src/components/third-party/bootstrap/Button/index.jsx
```diff
@@ -0,0 +1,40 @@
...
+    const {
+      inverse,
+      children,
+      dts,
+    } = this.props;
+
+    return (
+      <BootstrapButton
+        {..._.pick(this.props, _.keys(BootstrapButton.propTypes))}
+        className={classNames({ 'btn-inverse': inverse })}
```
pphminions: merge with props.classnames

src/components/third-party/bootstrap/Button/index.jsx
```diff
@@ -0,0 +1,62 @@
...
+      >{children}</BootstrapButton>
+    );
+  }
+
+  render() {
+    const {
+      disabled,
+      reason,
+    } = this.props;
+    return disabled && reason ? this.renderWithReason() : this.renderButton();
```
mdotwills: this ternary could be confusing, since you could be mislead into thinking `(disabled && reason) ?`. Could you please add some brackets here?

src/components/third-party/bootstrap/Button/index.spec.jsx
```diff
@@ -0,0 +1,49 @@
...
+    const element = shallow(<Button inverse>Test</Button>);
+    expect(element.prop('className')).to.equal('btn-inverse');
+  });
+
+  it('should support data-test-selectors', () => {
+    const element = shallow(<Button dts="test-button">Test</Button>);
+    expect(element.prop('data-test-selector')).to.equal('test-button');
+  });
+
+  it('should render disabled buttons without a reason popover', () => {
```
mdotwills: button~~s~~

src/components/third-party/bootstrap/Button/index.jsx
omgaz: @mdotwills Do you think it'd be made clearer through naming and adding a check on the the reason being valid.

```
  render() {
    const hasReason = this.props.reason.length > 0;
    const isDisabled = this.props.disabled;
    return isDisabled && hasReason ? this.renderWithReason() : this.renderButton();
  }
```

mdotwills: I don't really have an opinion over the naming here tbh, I think `(disabled && reason)` is short and sweet. It's more so those brackets that add value here.

omgaz: Actually, thinking about it now. `''` is already false. so the length check will do nothing.

And to protect against `<Button reason>test</Button>` we already have prop types.

mdotwills:
Merging.. 🚀

PR #643, opened 2017-10-26
Update: use prettier and make eslint compatible

xiaofan2406:
- `prettier` and `eslint-config-prettier`

- there are only a few prettier config. Under `prettier` key in `package.json`

- using `lint-staged` to ensure all staged files will be `prettier`ed

- add script in `package.json` to allow CI to do check code style using `prettier`

- add a script `npm run prettier` to format all code

- this PR also changes the pre-commit hook. Instead of the script file, change it to use `husky`.

Editors: https://prettier.io/docs/en/editors.html


src/components/adslot-ui/TreePicker/Node/index.jsx
```diff
@@ -66,60 +66,72 @@ class TreePickerNodeComponent extends React.Component {
...
           <GridCell stretch {...labelCellProps} dts="label">
             <TextEllipsis>
               <span>{nodeRenderer(node)}</span>
-              {!_.isEmpty(pathElement) ?
-                <span className={`${baseClass}-metadata`}> ({pathPrefix(node)}{pathElement})</span> :
-                null
-              }
+              {!_.isEmpty(pathElement) ? (
+                <span className={`${baseClass}-metadata`}>
+                  &nbsp;({pathPrefix(node)}
```
xiaofan2406: here is the only code change. Prettier change it to 
```
{' ' }
({pathPrefix(node)}
{pathElement})
```
But it breaks many tests due the way tests are written using `wrapper.children()`

docs/components/Contributors/index.jsx
```diff
@@ -14,43 +10,51 @@ class Contributors extends React.Component {
...
 
-
   getContributors() {
-    return new Promise((resolve) => {
-      fetch('https://api.github.com/repos/Adslot/adslot-ui/contributors').then(resolve);
-    }).then((response) => response.json());
+    return new Promise(resolve => {
+      fetch('https://api.github.com/repos/Adslot/adslot-ui/contributors').then(
+        resolve
+      );
```
omgaz: this looks really horrible :/

docs/components/Contributors/index.jsx
```diff
@@ -14,43 +10,51 @@ class Contributors extends React.Component {
...
+      this.state.contributors,
+      ({ login, avatar_url, contributions }) => (
+        <Avatar
+          givenName={login}
+          image={avatar_url} // eslint-disable-line camelcase
+          key={login}
+          tooltip={`Thanks, ${login}, for your ${contributions} contribution${contributions >
+          1
+            ? 's'
+            : ''}!`}
```
omgaz: I can't even make sense of this one :P Probably a case for moving this logic away from inline.

src/components/adslot-ui/TreePicker/Node/index.jsx
omgaz: I always feel like these spaces here are hacks anyway. That space is purely presentation and should be handled by the css with a `.treepickernode-component-metadata:before { content: ' ' }`

Like my other comments, if something looks weird it's probably because there's a better way to solve it.

docs/components/Contributors/index.jsx
xiaofan2406: yeah, there are some edge case looks super weird

omgaz: Maybe some brackets would help keep the `(contributions > 1)` together

omgaz:
LGTM, I'd consider changing to css on that nbsp.

Some things look odd, but oh well. Probably better ways of doing those parts anyway.

Appreciate this :D

src/components/adslot-ui/TreePicker/Node/index.jsx
xiaofan2406: agreed.

docs/examples/AlertInputExample.jsx
```diff
@@ -82,42 +80,56 @@ const exampleProps = {
       propType: 'defaultValue',
       type: 'string',
       defaultValue: 'text',
-    }, {
+    },
+    {
```
lteacher: Hmnm dunno about this one either a few places like this pointlessly expanding objects out.

src/components/adslot-ui/ListPickerPure/index.spec.jsx
```diff
@@ -39,18 +41,36 @@ describe('ListPickerPureComponent', () => {
...
 
     const gridHeaderCellElements = gridHeaderElement.find(GridCell);
-    expect(gridHeaderCellElements.first().children().text()).to.equal('Team');
-    expect(gridHeaderCellElements.last().children().text()).to.equal('Member');
+    expect(
+      gridHeaderCellElements
+        .first()
+        .children()
+        .text()
+    ).to.equal('Team');
```
lteacher: :broken_heart:

sonhanguyen:
I wasn't a fan of too much tooling but for this the work has been done by someone else and I'm happy to adopt :)

xiaofan2406:
@sonhanguyen it s really not much work. These formatting are done thru 1 command. and that the purpose of this PR, stop worrying about formatting.

docs/examples/ButtonExample.jsx
```diff
@@ -1,43 +1,76 @@
...
-    note: 'onClick(event)',
-  }, {
-    propType: 'bsSize',
-    type: 'string',
-    note: <span>We rarely use <pre>large</pre>, and should not use <pre>small</pre>.</span>,
-  }],
+  notes: (
+    <p>
+      See{' '}
+      <a
```
omgaz: look

pphminions:
LGTM 🍯  🦋

xiaofan2406:
@omgaz it is recommended strongly by prettier to use 80 as the line width https://prettier.io/docs/en/options.html#print-width

Thoughts?

omgaz:
@xiaofan2406 Scrolling is easy, so I don't mind. Whilst I'd like to keep consistency between adslot-ui and direct-web that linked documentation suggests we may get undesirable results.

For now, I'd consider enforcing 120; if we run into these undesirable traits we can consider dropping down to 80.

What do you think?

xiaofan2406:
yep. That sound good.

omgaz:
Restarted the travis job, Ive seen it fail a few times like this

omgaz:
lgtm

scripts/git-hooks/pre-commit
```diff
@@ -1,47 +0,0 @@
-#!/bin/bash
```
bchew: Would there be any impact to those who already have the pre-commit git hook symlinked before? Or does husky handle this transparently?

omgaz: If so, would this then be handled as a `breaking change` and communicate it through release notes?

xiaofan2406: it would only affect the dev flow. so should nt be breaking change.

I haven't really tested it since I didnt have git-hooks setup before this. But according to https://github.com/typicode/husky/issues/120 I think husky will not overwrite it.

I guess we need to notify everyone to remove the existing git hooks

omgaz: Cool :) Will add a note in the release notes.

@bchew I'm happy to merge, if it's good with you then merge away!

bchew:
Merging :rocket:

PR #644, opened 2017-10-27
Fix: Fix linting errors

lightbringer1991:


src/components/adslot-ui/InformationBox/index.spec.jsx
```diff
@@ -4,11 +4,10 @@ import { shallow } from 'enzyme';
 
 describe('InformationBoxComponent', () => {
   it('should render with props', () => {
-    const component = shallow(<InformationBox
-      title="render title here"
-      icon="assets/img#done"
-      children="<div>I am child</div>"
```
lightbringer1991: `children` should not be a prop. See `react/no-children-prop`

lightbringer1991:
No report found because the codecov didn't run in master due to travis failure

src/components/adslot-ui/InformationBox/index.spec.jsx
sonhanguyen: I don't think it even worked

omgaz:
Not sure why code-cov is saying there's no base report (assume it's because of previous failures). But I'll merge this.

Issue #635, opened 2017-10-12
Docs: Component Prop Types Maintenance


omgaz:
## Adslot UI has been updated!

<img src="https://i.imgflip.com/10t87m.jpg" width="200px" />

It has navigation. It has order. It has prop type definitions.

![https://user-images.githubusercontent.com/908155/30812712-ddb14e66-a24e-11e7-9e49-ddc702a02ec1.png](https://user-images.githubusercontent.com/908155/30812712-ddb14e66-a24e-11e7-9e49-ddc702a02ec1.png)

### How can I help?

Not all the notes and prop types are accurate and complete. This is a great way to get some GitHub cred and open a couple of PRs to kill some time.

Pick an example file `/docs/examples` (excluding TreePickerExample.jsx), find the component. Fill out the prop types in the example file. Open a PR or two (or all in one). Check off the list.

 - [x] AccordionExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] AlertExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] AlertInputExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] AvatarExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] BorderedWellExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] BreadcrumbExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] ButtonExample.jsx
 - [x] CardExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] CarouselExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] CheckboxExample.jsx
 - [x] ConfirmModalExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] DatePickerExample.jsx
 - [x] EmptyExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] FilePickerExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] FlexibleSpacerExample.jsx
 - [x] GridExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
 - [x] HelpIconPopoverExample.jsx
 - [x] ListPickerExample.jsx (https://github.com/Adslot/adslot-ui/pull/645)
[...]

PR #645, opened 2017-10-27
Docs: add missing proptypes

lightbringer1991:
Update missing prop-types, as per https://github.com/Adslot/adslot-ui/issues/635

### DIFF
--- a/docs/examples/AlertInputExample.jsx
+++ b/docs/examples/AlertInputExample.jsx
@@ -81,10 +81,13 @@ const exampleProps = {
     {
       propType: 'defaultValue',
       type: 'string',
-      defaultValue: 'text',
     }, {
       propType: 'value',
-      type: "string|number",
+      type: 'string|number',
+    }, {
+      propType: 'type',
+      type: "oneOf: 'text', 'number'",
+      defaultValue: 'text',
     }, {
       propType: 'min',
       type: 'number',
@@ -98,6 +101,13 @@ const exampleProps = {
       propType: 'suffixAddon',
       type: 'node',
     }, {
+      propType: 'alertStatus',
+      type: "oneOf: 'success', 'info', 'warning', 'error'",
+      defaultValue: 'success',
+      note: <span>As <pre>success</pre> is assumed, and help is always displayed independently, the accepted pattern
+      is to only use <pre>warning</pre> and <pre>error</pre> feedback states with this component.
+      Otherwise leave type undefined for <pre>success</pre>.</span>,
+    }, {
       propType: 'alertMessage',
       type: 'string',
     }, {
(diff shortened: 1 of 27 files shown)


docs/examples/AlertInputExample.jsx
```diff
@@ -98,6 +101,9 @@ const exampleProps = {
       propType: 'suffixAddon',
       type: 'node',
     }, {
+      propType: 'alertStatus',
+      type: "oneOf: 'success', 'info', 'warning', 'error'",
```
omgaz: Can you add a note that success and info should generally not be used. A success state should be assumed as default. Info is rarely needed.

Maybe: `As success is assumed, and help is always displayed independently the accepted pattern is to only use warnings and error feedback states with this component.`

docs/examples/SpinnerButtonExample.jsx
```diff
@@ -30,17 +30,24 @@ class SpinnerButtonExample extends React.Component {
...
-    note: <span>When <pre>isLoading</pre> is <pre>true</pre> the button is disabled.</span>,
-  }, {
-    propType: 'onClick',
-    type: 'func',
-    note: <span>Should set <pre>isLoading</pre> to true, and reset to <pre>false</pre> upon completion.</span>,
-  }],
+  exampleCodeSnippet: `
+    <SpinnerButton bsStyle="primary" isLoading={isLoading} onClick={onClick}>
+      Save Details
+    </SpinnerButton>
```
omgaz: At the cost of code-readability we want to try to preserve the outputted indentation. Because of the way spaces/line-breaks are preserved with back-ticks can we make sure we format like so:

```
exampleCodeSnippet: `<SpinnerButton bsStyle="primary" isLoading={isLoading} onClick={onClick}>
  Save Details
</SpinnerButton>`
```

The code, here, gets a little messier but it's the compromise for a consistent view. Unless we have a better solution :/

docs/examples/UserListPickerExample.jsx
```diff
@@ -70,15 +70,69 @@ const exampleProps = {
...
+      defaultValue: 'false',
+    },
+    {
+      propType: 'userHeaders',
+      type: '{ string: label, string: toggle }',
+      defaultValue: <pre>{JSON.stringify({ label: 'Team', toggle: 'Member' })}</pre>,
+    },
+    {
+      propType: 'users',
+      type: 'arrayOf {string: avatar, string: givenName, number: id, string: surname}',
```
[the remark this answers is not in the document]
lightbringer1991: that's what it is now in the component itself, just `avatar`.
I'm happy to change it, but maybe not in this PR since it will actually change the component.

omgaz:
It looks like we need to / should update the example component defaultValue to always render inside pre-tags like we do for the propType field. Then we can get rid of all the \<pre\>'s and keep consistency in the way they render.

lightbringer1991:
@omgaz so will it need to be done in this PR, or you're happy for it to be done in another one.

omgaz:
It can be done separately.

PR #648, opened 2017-11-03
Build: Get ready for Node 8

omgaz:
If Travis is okay, and after a little housekeeping we can push to node 8 officially.


lteacher:
LGTM - Merging

PR #649, opened 2017-11-12
Fix: Safari SVG issues

sonhanguyen:
`svg4everyone` was removed when we merged `adslot` and `alexandria`. This caused problems displaying `svg`s in Safari because `SvgSymbol` uses `href` instead of `xlink:href`. There may be other issues or other browsers affected that could be fixed by this.
https://adslot.atlassian.net/browse/ADS-364


omgaz:
ADS-123

omgaz:
LGTMerging 🍥 🍜

PR #650, opened 2017-11-17
Update: upgrade nuka-carousel to remove the react warning

xiaofan2406:
- upgrade nuka-carousel  to remove the React proptypes warning from 15 -> 15.5

- as part of ADS-458

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "mocha": "^3.5.2",
     "moment": "^2.18.1",
     "node-sass": "^4.5.3",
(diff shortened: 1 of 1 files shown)


lightbringer1991:
it's still there, but because we were packaging prod build of `adslot-ui` with dev build in direct, so those errors bubbles up

it's not the case anymore, but still nice to upgrade this

xiaofan2406:
@lightbringer1991 in direct-web, its is `react-redux` and `adslot-uid` (caused by this lib) causing the error

PR #651, opened 2017-11-17
Fix: fixing some proptypes definitions

xiaofan2406:
I went through the media signoff process, and found these three places in adslot-ui are producing warnings.

As part of ADS-458

### DIFF
--- a/src/components/adslot-ui/AlertInput/index.jsx
+++ b/src/components/adslot-ui/AlertInput/index.jsx
@@ -72,16 +72,14 @@ export default class AlertInput extends Component {
       placeholder,
       prefixAddon,
       suffixAddon,
-      alertStatus = '',
+      alertStatus,
       alertMessage,
       onValueChange,
     } = this.props;
 
-    const className = classnames({
-      [baseClass]: true,
-      [alertStatus]: true,
-      'is-focused': this.state.isFocused,
-    });
+    const className = classnames(baseClass, { [alertStatus]: alertStatus, 'is-focused': this.state.isFocused });
+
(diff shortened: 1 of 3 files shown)


src/components/adslot-ui/AlertInput/index.jsx
```diff
@@ -72,16 +72,14 @@ export default class AlertInput extends Component {
...
     } = this.props;
 
-    const className = classnames({
-      [baseClass]: true,
-      [alertStatus]: true,
-      'is-focused': this.state.isFocused,
-    });
+    const className = classnames(baseClass, { [alertStatus]: alertStatus, 'is-focused': this.state.isFocused });
+
+    const popoverClassName = classnames(`${baseClass}-popover`, { alertStatus });
```
mdotwills: _we could probably get rid of this `baseClass` thing too to make the class strings searchable_

src/components/adslot-ui/AlertInput/index.jsx
```diff
@@ -72,16 +72,14 @@ export default class AlertInput extends Component {
       placeholder,
       prefixAddon,
       suffixAddon,
-      alertStatus = '',
+      alertStatus,
```
sonhanguyen: does this break anything? for example `alertStatus` being used as key down bellow?

xiaofan2406: nay. It will be undefined. and `classnames` will know how to deal with falsy values.

xiaofan2406: I manully tested on the adslot-ui doc page. It looks the same as before

xiaofan2406:
yeah most of them. and some from the requestproductmodal as well

PR #652, opened 2017-11-24
Fix: svg href issue

xiaofan2406:
Context:
- There was a PR changing `xlink:href` to `href` because the former is deprecated
- Then safari stopped working so there was another PR that added `svg4everybody` polyfill
- Then it turns out that updating `href` does not work in the polyfill

### DIFF
--- a/src/components/alexandria/SvgSymbol/index.jsx
+++ b/src/components/alexandria/SvgSymbol/index.jsx
@@ -11,7 +11,7 @@ const SvgSymbol = (props) => {
 
   return (
     <svg className={`${componentClass}${classesList}`} onClick={onClick}>
-      <use href={href} />
+      <use href={href} xlinkHref={href} />
     </svg>
   );
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -11552,11 +11552,6 @@
         "has-flag": "2.0.0"
       }
     },
-    "svg4everybody": {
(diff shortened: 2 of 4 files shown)


xiaofan2406:
why the package-lock is so big

src/components/alexandria/SvgSymbol/index.jsx
```diff
@@ -11,7 +11,7 @@ const SvgSymbol = (props) => {
 
   return (
     <svg className={`${componentClass}${classesList}`} onClick={onClick}>
-      <use href={href} />
+      <use href={href} xlinkHref={href} />
```
xiaofan2406: use both so that safari picks up the href too.

@sonhanguyen I think this may be the best thing to do for now.

sonhanguyen:
maybe don't check in package lock, we'll do it when release?

xiaofan2406:
@sonhanguyen I think the package-lock is corrupted in master. and this should fix it. 

You can test it by clean you `node_modules` folder in adslot-ui. in latest master, do and `npm i` it will modify the package-lock and give warning says some sha is invalid

sonhanguyen:
may have something to do with automatic versioning done by release. It's a separate issue anw.

bchew:
package-lock fixed in this PR: https://github.com/Adslot/adslot-ui/pull/655

PR #653, opened 2017-11-24
New: add Nav component

lightbringer1991:
Add Nav component, with `borderless bsStyle`

This is a wrapper around `react-bootstrap` Nav component, with the addition of `borderless` style
Works well with `NavItem`, `NavBar`

after

EDIT: Updated to match platform styles

navtav
navtavb

### DIFF
--- a/docs/examples/NavigationExample.jsx
+++ b/docs/examples/NavigationExample.jsx
@@ -1,0 +1,66 @@
+import React from 'react';
+import Example from '../components/Example';
+import {
+  Nav,
+  NavItem,
+} from '../../src/dist-entry';
+
+
+class NavigationExample extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      activeKey: 0,
+    };
+
+    this.onSelect = this.onSelect.bind(this);
+  }
+
+  onSelect(eventKey) {
+    this.setState({ activeKey: eventKey });
+  }
+
+  render() {
+    return (
+      <Nav activeKey={this.state.activeKey} onSelect={this.onSelect}>
+        <NavItem eventKey={0} href="#" title="Dashboard Home" className="dashboard-tab">Dashboard</NavItem>
+        <NavItem eventKey={1} title="Analytics Reporting">Reports</NavItem>
+      </Nav>
+    );
+  }
+}
+
+const exampleProps = {
+  componentName: 'Navigation Tabs',
+  notes: '',
+  exampleCodeSnippet: `<Nav activeKey={this.state.activeKey} onSelect={this.onSelect}>
+  <NavItem eventKey={0} href="#" title="Dashboard Home" className="dashboard-tab"}>Dashboard</NavItem>
+  <NavItem eventKey={1} title="Analytics Reporting">Reports</NavItem>
+</Nav>`,
+  propTypes: [
+    {
+      propType: 'React Bootstrap Nav prop types',
+      type: '',
+      defaultValue: '',
+      note: <a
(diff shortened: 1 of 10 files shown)


vinteo:
how is this different from tabs?

lightbringer1991:
@vinteo it doesn't have any tab content body, just the navigation bar with a bunch of buttons

docs/examples/NavigationExample.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+class NavigationExample extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      activeKey: 1,
+    };
+
+    this.onSelect = this.onSelect.bind(this);
```
omgaz: https://adslot.atlassian.net/wiki/spaces/APT/pages/178225251/React

Can we use the class arrow functions. They're supported in Adslot UI through babel preset env.

docs/examples/NavigationExample.jsx
```diff
@@ -0,0 +1,75 @@
...
+  onSelect(eventKey) {
+    this.setState({ activeKey: eventKey });
+  }
+
+  render() {
+    return (
+      <Nav bsStyle="borderless" activeKey={this.state.activeKey} onSelect={this.onSelect}>
+        <NavItem eventKey={1} href="#">NavItem 1 content</NavItem>
+        <NavItem eventKey={2} title="Item">NavItem 2 content</NavItem>
+        <NavItem eventKey={3} disabled>NavItem 3 content</NavItem>
```
omgaz: Dashboard
Reports

And we can remove the disabled one from the example.

src/components/adslot-ui/Navigation/index.jsx
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+
+NavigationComponent.propTypes = _.assign({}, Nav.propTypes, {
+  bsStyle: PropTypes.oneOf(['tabs', 'pills', 'borderless']),
+  barPosition: PropTypes.oneOf(['top', 'bottom']),
+});
+
+NavigationComponent.defaultProps = {
+  bsStyle: 'pills',
```
omgaz: We should only be supporting borderless, so there shouldn't be a prop-type for this.

src/components/adslot-ui/Navigation/styles.scss
```diff
@@ -0,0 +1,55 @@
...
+.nav-borderless {
+  display: flex;
+  flex-direction: row;
+  font-size: $font-size-header;
+  color: $navbar-text-color;
+  font-weight: $font-weight-bold;
+
+  li {
+    list-style-type: none;
+    flex: 0 1 auto;
```
omgaz: Can you check this in FF and IE. Sometimes 0 can do weird things in IE and sometimes auto can do weird things in FF.

src/components/adslot-ui/Navigation/styles.scss
```diff
@@ -0,0 +1,55 @@
...
+      a {
+        color: $color-text-disabled;
+      }
+    }
+  }
+
+  &.top-bar {
+    li {
+      .active {
+        border-top: 4px solid $color-primary-selected;
```
omgaz: The default for buttons is grey, can we use the same default highlight here with the intention that it is overridden by the implementation of the feature?

lightbringer1991:
LGTM, can you remove need to remove the `package.lock` file?

omgaz:
@lightbringer1991 could do, but they're just point releases from an `npm i --no-optional` so this is valid

lightbringer1991:
@omgaz please merge when possible, thank you

PR #654, opened 2017-11-24
Docs: Add docs to dist so ready for gh-pages

omgaz:
Now when we run dist we generate extra files for github pages.

Release will need to be updated to reset gh-pages to master when we're done.

This is a quick solution to getting gh-pages done.


index.html
```diff
@@ -0,0 +1,35 @@
...
+  <link href="./dist/adslot-ui-docs.css" rel="stylesheet">
+</head>
+<body>
+  <noscript>
+    You need to enable JavaScript to run this app.
+  </noscript>
+  <div id="app"></div>
+
+  <script
+    src="https://cdnjs.cloudflare.com/ajax/libs/react/15.6.2/react.min.js"
```
lightbringer1991: I think if the docs is built via webpack, these libraries should be already included in the bundled files

vinteo:
Add the gh-pages link to the README?

index.html
omgaz: Not from the looks of it. Maybe the webpack dev server is injecting them in or removing them because they're in the externals?

If I take them out it breaks, so happy to leave them in for now.

lightbringer1991: yes if the library is in externals then it's not included in the bundled files

bchew:
Merging :rocket:

PR #656, opened 2017-11-28
Fix: use of React.PropTypes

sonhanguyen:
##### Changes
- Remove use of React.PropTypes
- Remove some react warnings in tests that are not critical but annoying

### DIFF
--- a/src/components/adslot-ui/HoverDropdownMenu/PopoverLinkItem/index.jsx
+++ b/src/components/adslot-ui/HoverDropdownMenu/PopoverLinkItem/index.jsx
@@ -1,12 +1,13 @@
-/* eslint-disable react/prop-types */
 import _ from 'lodash';
-import React, { PropTypes } from 'react';
+import React from 'react';
+import PropTypes from 'prop-types';
 import { Button } from 'react-bootstrap';
 import './styles.scss';
 
 
 class PopoverLinkItemComponent extends React.PureComponent {
   render() {
+    // eslint-disable-next-line react/prop-types
     const { target, title, url, isEnabled, onClick } = this.props;
 
     const buttonProps = {
--- a/src/components/adslot-ui/HoverDropdownMenu/index.spec.jsx
+++ b/src/components/adslot-ui/HoverDropdownMenu/index.spec.jsx
@@ -52,7 +52,7 @@ describe('HoverDropdownMenuComponent', () => {
   it('should render popover with list of links when `links` is not empty', () => {
     const component = shallow(
       <HoverDropdownMenu {...props}>
-        {_.map(links, (link) => <HoverDropdownMenu.Item {...link} />)}
+        {_.map(links, (link, idx) => <HoverDropdownMenu.Item key={idx} {...link} />)}
       </HoverDropdownMenu>
     );
     expect(component.find(Overlay)).to.have.length(1);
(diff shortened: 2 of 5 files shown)


src/components/adslot-ui/HoverDropdownMenu/index.spec.jsx
```diff
@@ -62,21 +62,21 @@ describe('HoverDropdownMenuComponent', () => {
...
-        {_.map(links, (link) => <HoverDropdownMenu.Item {...link} />)}
+        {_.map(links, (link, idx) => <HoverDropdownMenu.Item key={idx} {...link} />)}
       </HoverDropdownMenu>
     );
-    component.find('.hover-dropdown').simulate('mouseEnter', { target: 'some target' });
+    component.find('.hover-dropdown').simulate('mouseEnter', { target: _.noop });
     expect(component.state()).to.eql({
       isOpen: true,
-      target: 'some target',
+      target: _.noop,
```
sonhanguyen: the proptype does not accept string

omgaz:
LGTM but should target be defined as something else like `<div></div>`

src/components/adslot-ui/HoverDropdownMenu/index.spec.jsx
sharmajyo: you mean, we can't assign any string value to prop? or its just some special case where we can't

sonhanguyen: srr I mean `Overlay.target` does not accept string, it does accept `function(): DomElement` though

src/components/adslot-ui/HoverDropdownMenu/index.spec.jsx
```diff
@@ -139,6 +140,7 @@ describe('HoverDropdownMenuComponent', () => {
 
     beforeEach(() => {
       popoverProps = {
+        id: 'popover',
```
sonhanguyen: required for accessibility

src/components/adslot-ui/ListPicker/index.jsx
```diff
@@ -161,14 +161,14 @@ class ListPickerComponent extends React.Component {
           {
             _.isEmpty(linkButtons) ? null :
             <div className="pull-left">
-              {_.map(linkButtons, (linkButton) => (
+              {_.map(linkButtons, (linkButton, key) => (
```
xiaofan2406: just curious why this change?

src/components/adslot-ui/HoverDropdownMenu/PopoverLinkItem/index.jsx
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable react/prop-types */
```
xiaofan2406: is this line relevant to remove?

src/components/adslot-ui/ListPicker/index.jsx
sonhanguyen: https://github.com/Adslot/adslot-ui/pull/656/files#diff-8e90d83db60d7e3897dc4457f7ea133eR171

src/components/adslot-ui/HoverDropdownMenu/PopoverLinkItem/index.jsx
sonhanguyen: that looks like it's for some `renderX` function to pass linter, not really causes any runtime warning atm

sonhanguyen: srr it's actually for [line #42](https://github.com/Adslot/adslot-ui/pull/656/files/20055cfd494c96851feeff79b360dd6354e0da33#diff-0b124245e4f0bfd4f3073eacb42bd13dR42) to deal with eslint not being able to analyse

vinteo: is it possible to move it to disable for just that line?

lightbringer1991:
merging ...

PR #657, opened 2017-11-28
Add missing props for select example

ChaoLiangSuper:
Adding two props which are in the Select component code but not in the Select example


docs/examples/SelectExample.jsx
```diff
@@ -49,6 +49,8 @@ const exampleProps = {
     { value: 'uk', label: 'United Kingdom' },
   ]}
   placeholder="Countries"
+  value={this.state.selected}
+  onChange={this.onChangeHandler}
 />`,
   propTypes: [],
 };
```
ChaoLiangSuper: These two props exist in the code but not in the example

sonhanguyen: while you are at this, could you please add one more level indentation from line 42 to 54?

PR #658, opened 2017-12-01
Fix: move react-hot-loader to dev only, remove babel-cli

xiaofan2406:
- remove babel-cli, since we are not using it atm
- put browserlist in package.json so that both autoprefixer and preset-env knows about it
- include `react-hot-loader/babel` only in dev. Currently `react-hot-loader/babel` is included in the distributed files


omgaz:
Merging 🍜

PR #660, opened 2017-12-05
[Docs] Add max width AND add state for navigation bar

ChaoLiangSuper:
- [x] Set max width on page so prop types grid and code example don't go beyond 940px
  * Set code block a max width of 940px
  * Set prop-types table a max width of 1100px
  * Set width for each column in prop-types table

- [x] Set sidebar accordion state to only have one open at a time (close all others when you select a new heading)
  * The Navigation component's state only have the name the current open panel instead of the whole panels list

screen shot 2017-12-05 at 5 30 31 pm

### DIFF
--- a/docs/components/Example/index.jsx
+++ b/docs/components/Example/index.jsx
@@ -44,35 +44,38 @@ class Example extends React.PureComponent {
           {children}
         </div>
 
-        <SyntaxHighlighter language="html" style={github}>{exampleCodeSnippet}</SyntaxHighlighter>
-
+        <div className="adslot-ui-code-snippet">
+          <SyntaxHighlighter language="html" style={github}>{exampleCodeSnippet}</SyntaxHighlighter>
+        </div>
 
         <h3>PropTypes</h3>
-        <table className="table table-striped table-bordered">
-          <thead>
-            <tr>
-              <th>PropType</th>
-              <th>Type</th>
-              <th>Default Value</th>
-              <th>Notes</th>
-            </tr>
-          </thead>
-          <tbody>
-            {_.map(propTypes, ({ propType, type, defaultValue, note }) => (
-              <tr key={propType}>
-                <td><pre>{propType}</pre></td>
-                <td><pre>{type}</pre></td>
-                <td>{defaultValue}</td>
-                <td>{note}</td>
+        <div className="adslot-ui-proptype-table">
+          <table className="table table-striped table-bordered">
+            <thead>
+              <tr>
+                <th>PropType</th>
+                <th>Type</th>
+                <th>Default Value</th>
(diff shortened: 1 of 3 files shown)


omgaz:
Can you add a screenshot for design review.

docs/components/Example/styles.scss
```diff
@@ -8,6 +8,14 @@
     max-width: 940px;
   }
 
+  .adslot-ui-code-snippet {
+    max-width: 940px;
+  }
+
+  .adslot-proptype-table {
```
omgaz: .adslot-ui-proptype-table

docs/components/Navigation/index.jsx
```diff
@@ -7,37 +7,41 @@ import {
...
   togglePanel(panelId) {
-    const nextPanels = _.assign({}, this.state.panels);
-    const panel = _.find(nextPanels, { id: panelId });
-    panel.isCollapsed = !panel.isCollapsed;
-    this.setState({ panels: nextPanels });
+    const nextPanel = panelId === this.state.currentOpenPanel ? '' : panelId
+    this.setState({ currentOpenPanel: nextPanel })
+  }
+
+  generatePanels(){
```
lightbringer1991: this function seems to be used only once, considering putting it into `render()` instead
so
```jsx
const panels = _.map(this.props.componentsBySection, panelFactory(this.props.navigateTo, this.state.currentOpenPanel));
return (
  <div className="adslot-ui-navigation">
    <Accordion onPanelClick={this.togglePanel} panels={panels} />
  </div>
);
```

or even
```jsx
const accordionProps = {
  onPanelClick: this.togglePanel,
  panels: _.map(this.props.componentsBySection, panelFactory(this.props.navigateTo, this.state.currentOpenPanel)),
};

return (
  <div className="adslot-ui-navigation">
    <Accordion {...accordionProps} />
  </div>
);
```

omgaz:
LGTM, `fix-up` your commits into one and I'd be happy to merge.

docs/components/Example/index.jsx
```diff
@@ -44,35 +44,38 @@ class Example extends React.PureComponent {
...
-            </tr>
-          </thead>
-          <tbody>
-            {_.map(propTypes, ({ propType, type, defaultValue, note }) => (
-              <tr key={propType}>
-                <td><pre>{propType}</pre></td>
-                <td><pre>{type}</pre></td>
-                <td>{defaultValue}</td>
-                <td>{note}</td>
+        <div className="adslot-proptype-table">
```
omgaz: You've updated the css class to adslot**-ui**-proptype-table, as per CR comments, but not here.

omgaz:
Merging

PR #665, opened 2017-12-07
Fix: ESLint on ./docs

### LINKED ISSUE #664: Fix ESLint on Docs
ESLint is not running on the docs. Whilst we don't care about coverage on these files it'd be nice to have syntax consistency until we can get prettier in.

- [x] Exclude examples because of formatting issues.

omgaz:
ESLint was not running on docs.
ESLint should run on docs.
ESLint should still ignore examples (because of formatting requirements).

Resolves #664

### DIFF
--- a/docs/components/Contributors/index.jsx
+++ b/docs/components/Contributors/index.jsx
@@ -14,18 +14,16 @@ class Contributors extends React.Component {
     this.state = { contributors: [] };
     this.renderContributors = this.renderContributors.bind(this);
 
-    this.getContributors().then((contributors) => {
-      this.setState({ contributors });
-    });
+    this.getContributors();
   }
-
 
   getContributors() {
-    return new Promise((resolve) => {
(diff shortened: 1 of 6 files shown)


docs/components/Contributors/index.jsx
```diff
@@ -8,25 +8,23 @@ import {
 
 import './styles.scss';
 
+function getContributors() {
```
omgaz: getContributors should be a static method.

ChaoLiangSuper:
Maybe also need to apply scss lint to doc folder as well?
Current `"sass-lint": "sass-lint './src/**/*.scss' -v",`

docs/components/Contributors/index.jsx
lightbringer1991: is it because eslint complains that this function can be static?

omgaz:
@ChaoLiangSuper https://github.com/Adslot/adslot-ui/issues/666

Issue #631, opened 2017-10-08
Proposal: Add babel-plugin-react-bootstrap to dist build


lightbringer1991:
https://www.npmjs.com/package/babel-plugin-react-bootstrap

This is useful to cherrypick only the components that we use from `react-bootstrap` library
Below is the dist size before and after applying the plugin. We can drop ~100kb just by applying it.

Before
```bash
              Asset    Size   Chunks                    Chunk Names
  adslot-ui-main.js  754 kB  0, 1, 2  [emitted]  [big]  main
  adslot-ui-core.js  454 kB        1  [emitted]  [big]  core
 adslot-ui-extra.js  302 kB        2  [emitted]  [big]  extra
 adslot-ui-main.css  270 kB  0, 1, 2  [emitted]  [big]  main
 adslot-ui-core.css  145 kB        1  [emitted]         core
adslot-ui-extra.css  125 kB        2  [emitted]         extra
```

After
```bash
              Asset    Size   Chunks                    Chunk Names
  adslot-ui-main.js  657 kB  0, 1, 2  [emitted]  [big]  main
  adslot-ui-core.js  356 kB        1  [emitted]  [big]  core
 adslot-ui-extra.js  302 kB        2  [emitted]  [big]  extra
 adslot-ui-main.css  270 kB  0, 1, 2  [emitted]  [big]  main
 adslot-ui-core.css  145 kB        1  [emitted]         core
adslot-ui-extra.css  125 kB        2  [emitted]         extra
```

omgaz:
Would this be something you'd be interested at looking into, @lightbringer1991?

Just remember that we use Bootstrap CSS without using the Bootstrap React components. So `.col-xs-12` etc... would still need to be usable.

PR #668, opened 2017-12-08
Build: add babel-plugin-react-bootstrap

lightbringer1991:
Resolves https://github.com/Adslot/adslot-ui/issues/631

before
screen shot 2017-12-08 at 11 04 43 am

after
screen shot 2017-12-08 at 11 26 59 am

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-  "presets": ["./.babelrc.js"]
+  "presets": ["./.babelrc.js"],
+  "plugins": ["react-bootstrap"]
 }
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "webpack-merge": "^4.1.0"
   },
   "dependencies": {
+    "babel-plugin-react-bootstrap": "0.0.2",
     "diff-match-patch": "^1.0.0",
     "lodash": "^4.17.4",
     "react": "^15.6.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -688,6 +688,11 @@
         "babel-runtime": "6.26.0"
       }
     },
+    "babel-plugin-react-bootstrap": {
+      "version": "0.0.2",
+      "resolved": "https://registry.npmjs.org/babel-plugin-react-bootstrap/-/babel-plugin-react-bootstrap-0.0.2.tgz",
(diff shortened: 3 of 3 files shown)


.babelrc
```diff
@@ -1,3 +1,4 @@
 {
-  "presets": ["./.babelrc.js"]
+  "presets": ["./.babelrc.js"],
+  "plugins": ["react-bootstrap"]
```
xiaofan2406: should put the plugins in the .babelrc.js file. This file is not intended to be removed once we move to babel 7

lightbringer1991: I think we can just move now, as the `.babelrc.js` is being included as presets, and not the whole babel config. Therefore all plugins in js file are within `presets` attributes

xiaofan2406: presets includes plugins

lightbringer1991: hmm I just tried, but it doesn't seem to pick up the plugins inside `.babelrc.js`

feels like it's incorrect syntax, as the end result of .babelrc is something like
```json
{
  "presets": [
    "presets": ["env", "react"],
    "plugins": ["transform-strict-mode", "react-bootstrap"]
  ]
}
```

while it should be
```json
{
  "presets": ["env", "react"],
  "plugins": ["transform-strict-mode", "react-bootstrap"]
}
```

please correct me if I understand it wrong

package.json
```diff
@@ -104,6 +104,7 @@
     "webpack-merge": "^4.1.0"
   },
   "dependencies": {
+    "babel-plugin-react-bootstrap": "0.0.2",
```
xiaofan2406: also this should be in devdeps

.babelrc
xiaofan2406: hum. let me do some testing

xiaofan2406: @lightbringer1991 it is definitely picking up the plugins in the .babelrc.js file. I tested locally with class property plugin.

It can be the plugin itself not function correctly. I see another one named `babel-plugin-react-bootstrap-fix`.

omgaz:
Swish! Looks good if the qualms have been resolved.

omgaz:
@lightbringer1991 @xiaofan2406 what's happening with this PR? Can we merge?

lightbringer1991:
@omgaz I'll need more time on this, so could you please exclude this PR from the release

Issue #666, opened 2017-12-07
Fix SASSLint on Docs


omgaz:
Similar to https://github.com/Adslot/adslot-ui/issues/664

PR #671, opened 2017-12-11
Fix: Apply sass lint to doc folder

ChaoLiangSuper:
- [x] Apply sass lint check to Document folder
  <s>* Add ignore folder `docs/examples`</s>
- [x] Change some scss code based on sass lint

Issue page https://github.com/Adslot/adslot-ui/issues/666

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -1,7 +1,9 @@
 options:
   formatter: stylish
 files:
-  include: '**/*.scss'
+  include:
+    - 'src/**/*.scss'
+    - 'docs/**/*.scss'
 rules:
   # Extends
   extends-before-mixins: 2
--- a/docs/components/Header/styles.scss
+++ b/docs/components/Header/styles.scss
@@ -1,6 +1,8 @@
+@import '~styles/variable';
+
 .adslot-ui-header {
   > .pagetitle-component {
-    background: #fff;
+    background: $color-background;
 
     .svg-symbol-component {
       width: 156px;
--- a/docs/components/Example/index.jsx
+++ b/docs/components/Example/index.jsx
@@ -12,7 +12,10 @@ class Example extends React.PureComponent {
     const { children, componentName, notes, exampleCodeSnippet, propTypes } = this.props;
 
     return (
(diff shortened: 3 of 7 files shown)


.sass-lint.yml
```diff
@@ -1,7 +1,10 @@
 options:
   formatter: stylish
 files:
-  include: '**/*.scss'
+  include:
+    - 'src/**/*.scss'
+    - 'docs/**/*.scss'
+  ignore: 'docs/examples/**/*.scss'
```
vinteo: why do we want to ignore the examples?

ChaoLiangSuper: All examples are using #ID to defined width. 
But this will throw problems in sass lint. `Bad: highly-specific styling for a single element via ID`

lightbringer1991: it seems a little bit weird here, since we want to apply linting to docs, but ignoring some files
have you flow this past @omgaz ?

vinteo: I would find a way to disable the rule for that folder instead of ignoring the folder altogether

vinteo: or change the styling

ChaoLiangSuper: So, I will change the style.

ChaoLiangSuper: Updated.

omgaz:
LGTM

xiaofan2406:
not sure why `package-lock` got changed. there is no dependency changes

cc @bchew

ChaoLiangSuper:
@xiaofan2406  Yep. This change happens after `npm install`.  And I saw the `package-lock` has only changed the style of line.

docs/components/Header/styles.scss
```diff
@@ -1,6 +1,8 @@
+$background: #fff;
+
 .adslot-ui-header {
   > .pagetitle-component {
-    background: #fff;
+    background: $background;
```
omgaz: This is already defined in `color.scss` as `$color-background` which we should import and use.

ChaoLiangSuper: Fixed

omgaz:
LGTMerging 🍜 🍥

PR #672, opened 2017-12-11
Fix: dist error

xiaofan2406:
change the `env` preset so that it does transpile to commonjs modules. Because `src/dist-entry/index.js` uses `module.exorts` instead of es6 modules.

Ideally, we can utilise babel export extension plugin to fix `src/dist-entry/index.js`, but for now this fix should work

And subsequently need to modify hot-reloader setup.


omgaz:
Looks like it fixes it locally after I use this branch and re-install my node dependencies. Will re-run and see if it resolves.

Issue #673, opened 2017-12-14
Fix: Search bar svg is not showing in production


ChaoLiangSuper:
## Types of issue

- [ ] Bug (High impact or critical to a production build)
- [x] Bug (Low impact or not working as expected)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Something else (idea, build tool, documentation improvement)

## Description

The svg in search bar component is missing

## Possible Solution

Change the path of the svg in search component

## Screenshots

![screen shot 2017-12-15 at 10 17 38 am](https://user-images.githubusercontent.com/15777593/34019077-35939208-e181-11e7-9ca5-ecf4d1857485.png)

PR #674, opened 2017-12-14
Fix: Fix search component svg problem

ChaoLiangSuper:
## Description

Fix the missing svg in search component.
Issue: https://github.com/Adslot/adslot-ui/issues/673

- [ ] Yes
- [x] No

### Before
screen shot 2017-12-15 at 10 17 38 am

### After
screen shot 2017-12-15 at 10 35 10 am

## Check-list:

### DIFF
--- a/src/components/adslot-ui/Search/index.jsx
+++ b/src/components/adslot-ui/Search/index.jsx
@@ -101,11 +101,11 @@ Search.defaultProps = {
   placeholder: '',
   svgSymbolCancel: {
     classSuffixes: ['gray-darker'],
-    href: '/assets/svg-symbols.svg#cancel',
+    href: './docs/assets/svg-symbols.svg#cancel',
   },
   svgSymbolSearch: {
     classSuffixes: ['gray-light'],
-    href: '/assets/svg-symbols.svg#search',
+    href: './docs/assets/svg-symbols.svg#search',
   },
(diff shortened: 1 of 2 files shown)


src/components/adslot-ui/Search/index.jsx
```diff
@@ -101,11 +101,11 @@ Search.defaultProps = {
   placeholder: '',
   svgSymbolCancel: {
     classSuffixes: ['gray-darker'],
-    href: '/assets/svg-symbols.svg#cancel',
+    href: './docs/assets/svg-symbols.svg#cancel',
```
lightbringer1991: hmm this is the actual component, I don't think it should refer to the documentation files.

ChaoLiangSuper: Currently, every SVG is using the same path somehow under the `./docs/assets/...`. Except for this one. That's the reason for missing this SVG.
But, I agree with not putting SVG inside the `./docs/...`

omgaz: Our webpack build will convert this to a data-url (IIRC), so although it's not ideal it shouldn't be exported as such.

The reason the assets were moved to docs was based on priorities of having a preference of `looks good in github pages' so the dev teams get the right view.

PR #686, opened 2018-01-11
[ADS-645] Remove extra chunk and make moment external

sonhanguyen:
## Description
- Remove the extra bundle
- Externalizing `moment`js

- Currently, the build generates 3 bundles: "core", "extra" (that contains `react-datepicker`) and "main" that has everything. The splitting was done mainly because `react-datepicker` has `moment`js bundled which we don't use but Symphony does, and `moment`'s size is considerable.
- `direct-web` however has not switched to only import `core`, so `moment` code is duplicated anyway.
for SYMPHONY in case they don't already have `moment` as a dependency, If they do this saves them 240K as well

- [x] Yes
- [ ] No

## Check-list:

### DIFF
--- a/src/dist-entry/extra.js
+++ b/src/dist-entry/extra.js
@@ -1,4 +1,0 @@
-import 'styles/_react-datepicker-custom.scss';
-import DatePicker from 'react-datepicker/dist/react-datepicker';
-
-export { DatePicker }; // eslint-disable-line import/prefer-default-export
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -4,7 +4,7 @@ if (process.env.TYPE === 'development') {
   plugins = [...plugins, 'react-hot-loader/babel'];
 }
 
-const presets = ['env', 'react'];
+const presets = process.env.TYPE === 'development' ? [['env', { modules: false }], 'react'] : ['env', 'react'];
 
 module.exports = {
   presets,
--- a/config/dist.js
+++ b/config/dist.js
@@ -8,9 +8,7 @@ module.exports = () => {
   const buildType = process.env.TYPE || 'development';
 
   let entries = {
-    main: resolve(__dirname, '../src/dist-entry'),
-    core: resolve(__dirname, '../src/dist-entry/core'),
-    extra: resolve(__dirname, '../src/dist-entry/extra'),
+    main: resolve(__dirname, '../src'),
     docs: resolve(__dirname, '../docs/run'),
   };
   const plugins = [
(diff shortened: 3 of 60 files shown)


omgaz:
The issue wasn't with `moment` it was with `moment-tz`, and where further breaking changes may appear. Have a look at https://github.com/Hacker0x01/react-datepicker#localization and check localisations.

omgaz:
Apart from the comment around localisation, LGTM.

sonhanguyen:
@omgaz I don't see many problems because `moment` would be the in same line to `lodash` by that logic. `react-datepicker` externalizes it too as a `peerDependencies`. We initially did `import 'react-datepicker/dist/react-datepicke'` which wasn't the intended way, that's why it was included.

omgaz:
@sonhanguyen did you send this around for another review?

src/dist-entry/extra.js
```diff
@@ -1,4 +0,0 @@
-import 'styles/_react-datepicker-custom.scss';
```
xiaofan2406: 👍

sonhanguyen:
@omgaz @xiaofan2406, should we make moment an optional dependency?

xiaofan2406:
It should be a peerDeps, like `react`

src/dist-entry/extra.js
xiaofan2406: by removing these extra entries, you can revert changes made in https://github.com/Adslot/adslot-ui/pull/672 (see its description)

sonhanguyen:
@xiaofan2406 yeah but peer means required. We happen to have `moment` in `direct-web` too, but we didn't include `moment` because of `adslot-ui`. We don't actually use its datepicker component.

sonhanguyen:
@omgaz what do you think about react as a peerDep, seems like the way most projects do it.

src/dist-entry/extra.js
sonhanguyen: just `babelrc` you mean? Keeping react hot stuff?

omgaz:
moment is a requirement so it does make sense for it to be a peer dependency like react, but we can deal with it in another PR.

Also, we do use the react-datepicker in all projects.

xiaofan2406:
I remember peer deps will come off as warning not error. 

So if a project uses adslot-ui and dont use the date picker component, the project can choose to not to install moment, and other components should be working correctly

src/dist-entry/extra.js
sonhanguyen: reverted

omgaz:
Rebased for release.

PR #693, opened 2018-02-15
Update: Allow treepicker to accept both path or ancestor as props

ChaoLiangSuper:
Allow treePickerNode component using either `path` or `ancestors` to identify its path.

## Description

1. Add new prop type `ancestors` to treePickerNode component
2. Set both `path` and `ancestors` to be not required
3. Add error message to treePickerNode component, if neither `path` nor `ancestors` are provided.

No need to update the docs at this moment, in the following PR will add examples for treePickerNode component, and add document in that PR

Adding `ancestors` props in order to solve real case problem such as legacy naming as `ancestors`. 

- [ ] Yes
- [x] No

## How Has This Been Tested?

1. Passing a node to treePickerNode component, and this node does not contain `path` or `ancestors`
2. The component will throw an error message.

## Check-list:

### DIFF
--- a/config/dev.js
+++ b/config/dev.js
@@ -29,18 +29,6 @@ module.exports = merge(commonConfig, {
   module: {
     rules: [
       {
-        test: /\.(js|jsx)$/,
-        loader: 'babel-loader',
-        include: resolve(__dirname, '../docs'),
-        exclude: /node_modules/,
-        options: {
-          // This is a feature of `babel-loader` for webpack (not Babel itself). It enables caching results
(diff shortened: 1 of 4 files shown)


config/dev.js
```diff
@@ -28,18 +28,6 @@ module.exports = merge(commonConfig, {
   },
   module: {
     rules: [
-      {
```
ChaoLiangSuper: Remove duplication settings here. Not related to this PR.

ChaoLiangSuper:
Tested in the real case, after this change, the error message does not exist.

omgaz:
Merging 🍜

PR #694, opened 2018-02-19
Docs: Split propTypes to an isolated component

ChaoLiangSuper:
Create a new component for proptype table.

## Description

Move proptype table to a separate component for reuse purpose.

The PR is will update the document, then proptype table component can be reusable. 
After this, in the following PRs, will be able to add multiple proptype tables to one component example.

- [ ] Yes
- [x] No

Not need.

## Check-list:

### DIFF
--- a/docs/components/PropTypeTable/index.jsx
+++ b/docs/components/PropTypeTable/index.jsx
@@ -1,0 +1,62 @@
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Empty } from '../../../src';
+
+import './styles.scss';
+
+class PropTypeTable extends React.PureComponent {
+  render() {
+    const { propTypes } = this.props;
+
+    return (
+      <div>
+        <h3>PropTypes</h3>
+        <div className="adslot-ui-proptype-table">
+          <table className="table table-striped table-bordered">
+            <thead>
+              <tr>
+                <th className="col-prop-type">PropType</th>
+                <th className="col-type">Type</th>
+                <th className="col-default-value">Default Value</th>
+                <th className="col-notes">Notes</th>
+              </tr>
+            </thead>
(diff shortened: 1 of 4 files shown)


docs/components/PropTypeTable/index.jsx
```diff
@@ -0,0 +1,62 @@
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Empty } from '../../../src';
+
+import './styles.scss';
+
+class PropTypeTable extends React.PureComponent {
```
ChaoLiangSuper: Any better naming suggestion?

docs/components/PropTypeTable/styles.scss
```diff
@@ -0,0 +1,45 @@
...
+    tr {
+      td {
+
+        &:empty {
+          &::before {
+            content: '—';
+          }
+        }
+
+        &:nth-child(1) {
```
nimishjha: Instead of nth-child can we use class names? It's better semantically and also more robust.

ChaoLiangSuper: This is just to apply the width to each column from left to the right. Do we need a name for each column? @nimishjha

lightbringer1991: ideally a class for each column is not a bad thing imo, since if we swap the columns around, the whole width column thingy will be messed up. At least with the class names, we don't have to worry about it.

Also there are only 4 columns, so I guess it's easy to name them. It's not like we're generating dynamic amount of columns anyway.

ChaoLiangSuper: OK, so will turn it into a class name. @lightbringer1991 @nimishjha

omgaz:
merging 🍜

PR #695, opened 2018-02-20
Update: Allow custom class to grid cell component

ChaoLiangSuper:
## Description

- Add `addonClasses` prop to grid cell component
- Added one example using custom class to doc
- Added test of using custom class

In the real case, when using `GridCell` component for a large paragraph or long text. The text will go out of the border. So adding a custom class like `width: 100%` will solve this problem. 

Moreover, providing custom class will allow customization for this component.

- [ ] Yes
- [x] No

## How Has This Been Tested?

1. Loading `GridCell` component and passing an array of class names
2. After render, those class names will be added on to the component.

screen shot 2018-02-20 at 11 51 47 am

## Check-list:

### DIFF
--- a/docs/examples/GridExample.jsx
+++ b/docs/examples/GridExample.jsx
@@ -23,6 +23,14 @@ class GridExample extends React.PureComponent {
           <GridCell>Content</GridCell>
           <GridCell onClick={cellClicked}>This Cell logs clicks.</GridCell>
         </GridRow>
+        <GridRow>
+          <GridCell addonClassNames={['full-width']}>
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus finibus suscipit velit quis tempor. Nunc
+            non lorem viverra, tincidunt dui ut, dictum enim. Nam quis ligula ac mi egestas scelerisque. Phasellus
+            posuere tellus quis nisl vulputate efficitur. Nulla laoreet ut ex vitae pharetra. Vivamus felis eros,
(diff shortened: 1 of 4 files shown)


docs/examples/GridExample.jsx
```diff
@@ -23,30 +23,47 @@ class GridExample extends React.PureComponent {
           <GridCell>Content</GridCell>
           <GridCell onClick={cellClicked}>This Cell logs clicks.</GridCell>
         </GridRow>
+        <GridRow>
+          <GridCell addonClasses={['full-width']}>
```
ChaoLiangSuper: Better naming for this prop?

nimishjha: ```customClasses``` maybe? Doesn't really matter.

omgaz: For now I'd aim for consistency. I think additionalclassnames is used msot places but 'addonClassnames' may be better

docs/examples/styles.scss
```diff
@@ -63,3 +63,7 @@
     }
   }
 }
+
+.full-width {
```
omgaz: Maybe we should start isolating these docs only styles like .docs__full-width

ChaoLiangSuper: Not sure, this class is only used in this example not even for docs. I agree with adding prefix as `doc__`, but not sure it is needed here? @omgaz

omgaz: Sure thing.

omgaz:
Merging 🍜

PR #711, opened 2018-04-17
[ADS-868] [SubTask of Widening] Update: Update tile grid component

ChaoLiangSuper:
Update tile grid component to allow image background, width, and distributed display

TODO
- [x] Docs
- [x] Tests

## Description

After update:
The `Tile Grid` will accept props of shape 
``` 
TileGrid.propTypes = {
  title: PropTypes.string.isRequired,
  items: PropTypes.arrayOf(
    PropTypes.shape({
      id: idPropType.isRequired,
      classSuffix: PropTypes.string.isRequired,
      title: PropTypes.string.isRequired,
      imgLink: PropTypes.string,
      width: PropTypes.number,
      maxWidth: PropTypes.number,
      imgAlign: PropTypes.oneOf(['left', 'right', 'center']),
    })
  ).isRequired,
  onItemClick: PropTypes.func.isRequired,
  distributed: PropTypes.bool,
};
```

- [ ] Yes
- [x] Maybe (Introduced new props and changed styles, but didn't change previous props)
- [ ] No

apr-17-2018 14-50-04

Current docs:
screen shot 2018-04-19 at 11 37 55 am

## Check-list:


lightbringer1991:
just wondering if it is easy to switch to something like this
```jsx
< Tile.Grid {...props}>
  <Tile.Item {...itemProps1} />
  <Tile.Item {...itemProps2} />
  ...
</Tile.Grid>
```

So that you don't have to pass in the whole list of items, and therefore removing the necessity of the prop shape checking.

ChaoLiangSuper:
@lightbringer1991 
I think, because we dont know how many items the user will pass in, so in your solution, we still need a list of `<Tile.Item />` is that right?

And also, currently, we just pass a list of items to the TileGrid component. If define two components `Tile.Grid` and `Tile.Item`. It is going to be a breaking change.

docs/examples/TileGridExample.jsx
```diff
@@ -1,61 +1,160 @@
...
+          title="Browse by category (Using background images and positions)"
+          items={[
+            { id: '0', classSuffix: 'left', title: 'Left', imgLink: exampleImageLink },
+            { id: '1', classSuffix: 'center', title: 'Center', imgLink: exampleImageLink, imgAlign: 'center' },
+            { id: '2', classSuffix: 'right', title: 'Right', imgLink: exampleImageLink, imgAlign: 'right' },
+          ]}
+          onItemClick={this.onClick}
+          distributed
+        />
+      </Fragment>
```
omgaz: Can we choose one meaningful example that demonstrates the usage in Marketplace.

src/components/alexandria/TileGrid/index.jsx
```diff
@@ -5,35 +5,75 @@ import PropTypes from 'prop-types';
 import idPropType from '../../prop-types/idPropType';
 import './styles.scss';
 
-const TileGrid = ({ title, items, onItemClick }) => {
-  const baseClass = 'tile-grid-component';
+const defaultWidth = 204; // 204px
+const defaultMaxWidth = 295; // 295px
```
omgaz: Are there reasons behind these numbers or are they just part of the design?

PR #712, opened 2018-04-23
Fix: replace <pre> with <code> to remove console errors

nimishjha:
## Description
The console was full of errors complaining that `<pre>` elements cannot be contained inside `<p>` elements. Also, we were using `<pre>` as inline elements in a lot of places, which it's not meant to do. This PR replaces all inline instances of `<pre>` with `<code>` elements. 

Errors in the console are bad.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Only the documentation has been changed, so it won't affect any functionality.

### Before

pre-in-p 01 before

### After

pre-in-p 02 after

## Check-list:


ChaoLiangSuper:
Merging .....

Issue #700, opened 2018-02-22
Proposal: New checkbox component to replace iCheck dependency


ChaoLiangSuper:
The iCheck lib is out of date and not been updated for months. So, we might need our own checkbox component.

## Types of issue
- [x] Replace existed component 
- [ ] Deprecate `react-icheck`

## Description

- Build a new checkbox component
- Remove the old dependency

![screen shot 2018-04-19 at 14 44 56](https://user-images.githubusercontent.com/908155/38971818-47452388-43e0-11e8-8291-2d456f33b601.png)

lightbringer1991:
if it is not so much different than iCheck, we can fork the library and maintain it ourselves (e.g. fixing bugs or implement new logic), without having to reinvent the wheel.

ChaoLiangSuper:
@lightbringer1991 Sorry for the late update, I didn't get the notification for your comment 😂. I think to redo the icheck component because we can also reduce the size of our project. For instance, those themes in icheck we don't need them.

omgaz:
Should try to support a painless migration from iCheck.
Should support radio and checkbox (returning a single value).
Should support radio and checkbox groups (returning an array or key/values).
Should support labels.
Should support disabled state.
Shouldn't support themes or style customisations.


Follow-up PR:
 - Should support NEW functionality of a tri-state.

ChaoLiangSuper:
PR got merged. Closing this issue.

omgaz:
This issue is not yet resolved, the radios have not yet been implemented as part of this issue - it's only halfway there.

lightbringer1991:
if @nimishjha is busy with something else, I can give it a try on Radio buttons.

lightbringer1991:
reopen since icheck has not been removed yet.

ChaoLiangSuper:
Closing this issue and open another issue for removing the iCheck lib.
https://github.com/Adslot/adslot-ui/issues/727

PR #716, opened 2018-05-02
[ADS-1046] Checkbox/CheckboxGroup components

nimishjha:

https://github.com/Adslot/adslot-ui/issues/700
https://adslot.atlassian.net/browse/ADS-1046

- [ ] Yes
- [ ] No

## Check-list:

### DIFF
--- a/docs/examples/CheckboxGroupExample.jsx
+++ b/docs/examples/CheckboxGroupExample.jsx
@@ -1,0 +1,64 @@
+import _ from 'lodash';
+import React from 'react';
+import Example from '../components/Example';
+import { CheckboxGroup, Checkbox } from '../../src';
+
+const onChangeGroup = (value, event, name) => {
+  _.noop();
+};
+
+const onChangeIndividual = (value, event, name) => {
+  _.noop();
+};
+
+class CheckboxGroupExample extends React.PureComponent {
+  render() {
+    return (
+      <CheckboxGroup name="movies" value={['terminator', 'predator']} onChange={onChangeGroup}>
+        <Checkbox label="The Terminator" value="terminator" onChange={onChangeIndividual} />
+        <Checkbox label="Predator" value="predator" />
+        <Checkbox label="The Sound of Music" value="soundofmusic" />
+      </CheckboxGroup>
+    );
+  }
+}
+
+const exampleProps = {
+  componentName: 'CheckboxGroup',
+  notes: 'Contains individual checkboxes. The state of child checkboxes is held in an array.',
+  exampleCodeSnippet: `<CheckboxGroup name="movies" value={['terminator', 'predator']} onChange={onChangeGroup}>
+  <Checkbox label="The Terminator" value="terminator" onChange={onChangeIndividual} />
+  <Checkbox label="Predator" value="predator" />
+  <Checkbox label="The Sound of Music" value="soundofmusic" />
+</CheckboxGroup>`,
+  propTypeSectionArray: [
+    {
+      label: '',
+      propTypes: [
+        {
+          propType: 'name',
+          type: 'string',
+        },
+        {
+          propType: 'value',
+          type: 'arrayOf(string: value, ...)',
+          note: "The strings must be the values of the group's child Checkboxes",
+        },
+        {
+          propType: 'children',
+          type: '<Checkbox /> elements',
+        },
+        {
+          propType: 'onChange',
+          type: 'Function',
+        },
+      ],
+    },
+  ],
+};
+
+export default () => (
+  <Example {...exampleProps}>
+    <CheckboxGroupExample />
+  </Example>
+);
--- a/src/components/adslot-ui/Checkbox/index.jsx
+++ b/src/components/adslot-ui/Checkbox/index.jsx
@@ -1,0 +1,89 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { expandDts } from 'lib/utils';
+
+import './styles.scss';
+
+class Checkbox extends React.Component {
+  static getDerivedStateFromProps(newProps, prevState) {
+    let isChecked = prevState.checked;
+    let isDisabled = prevState.disabled;
+    if (newProps.checked) isChecked = newProps.checked;
+    if (newProps.disabled) isDisabled = newProps.disabled;
+    return {
+      checked: isChecked,
+      disabled: isDisabled,
+    };
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      checked: props.checked,
+      disabled: props.disabled,
+    };
+    this.onChangeDefault = this.onChangeDefault.bind(this);
+    if (this.props.onChange) {
+      this.onChange = this.props.onChange.bind(this);
+    }
+  }
+
+  onChangeDefault(event) {
+    const isChecked = Boolean(event.target.checked);
+    const isDisabled = Boolean(event.target.disabled);
+    this.setState(() => ({
+      checked: isChecked,
+      disabled: isDisabled,
+    }));
+    if (this.onChange) {
+      this.onChange(event, this.props.name);
+    }
+  }
+
+  render() {
+    const { name, value, label, dts } = this.props;
+    const optional = {
+      id: this.props.id ? this.props.id : null,
+      className: this.props.className ? this.props.className : null,
+    };
+    if (this.props['data-name']) {
+      optional['data-name'] = this.props['data-name'];
+    }
+    return (
+      <label className="checkbox">
+        <input
+          type="checkbox"
+          name={name}
+          value={value}
+          onChange={this.onChangeDefault}
+          disabled={this.state.disabled}
+          checked={this.state.checked}
+          {...expandDts(dts)}
+          {...optional}
+        />
+        {label ? <span>{label}</span> : null}
+      </label>
+    );
+  }
+}
+
+Checkbox.propTypes = {
+  id: PropTypes.string,
+  className: PropTypes.string,
+  'data-name': PropTypes.string,
+  name: PropTypes.string,
+  label: PropTypes.node,
+  value: PropTypes.string,
+  dts: PropTypes.string,
+  disabled: PropTypes.bool,
+  checked: PropTypes.bool,
+  onChange: PropTypes.func,
+};
+
+Checkbox.defaultProps = {
+  dts: '',
+  disabled: false,
+  checked: false,
+};
+
+export default Checkbox;
--- a/src/components/adslot-ui/Checkbox/styles.scss
+++ b/src/components/adslot-ui/Checkbox/styles.scss
@@ -1,0 +1,14 @@
+.checkbox {
+  line-height: 16px;
+  display: block;
+
+  span {
+    margin: 0 0 0 5px;
+  }
+
+  input {
+    margin: 0;
+    line-height: 16px;
+    display: inline-block;
+  }
+}
--- a/src/components/adslot-ui/CheckboxGroup/index.jsx
+++ b/src/components/adslot-ui/CheckboxGroup/index.jsx
@@ -1,0 +1,72 @@
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const mapPropsToState = (props, state) => {
+  const tempState = state;
+  if (props.value) {
+    _.forOwn(state.value, (value, key) => {
+      tempState.value[key] = props.value.indexOf(key) !== -1;
+    });
+  }
+  return tempState;
+};
+
+class CheckboxGroup extends React.Component {
+  static getDerivedStateFromProps(nextProps, prevState) {
+    return mapPropsToState(nextProps, prevState);
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { value: {} };
+    React.Children.map(this.props.children, child => {
+      this.state.value[child.props.value] = false;
+    });
+    if (this.props.value) {
+      this.state.value = mapPropsToState(this.props, this.state.value);
+    }
+    this.renderChildren = this.renderChildren.bind(this);
+    this.onChangeDefault = this.onChangeDefault.bind(this);
+  }
+
+  onChangeDefault(event) {
+    const checkboxValue = event.target.value;
+    const tempValue = _.cloneDeep(this.state.value);
+    this.setState(() => {
+      tempValue[checkboxValue] = !this.state.value[checkboxValue];
+      if (this.props.onChange) {
+        this.props.onChange(tempValue, event, this.props.name);
+      }
+      return { value: tempValue };
+    });
+  }
+
+  renderChildren() {
+    return React.Children.map(this.props.children, child =>
+      React.cloneElement(child, {
+        name: this.props.name,
+        checked: this.state.value[child.props.value],
+        onChange: (...args) => {
+          if (child.props.onChange) child.props.onChange(...args);
+          this.onChangeDefault(...args);
+        },
+      })
+    );
+  }
+
+  render() {
+    const classes = `checkbox-group ${this.props.className}`;
+    return <div className={classes}>{this.renderChildren()}</div>;
+  }
+}
+
+CheckboxGroup.propTypes = {
+  name: PropTypes.string.isRequired,
+  value: PropTypes.arrayOf(PropTypes.string),
+  children: PropTypes.arrayOf(PropTypes.element).isRequired,
+  className: PropTypes.string,
+  onChange: PropTypes.func,
+};
+
+export default CheckboxGroup;
--- a/src/components/adslot-ui/CheckboxGroup/index.spec.jsx
+++ b/src/components/adslot-ui/CheckboxGroup/index.spec.jsx
@@ -1,0 +1,93 @@
+import React from 'react';
+import sinon from 'sinon';
+import { shallow, mount } from 'enzyme';
+import { Checkbox } from 'adslot-ui';
+import CheckboxGroup from '.';
+
+describe('CheckboxGroup', () => {
+  it('should render with props', () => {
+    const component = shallow(
+      <CheckboxGroup name="movies" value={['terminator', 'predator']} className="custom-class">
+        <Checkbox label="The Terminator" value="terminator" />
+        <Checkbox label="Predator" value="predator" />
+        <Checkbox label="The Sound of Music" value="soundofmusic" />
+      </CheckboxGroup>
+    );
+
+    expect(component.hasClass('custom-class')).to.equal(true);
+    const childCheckboxes = component.find('Checkbox');
+    expect(childCheckboxes.length).to.equal(3);
+    expect(component.state().value).to.eql({ terminator: true, predator: true, soundofmusic: false });
+  });
+
+  it('should handle checkbox change events', () => {
+    const onChangeGroup = sinon.spy();
+    const onChangeIndividual = sinon.spy();
+    const component = shallow(
+      <CheckboxGroup name="movies" value={['terminator', 'predator']} onChange={onChangeGroup}>
+        <Checkbox label="The Terminator" value="terminator" onChange={onChangeIndividual} />
+        <Checkbox label="Predator" value="predator" />
+        <Checkbox label="The Sound of Music" value="soundofmusic" />
+      </CheckboxGroup>
+    );
+
+    const childCheckboxes = component.find('Checkbox');
+    const firstChild = childCheckboxes.at(0);
+    const event = { target: { value: 'terminator', checked: false } };
+    firstChild.simulate('change', event);
+    expect(component.state().value).to.eql({ terminator: false, predator: true, soundofmusic: false });
+    expect(onChangeGroup.callCount).to.equal(1);
+    expect(onChangeIndividual.callCount).to.equal(1);
+  });
+
+  it('should render without onChange or value props', () => {
+    const component = shallow(
+      <CheckboxGroup name="movies">
+        <Checkbox label="The Terminator" value="terminator" />
+        <Checkbox label="Predator" value="predator" />
+        <Checkbox label="The Sound of Music" value="soundofmusic" />
+      </CheckboxGroup>
+    );
+
+    const childCheckboxes = component.find('Checkbox');
+    expect(childCheckboxes.length).to.equal(3);
+  });
+
+  it('should handle change events without a custom onChange handler', () => {
+    const event = { target: { value: 'terminator', checked: true } };
+    const component = shallow(
+      <CheckboxGroup name="movies">
+        <Checkbox label="The Terminator" value="terminator" />
+        <Checkbox label="Predator" value="predator" />
+        <Checkbox label="The Sound of Music" value="soundofmusic" />
+      </CheckboxGroup>
+    );
+    const firstChild = component.find('Checkbox').at(0);
+    firstChild.simulate('change', event);
+    expect(component.state().value).to.eql({ terminator: true, predator: false, soundofmusic: false });
+  });
+
+  it('should handle props changes', () => {
+    const component = mount(
+      <CheckboxGroup name="movies" value={['terminator', 'predator']} className="custom-class">
+        <Checkbox label="The Terminator" value="terminator" />
+        <Checkbox label="Predator" value="predator" />
+        <Checkbox label="The Sound of Music" value="soundofmusic" />
+      </CheckboxGroup>
+    );
+    component.setProps({ value: ['terminator', 'soundofmusic'] });
+    expect(component.state().value).to.eql({ terminator: true, predator: false, soundofmusic: true });
+  });
+
+  it('should handle props changes when no value is given', () => {
+    const component = mount(
+      <CheckboxGroup name="movies" value={['terminator', 'predator']} className="custom-class">
+        <Checkbox label="The Terminator" value="terminator" />
+        <Checkbox label="Predator" value="predator" />
+        <Checkbox label="The Sound of Music" value="soundofmusic" />
+      </CheckboxGroup>
+    );
+    component.setProps({});
+    expect(component.state().value).to.eql({ terminator: true, predator: true, soundofmusic: false });
+  });
+});
--- a/docs/components/Layout/index.jsx
+++ b/docs/components/Layout/index.jsx
@@ -14,6 +14,7 @@ import TextareaExample from '../../examples/TextareaExample';
 import TextEllipsisExample from '../../examples/TextEllipsisExample';
 import AlertExample from '../../examples/AlertExample';
 import CheckboxExample from '../../examples/CheckboxExample';
+import CheckboxGroupExample from '../../examples/CheckboxGroupExample';
 import RadioExample from '../../examples/RadioExample';
 import SelectExample from '../../examples/SelectExample';
 import DatePickerExample from '../../examples/DatePickerExample';
@@ -74,6 +75,7 @@ const componentsBySection = {
     'list-picker',
     'user-list-picker',
     'checkbox',
+    'checkbox-group',
     'radio',
     'select',
     'date-picker',
@@ -174,6 +176,7 @@ class PageLayout extends React.Component {
             <ListPickerExample />
             <UserListPickerExample />
             <CheckboxExample />
+            <CheckboxGroupExample />
             <RadioExample />
             <SelectExample />
             <DatePickerExample />
--- a/docs/examples/CheckboxExample.jsx
+++ b/docs/examples/CheckboxExample.jsx
@@ -1,41 +1,69 @@
+import _ from 'lodash';
 import React from 'react';
 import Example from '../components/Example';
 import { Checkbox } from '../../src';
 
+const onChange = (value, event, name) => {
+  _.noop();
+};
+
 class CheckboxExample extends React.PureComponent {
   render() {
-    return <Checkbox label="Agree to the terms and conditions." />;
+    return (
+      <Checkbox
+        name="Name goes here"
+        label="Label goes here"
+        value="Value goes here"
+        dts="data-test-selector-goes-here"
+        onChange={onChange}
+      />
+    );
   }
 }
 
 const exampleProps = {
   componentName: 'Checkbox',
-  designNotes: (
-    <p>
-      <span className="text-bold">Checkbox</span> used for making one or more selections from multiple options.
-    </p>
-  ),
-  exampleCodeSnippet: '<Checkbox label="Agree to the terms and conditions." />',
-  notes: (
-    <p>
-      See <a href="https://github.com/luqin/react-icheck">React iCheck Documentation</a>
-    </p>
-  ),
+  notes: '',
+  exampleCodeSnippet: `<Checkbox
+	name="Name goes here"
+	label="Label goes here"
+	value="Value goes here"
+	checked={true}
+	disabled={true}
+	dts="data-test-selector-goes-here"
+/>`,
   propTypeSectionArray: [
     {
+      label: '',
       propTypes: [
         {
+          propType: 'name',
+          type: 'string',
+        },
+        {
           propType: 'label',
-          type: 'node',
-          note: 'Usually fine to rely on a string but can pass HTML e.g. for a url.',
+          type: 'string',
         },
         {
           propType: 'value',
           type: 'string',
+          note: 'Required.',
+        },
+        {
+          propType: 'checked',
+          type: 'bool',
+        },
+        {
+          propType: 'disabled',
+          type: 'bool',
+        },
+        {
+          propType: 'dts',
+          type: 'string',
         },
         {
           propType: 'onChange',
-          type: 'func',
+          type: 'function',
         },
       ],
     },
--- a/src/components/adslot-ui/Checkbox/index.spec.jsx
+++ b/src/components/adslot-ui/Checkbox/index.spec.jsx
@@ -1,0 +1,102 @@
+import React from 'react';
+import { shallow, mount } from 'enzyme';
+import sinon from 'sinon';
+import Checkbox from '.';
+
+describe('Checkbox', () => {
+  it('should render with props', () => {
+    const component = shallow(
+      <Checkbox label="The Terminator" name="movies" value="terminator" dts="checkbox-terminator" />
+    );
+    const checkboxElement = component.find('input[type="checkbox"]');
+    const labelElement = component.find('label');
+    expect(labelElement.text()).to.equal('The Terminator');
+    expect(checkboxElement).to.have.length(1);
+    expect(checkboxElement.prop('checked')).to.equal(false);
+    expect(component.find('[name="movies"]')).to.have.length(1);
+    expect(component.find('[value="terminator"]')).to.have.length(1);
+    expect(component.find('[data-test-selector="checkbox-terminator"]')).to.have.length(1);
+  });
+
+  it('should render with just label', () => {
+    const component = shallow(<Checkbox label="Label goes here" />);
+    const checkboxElement = component.find('input[type="checkbox"]');
+    expect(checkboxElement).to.have.length(1);
+    const labelElement = component.find('label');
+    expect(labelElement.text()).to.equal('Label goes here');
+  });
+
+  it('should render with just onChange', () => {
+    const onChangeHandler = sinon.spy();
+    const component = shallow(<Checkbox onChange={onChangeHandler} />);
+    const checkboxElement = component.find('input[type="checkbox"]');
+    const event = { target: { checked: true, disabled: false } };
+    checkboxElement.simulate('change', event);
+    expect(onChangeHandler.callCount).to.equal(1);
+    expect(component.state()).to.eql({ checked: true, disabled: false });
+  });
+
+  it('should render with id, className, and data-name', () => {
+    const component = shallow(<Checkbox id="checkboxId" className="checkboxClass" data-name="checkboxName" />);
+    const checkboxElement = component.find('input[type="checkbox"]');
+    expect(checkboxElement.hasClass('checkboxClass')).to.equal(true);
+    expect(component.find('#checkboxId')).to.have.length(1);
+    expect(component.find('[data-name="checkboxName"]')).to.have.length(1);
+  });
+
+  it('should handle change event', () => {
+    const onChangeHandler = sinon.spy();
+    const component = shallow(
+      <Checkbox
+        label="The Terminator"
+        name="movies"
+        value="terminator"
+        onChange={onChangeHandler}
+        dts="checkbox-terminator"
+      />
+    );
+    const checkboxElement = component.find('input[type="checkbox"]');
+    const event = { target: { checked: true, disabled: true } };
+    checkboxElement.simulate('change', event);
+    expect(onChangeHandler.callCount).to.equal(1);
+    expect(component.state()).to.eql({ checked: true, disabled: true });
+  });
+
+  it('should render without a label', () => {
+    const component = shallow(<Checkbox name="movies" value="terminator" />);
+    const labelElement = component.find('label');
+    expect(labelElement.text()).to.equal('');
+  });
+
+  it('should render with checked and disabled states', () => {
+    const component = shallow(
+      <Checkbox
+        name="movies"
+        value="terminator"
+        checked={true} // eslint-disable-line
+        disabled={true} // eslint-disable-line
+      />
+    );
+    expect(component.state()).to.eql({ checked: true, disabled: true });
+  });
+
+  it('should handle change events without a custom onChange handler', () => {
+    const component = shallow(<Checkbox name="movies" value="terminator" />);
+    const checkboxElement = component.find('input[type="checkbox"]');
+    const event = { target: { checked: true, disabled: false } };
+    checkboxElement.simulate('change', event);
+    expect(component.state()).to.eql({ checked: true, disabled: false });
+  });
+
+  it('should handle props changes', () => {
+    const component = mount(<Checkbox name="movies" value="terminator" />);
+    component.setProps({ checked: true, disabled: true });
+    expect(component.state()).to.eql({ checked: true, disabled: true });
+  });
+
+  it('should handle props changes when no relevant props are given', () => {
+    const component = mount(<Checkbox name="movies" value="terminator" />);
+    component.setProps({ foo: 'bar' });
+    expect(component.state()).to.eql({ checked: false, disabled: false });
+  });
+});
--- a/src/components/adslot-ui/index.js
+++ b/src/components/adslot-ui/index.js
@@ -1,6 +1,8 @@
 import Accordion from 'adslot-ui/Accordion';
 import AlertInput from 'adslot-ui/AlertInput';
 import Carousel from 'adslot-ui/Carousel';
+import Checkbox from 'adslot-ui/Checkbox';
+import CheckboxGroup from 'adslot-ui/CheckboxGroup';
 import ConfirmModal from 'adslot-ui/ConfirmModal';
 import HelpIconPopover from 'adslot-ui/HelpIconPopover';
 import FilePicker from 'adslot-ui/FilePicker';
@@ -28,6 +30,8 @@ export {
   Accordion,
   AlertInput,
   Carousel,
+  Checkbox,
+  CheckboxGroup,
   ConfirmModal,
   fastStatelessWrapper,
   FilePicker,
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,4 @@
 // Export the consumable components.
-import Checkbox from 'react-icheck/lib/Checkbox';
 import Radio from 'react-icheck/lib/Radio';
 import RadioGroup from 'react-icheck/lib/RadioGroup';
 import Select from 'react-select';
@@ -52,6 +51,8 @@ import {
   Accordion,
   AlertInput,
   Carousel,
+  Checkbox,
+  CheckboxGroup,
   ConfirmModal,
   fastStatelessWrapper,
   FilePicker,
@@ -87,6 +88,7 @@ export {
   Card,
   Carousel,
   Checkbox,
+  CheckboxGroup,
   ConfirmModal,
   Dropdown,
   DatePicker,


nimishjha:
Will change `CheckboxSingle` to `CheckBox` when it's good to go. Also, please ignore the console logs for now.

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+class CheckboxSingle extends React.Component {
+	constructor(props) {
+		super(props);
+		this.state = {
+      checked: props.checked,
+      disabled: props.disabled,
+      name: props.name,
+      value: props.value,
+      dts: props.dts,
+      label: props.label,
```
omgaz: state, dts, and name don't need to be managed by state they should remain in the props.

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+import './styles.scss';
+
+class CheckboxSingle extends React.Component {
+	constructor(props) {
+		super(props);
+		this.state = {
+      checked: props.checked,
+      disabled: props.disabled,
+      name: props.name,
```
omgaz: Do we need name?

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+		return(
+  <div className='checkbox-single'>
+    <label>
+      <input
+        type='checkbox'
+        name={this.state.name}
+        value={this.state.value}
+        onChange={onChangeHandler}
+        data-test-selector={this.state.dts}
```
omgaz: We have an `expandDts` helper we can use for this.

src/components/adslot-ui/CheckboxSingle/styles.scss
```diff
@@ -0,0 +1,9 @@
+.checkbox-single label {
+	line-height: 20px;
```
omgaz: This seems too much, I think it might be 16px - let me check.

omgaz: Yup, can you make this 16px to match the w/h of the checkbox

src/components/adslot-ui/CheckboxSingle/styles.scss
```diff
@@ -0,0 +1,9 @@
...
+}
+
+.checkbox-single input {
+	margin: 2px 5px 0 0;
```
omgaz: As a label can be optional, we should only add a margin on the right if there's one present. Maybe wrap the label text in a span and add a margin-left to that span?

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+  label: PropTypes.string,
+  value: PropTypes.string.isRequired,
+  dts: PropTypes.string,
+  disabled: PropTypes.bool,
+  checked: PropTypes.bool,
+  onChangeCallback: PropTypes.func,
+};
+
+CheckboxSingle.defaultProps = {
+  label: '',
```
omgaz: A label is optional and should be allowed to be undefined and shouldn't default to ''.

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+      {this.state.label}
+    </label>
+  </div>
+		);
+	}
+}
+
+CheckboxSingle.propTypes = {
+  name: PropTypes.string,
+  label: PropTypes.string,
```
omgaz: We should optionally allow for <node> so we can pass through formatted html e.g. `<span>Required <HelpIconPopover>This field is required</HelpIconPopover></span>`

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './styles.scss';
+
+class CheckboxSingle extends React.Component {
+	constructor(props) {
+		super(props);
+		this.state = {
+      checked: props.checked,
```
omgaz: Just some small indentation commit issues.

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+class CheckboxGroup extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: this.props.value || [],
+    };
+    this.renderChildren = this.renderChildren.bind(this);
+    this.onChangeCallback = this.onChangeCallback.bind(this);
```
omgaz: The arrow function binding issue for Adslot UI should have been merged, I think we can adopt the new style now:

https://adslot.atlassian.net/wiki/spaces/APT/pages/178225251/React

Can you have a look, if it fails to work we should re-open the issue.

omgaz:
Getting there! Good effort with your first React Component.

src/components/adslot-ui/CheckboxSingle/index.jsx
nimishjha: We need it for usage in a group, because in that case name will be shared across all child checkboxes.

src/components/adslot-ui/CheckboxSingle/styles.scss
nimishjha: Done

src/components/adslot-ui/CheckboxSingle/styles.scss
nimishjha: Done

src/components/adslot-ui/CheckboxSingle/index.jsx
nimishjha: Done

src/components/adslot-ui/CheckboxSingle/index.jsx
nimishjha: Done

src/components/adslot-ui/CheckboxSingle/index.jsx
nimishjha: Done

src/components/adslot-ui/CheckboxSingle/index.jsx
nimishjha: Done

src/components/adslot-ui/CheckboxSingle/index.jsx
nimishjha: Fixed

src/components/adslot-ui/CheckboxGroup/index.jsx
lightbringer1991: it should be working. We already have some code that adapt this style, I think.

docs/examples/CheckboxGroupExample.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import Example from '../components/Example';
+import { CheckboxSingle } from '../../src';
+import { CheckboxGroup } from '../../src';
```
mdotwills: this can be just `import { CheckboxSingle, CheckboxGroup } from '../../src';`

src/components/adslot-ui/CheckboxGroup/index.jsx
nimishjha: It's not, I get a "Parsing error: unexpected token ="

docs/examples/CheckboxGroupExample.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+const Konsole = window.console;
```
mdotwills: is this a special or wrapped `console.log`?

nimishjha: Just an alias, to get around the linter. Only for development purposes, will be gone soon.

docs/examples/CheckboxGroupExample.jsx
nimishjha: Yep, fixing.

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  onChangeCallback(checkboxParams, event) {
```
mdotwills: Maybe for consistency we can drop the `Callback` in the function name here, we don't do it anywhere else. We can just have it as `onChange` (see [here](https://github.com/Adslot/adslot-ui/search?utf8=%E2%9C%93&q=onchange&type=) and specifically [here](https://github.com/Adslot/adslot-ui/blob/master/src/components/adslot-ui/Search/index.jsx#L17)) 👍

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+    let newValue;
+    if(checkboxParams.checked){
+      newValue = this.state.value.concat([checkboxParams.value]);
```
mdotwills: I see two issues here, the first being we are modifying the `state` directly here, which is not permissible to do outside of the `constructor`. The other issue is that we are setting the state based on the previous value of the same state. Due to the way React bunches state updates, if we need to know the previous state during `setState`, we need to use the function call of `setState` to guarantee we are operating on the latest updated state before setting it:

```javascript
if(checkboxParams.checked) {
  this.setState((prevState) => ({ value: prevState.concat([checkboxParams.value] }));
} else {
  this.setState(prevState) => ({ value: prevState.filter(currentValue => currentValue !== checkboxParams.value) }));
}
```
(just double check the logic there)

mdotwills: also notice if you want that you can write the `(prevState) => ({ value: prevState.concat([checkboxParams.value] })` part as a function, so

```javascript
const updater = (prevState) => ({ value: prevState.concat([checkboxParams.value] });
```
and then later
```javascript
this.setState(updater);
```
(obv. better names are needed than this example haha)

src/components/adslot-ui/CheckboxGroup/index.jsx
nimishjha: There's already an `onChange` function which is always called, the `onChangeCallback` is optional and is only called when it's provided. Can rename this to something else, open to suggestions.

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,93 @@
...
+    if(this.state.disabled) {
+      booleanAttributes.disabled = 'disabled';
+    }
+    if(this.state.checked) {
+      booleanAttributes.checked = 'checked';
+    }
+
+    if(label)
+    {
+  		return(
```
mdotwills: I think the linter might not have picked up the inconsistent indentation here

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+  onChange(event) {
+    const isChecked = Boolean(event.target.checked);
+    const isDisabled = Boolean(event.target.disabled);
+    this.setState({
+      checked: isChecked,
+      disabled: isDisabled,
+    });
+    if(this.onChangeCallback) {
+      this.onChangeCallback({value: this.state.value, checked: isChecked}, event, this.props.name);
```
mdotwills: If you needed to guarantee that `this.state.value` was completely up-to-date, you could move this to be inside of the callback after the object argument version of `setState`.

```javascript
this.setState({
  checked: isChecked,
  disabled: isDisabled,
}, () => {
  if(this.onChangeCallback) {
    this.onChangeCallback({ value: this.state.value, checked: isChecked }, event, this.props.name);
  }
});
```
My understanding is that if there are bunched state updates waiting from other `setState` calls, this callback will capture the updated state from all updates.
_(Note: callback may fire even if the state does not update)._ But given we're not updating `this.state.value` here, it should be okay as is. Just something to consider when debugging.

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+        {this.renderChildren()}
+      </div>
+    );
+  }
+}
+
+CheckboxGroup.propTypes = {
+  name: PropTypes.string.isRequired,
+  value: PropTypes.arrayOf(PropTypes.string),
+	children: PropTypes.arrayOf(PropTypes.element).isRequired,
```
mdotwills: space req

src/components/adslot-ui/CheckboxGroup/index.jsx
mdotwills: ^ did you run `npm i`?

src/components/adslot-ui/CheckboxGroup/index.jsx
nimishjha: `concat` doesn't mutate the existing arrays - it returns a new array, so `state` is not being modified there. The `state` is updated in a `setState` call on line 28.

src/components/adslot-ui/CheckboxSingle/index.jsx
nimishjha: Yeah, I'm aware of that.

src/components/adslot-ui/CheckboxGroup/index.jsx
mdotwills: good point about `concat` not mutating state. But the latter point still remains, you need to call `setState` with a function and not an object.

src/components/adslot-ui/CheckboxGroup/index.jsx
nimishjha: ?

mdotwills: screen shot 2018-05-08 at 10 40 56 am

nimishjha: Hmm, no idea where the random tab came from. Fixed.

src/components/adslot-ui/CheckboxGroup/index.jsx
nimishjha: Actually, no, because I'm not accessing the current value of `state` in `setState` at all.

src/components/adslot-ui/CheckboxGroup/index.jsx
nimishjha: Yes. I think @omgaz mentioned yesterday that the arrow function update had been applied to direct-web, but not adslot-ui.

src/components/adslot-ui/CheckboxGroup/index.jsx
mdotwills: @nimishjha you are, via `newValue` which depends on `this.state.value`

nimishjha: Yes, but I'm not accessing current state *inside* setState...

nimishjha: Went ahead and changed to function syntax for `setState` anyway. No harm in following best practices...

src/components/adslot-ui/CheckboxGroup/index.jsx
lteacher: Definitely the plugin isnt there. I added randomly in some PR. Would be nice to have though in another PR maybe I should do that.

src/components/adslot-ui/CheckboxGroup/index.jsx
lteacher: Im prob not a fan so much of the extra `onChangeCallback` as well mostly just because its an extra prop that shouldn't be externally used by the consumer as you just need it for the group. So um I guess my comment is really for the actual `Checkbox`... 

I did have some trouble seeing where the `onChange` sits on the `Checkbox` itself though, seems like there is no actual prop there?

src/components/adslot-ui/CheckboxGroup/index.jsx
lteacher: Does seem like @mdotwills has a valid point though also seems like there are a few components everywhere in direct-web and maybe here too that are not exactly correct or could suffer from the issue.

I mean the documentation doesn't refer to anything being done in the anon function execution of `setState`, it specifically refers to a `.state`.

This example from the documentation is using the state in the current scope
```javascript
this.setState({
  // this.state.counter is in the outer scope and being applied to the object passed to setState
  counter: this.state.counter + this.props.increment,
});
```

src/components/adslot-ui/CheckboxGroup/index.jsx
nimishjha: For the individual checkbox, there's a default onChange that's always called, to update its state, and if an onChangeCallback has been provided, whether from a parent CheckboxGroup or another component, onChange() calls it after it's finished executing. I'm open to suggestions for the function names, or if someone has a better approach to solve the problem, that's also welcome.

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+}
+
+CheckboxSingle.propTypes = {
+  name: PropTypes.string,
+  label: PropTypes.node,
+  value: PropTypes.string.isRequired,
+  dts: PropTypes.string,
+  disabled: PropTypes.bool,
+  checked: PropTypes.bool,
+  onChangeCallback: PropTypes.func,
```
lteacher: Per other discussion I was just wondering about this `onChangeCallback` here as I was expecting the consumer could provide an `onChange` but instead they must provide an `onChangeCallback` to get anything meaningful out of it. If that is the only function then the prop should be `onChange` and change other internal to something else is my thought.

nimishjha: Renamed default onChange to onChangeDefault, and user-provided one to onChange.

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+      this.props.onChange(newValue, event, this.props.name);
+    }
+  }
+
+  renderChildren() {
+    return React.Children.map(this.props.children, (child) =>
+      React.cloneElement(child, {
+        name: this.props.name,
+        checked: this.state.value.indexOf(child.props.value) >= 0,
+        onChange: this.onChangeDefault,
```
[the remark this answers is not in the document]
nimishjha: Yeah, that's a potential use case. Suggestions for function name? `onChangeCustom`?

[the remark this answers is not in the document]
lteacher: Yeah i dunno i mean prob can put where but it should be straight-forward i guess like:

```javascript
renderChildren() {
  return React.Children.map(this.props.children, (child) =>
    React.cloneElement(child, {
      name: this.props.name,
      checked: this.state.value.indexOf(child.props.value) >= 0,
      onChange: (...args) => {
        if (child.props.onChange) child.props.onChange(...args);

        this.onChangeDefault(...args)
      },
    }));
}
```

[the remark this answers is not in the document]
nimishjha: Clever, will do.

ChaoLiangSuper:
I think this file `src/components/adslot-ui/CheckboxGroup/styles.scss` is empty?

docs/examples/CheckboxGroupExample.jsx
```diff
@@ -0,0 +1,66 @@
+import _ from 'lodash';
+import React from 'react';
+import Example from '../components/Example';
+import { CheckboxGroup, CheckboxSingle } from '../../src';
+
+const onChangeGroup = function(value, event, name) {
```
mdotwills: Can use the `const onChangeGroup = (value, event, name) => {` syntax here instead? We don't seem to be using this style [anywhere else in the repo](https://github.com/Adslot/adslot-ui/search?p=1&q=function&type=&utf8=%E2%9C%93).

docs/examples/CheckboxGroupExample.jsx
```diff
@@ -0,0 +1,66 @@
...
+  _.noop();
+};
+
+const onChangeIndividual = function(value, event, name) {
```
mdotwills: `onChangeSingle` perhaps to align with your other naming convention?

docs/examples/CheckboxGroupExample.jsx
lightbringer1991: also it looks trivial I guess, can just use `_.noop` directly. E.g.
```jsx
<CheckboxGroup name="movies" value={['terminator', 'predator']} onChange={_.noop}></CheckboxGroup>
```

docs/examples/CheckboxGroupExample.jsx
nimishjha: CheckboxSingle will be renamed to Checkbox before merging, so...

docs/examples/CheckboxGroupExample.jsx
nimishjha: @mdotwills will do. @lightbringer1991 I wanted to make it clear what parameters the change handler receives.

src/components/adslot-ui/CheckboxGroup/index.spec.jsx
```diff
@@ -0,0 +1,55 @@
...
+        <CheckboxSingle label="Predator" value="predator" />
+        <CheckboxSingle label="The Sound of Music" value="soundofmusic" />
+      </CheckboxGroup>
+    );
+
+    const childCheckboxes = component.find('CheckboxSingle');
+    expect(childCheckboxes.length).to.equal(3);
+  });
+
+  it('should render without onChange or value props', () => {
```
mdotwills: this test name is the same as the one before. Can we be more specific and differentiate these test descriptions?

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+import PropTypes from 'prop-types';
+
+import './styles.scss';
+
+class CheckboxGroup extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: this.props.value || [],
+    };
```
lightbringer1991: just my 2 cents here.

Would it be easier to have the value state as follow
```js
value = {
  '<checkbox name 1>': true,
  '<checkbox name 2>': false,
  ...
};
```

This way, all you need to do for the onChange is just toggling the corresponding attribute. And also simplify this line of code `this.state.value.indexOf(child.props.value) >= 0` into `this.state.value[child.props.value]`.

Only down side is that you will need to construct the thing when the list of children checkbox changes.

I don't mind the array approach though.

src/components/adslot-ui/CheckboxGroup/index.spec.jsx
```diff
@@ -0,0 +1,55 @@
...
+        <CheckboxSingle label="The Sound of Music" value="soundofmusic" />
+      </CheckboxGroup>
+    );
+
+    const childCheckboxes = component.find('CheckboxSingle');
+    expect(childCheckboxes.length).to.equal(3);
+
+    const firstChild = childCheckboxes.at(0);
+    const event = { value: 'terminator', checked: false };
+    firstChild.simulate('change', event);
```
mdotwills: This test can be split to 'should render correctly' and 'should handle checkbox change events'. Makes it easier to narrow down issues when a test fails.
- first test up to L20 ensures necessary components are there and no crash
- second test performs the `simulate` test.

src/components/adslot-ui/CheckboxGroup/index.jsx
nimishjha: Yes, I considered that, but it's not like we'll ever have CheckboxGroups with a million child checkboxes, so `array.indexOf`won't make any difference to performance, and as you noted, you'd need to change the array every time you want to change the value of the group. This way is easier to use.

src/components/adslot-ui/CheckboxGroup/index.spec.jsx
nimishjha: OK

src/components/adslot-ui/CheckboxSingle/index.spec.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+  it('should render with checked and disabled states', () => {
+    const onChangeHandler = sinon.spy();
+    const component = shallow(
+      <CheckboxSingle
+        name="movies"
+        value="terminator"
+        onChange={onChangeHandler}
+        checked={true} // eslint-disable-line
+        disabled={true} // eslint-disable-line
```
mdotwills: maybe we need to update our linter rule here?

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+  dts: PropTypes.string,
+  disabled: PropTypes.bool,
+  checked: PropTypes.bool,
+  onChange: PropTypes.func,
+};
+
+CheckboxSingle.defaultProps = {
+  dts: '',
+  disabled: false,
+  checked: false,
```
lightbringer1991: it doesn't look like you're using `disabled` and `checked` from props, except from constructor.
Does that mean the props are only useful at first pass, and then get discarded afterwards?

src/components/adslot-ui/CheckboxSingle/index.spec.jsx
nimishjha: It if can be configured to only check for boolean attributes on actual HTML5 elements, then sure. Disabling the rule entirely is probably not a good idea, because the boolean attribute rules are counterintuitive and it's good to have a linter rule enforcing them.

src/components/adslot-ui/CheckboxSingle/styles.scss
```diff
@@ -0,0 +1,17 @@
...
+  }
+
+  input {
+    margin: 0;
+    line-height: 16px;
+    display: inline-block;
+  }
+}
+
+
```
lightbringer1991: I guess we can remove one empty space

src/components/adslot-ui/CheckboxSingle/index.spec.jsx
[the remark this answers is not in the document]
mdotwills: ^ yea that's true. I read about a draw back of this approach somewhere though but I can't find a link to what I was reading.

docs/examples/CheckboxSingleExample.jsx
```diff
@@ -0,0 +1,78 @@
...
+          type: 'string',
+        },
+        {
+          propType: 'label',
+          type: 'string',
+        },
+        {
+          propType: 'value',
+          type: 'string',
+        },
```
lightbringer1991: seems like `value` is required, so maybe we should indicate that in the table.

nimishjha: Good catch, fixed.

src/components/adslot-ui/CheckboxSingle/index.jsx
nimishjha: Yeah, the props just give it the initial state (checked and/or disabled), subsequent changes are handled onChange.

nimishjha: Good catch, I need to handle that use case.

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+    const classes = `checkbox-group ${this.props.customClasses}`;
+    return <div className={classes}>{this.renderChildren()}</div>;
+  }
+}
+
+CheckboxGroup.propTypes = {
+  name: PropTypes.string.isRequired,
+  value: PropTypes.arrayOf(PropTypes.string),
+  children: PropTypes.arrayOf(PropTypes.element).isRequired,
+  customClasses: PropTypes.string,
```
lteacher: I wish we had a consistent name for these classes. I looked through the project though and couldn't see any solid naming. I'm wondering most of the time why these are not simply `className` which then includes some class since you cant even give `className` anyway to override the provided `checkbox-group` class so would it matter? 

I just think the thing should be consistent for all components and since the components are used like elements the right thing is just `className` in the same way all the react-bootstrap ones are `className`. But anyway, that isn't done consistently as mentioned so.... @omgaz

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+      disabled: isDisabled,
+    }));
+    if (this.onChange) {
+      this.onChange({ value: this.props.value, checked: isChecked }, event, this.props.name);
+    }
+  }
+
+  render() {
+    const { name, value, label, dts } = this.props;
+    const onChangeHandler = this.onChangeDefault;
```
lteacher: Why bother to reassign this `onChangeDefault` to `onChangeHandler` instead of just using `onChangeDefault` directly or naming it `onChangeHandler` if preferred?

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+    const booleanAttributes = { disabled: '', checked: '' };
+    if (this.state.disabled) {
+      booleanAttributes.disabled = 'disabled';
+    }
+    if (this.state.checked) {
+      booleanAttributes.checked = 'checked';
+    }
```
lteacher: Just wondering why using this particular pattern, was there some kind of lint or react warning? Needs like 6 extra lines to do this block instead of just setting the props like:

```html
<input
  disabled={this.state.disabled}
  checked={this.state.checked}
/>
```
Unless i missed something I thought that syntax is ok.

src/components/adslot-ui/CheckboxSingle/index.spec.jsx
lteacher: I think we disabled the linter rule because other people prefer not having to specify `true`/`false`

src/components/adslot-ui/CheckboxGroup/index.jsx
lightbringer1991: I would like to have it as `className`, so it's more consistent with what we already have.

src/components/adslot-ui/CheckboxSingle/index.spec.jsx
```diff
@@ -0,0 +1,58 @@
...
+      />
+    );
+    expect(component.state()).to.eql({ checked: true, disabled: true });
+  });
+
+  it('should render without a custom onChange handler', () => {
+    const component = shallow(<CheckboxSingle name="movies" value="terminator" />);
+    const checkboxElement = component.find('input[type="checkbox"]');
+    const event = { target: { name: 'movies', value: 'terminator' } };
+    checkboxElement.simulate('change', event);
```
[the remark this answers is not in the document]
lteacher: Hmm yeah does seem like the coverage is correct about this. But is troublesome to test I guess in this case. I suppose since `this.onChange` is dynamically set then you can actually check that its undefined on the `component.instance()`

For sure instead of the comment just change the test to:
_'should simulate the onChange event without a custom onChange handler'_

And add test to ensure as above, the instance `onChange` is undefined and the `onChangeDefault` was called once.

src/components/adslot-ui/CheckboxSingle/index.jsx
lteacher: Though if you do change to consider the props the logic obviously cant just use state.

src/components/adslot-ui/CheckboxGroup/index.spec.jsx
```diff
@@ -0,0 +1,67 @@
...
+    const event = { value: 'terminator', checked: true };
+    const component = shallow(
+      <CheckboxGroup name="movies">
+        <CheckboxSingle label="The Terminator" value="terminator" />
+        <CheckboxSingle label="Predator" value="predator" />
+        <CheckboxSingle label="The Sound of Music" value="soundofmusic" />
+      </CheckboxGroup>
+    );
+    const firstChild = component.find('CheckboxSingle').at(0);
+    firstChild.simulate('change', event);
```
lteacher: There are no expectations in the test similar to other comment I think we should aim to expect something even if it is to test that an `onChange` fires and doesn't fail _(the test description)_ because its looking at an internal definition. That might be why its getting picked up in the coverage check because of the dynamic assignment.

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      checked: props.checked,
+      disabled: props.disabled,
+    };
+    this.onChangeDefault = this.onChangeDefault.bind(this);
+    if (this.props.onChange) {
+      this.onChange = this.props.onChange.bind(this);
+    }
```
lteacher: Actually I think you could maybe ditch some of this right with the binding if you moved the check, though there is already a check and im thinking the binding probably isnt required too for this on `props`.

Because `this.onChange` is called in a bound function I assume it would not need this additional binding, but specifically I am asking if what about just checking later in the `onChangeDefault` body

```javascript
if (this.props.onChange) {
  this.props.onChange({ value: this.props.value, checked: isChecked }, event, this.props.name);
}
```

Then you dont need _(but probably dont need the binding anyway)_ the dynamic assignment and just costs the extra `.props` characters which is minimal

src/components/adslot-ui/CheckboxGroup/index.spec.jsx
nimishjha: The reason it *was* getting picked up is because if you have an if condition without an else, you still need to test for the "else" case of that if condition. There's a good discussion [here](https://github.com/gotwarlost/istanbul/issues/35). I've added an expect clause to the test and changed the test description.

src/components/adslot-ui/CheckboxSingle/index.jsx
lteacher: Then regarding the other test related comments you should be able to have:

`it should execute the onChangeDefault method onChange` or similar wording for without the `onChange` prop

And then 

`it should execute the onChangeDefault and call the custom onChange if proviced` for the second part. Then you have the two branches of coverage. On the first check the calledOnce on the `onChangeDefault` and on the second check the calledOnce on both the `onChangeDefault` and the provided prop

nimishjha: Once that arrow function babel plugin PR gets merged someone can go through and replace all the .bind() calls with => notations.

src/components/adslot-ui/CheckboxGroup/index.spec.jsx
lteacher: I checked the discussion the OP there is not correct its about perspective. These tests are testing the functionality of the `onChange` simulation not the render as mentioned in the test description originally. So you must test both branches for at least positive testing.

That simply means to test that the `onChangeDefault` is called for the no prop case which then will trigger the else branch, then to check `onChangeDefault` and the `props.onChange` which is passed in externally which hits the if block. Before it wasn't tested, then adding this above covered the logical flow but tests should expect some outcome so of course its best practice to expect some kind of result even if its just the successful execution of some function without error.

src/components/adslot-ui/CheckboxSingle/index.jsx
lteacher: I think there is some delay now on that because this repo uses a deprecated package in `karma-coverage` so I didn't get that one finished.

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,98 @@
...
+
+class CheckboxSingle extends React.Component {
+  static getDerivedStateFromProps(newProps, prevState) {
+    const isChecked = newProps.checked || prevState.checked;
+    const isDisabled = newProps.disabled || prevState.disabled;
+    return {
+      checked: isChecked,
+      disabled: isDisabled,
+    };
+  }
```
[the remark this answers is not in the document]
nimishjha: Good catch, I need to fix the logic.

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+      disabled: isDisabled,
+    }));
+    if (this.onChange) {
+      this.onChange({ value: this.props.value, checked: isChecked }, event, this.props.name);
+    }
+  }
+
+  render() {
+    const { name, value, label, dts } = this.props;
+    if (label) {
```
xiaofan2406: this conditional re-render will cause the input to be unmounted and lost its all native values, because the tree is different

Should render the same structure, with or without the label, and do the conditional check in L57 `{label ? <span>{label}</span>} : null`

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+    );
+  }
+}
+
+CheckboxSingle.propTypes = {
+  name: PropTypes.string,
+  label: PropTypes.node,
+  value: PropTypes.string.isRequired,
+  dts: PropTypes.string,
+  disabled: PropTypes.bool,
```
xiaofan2406: the `disabled` input attribute is fixed with `this.state.disabled`, then the only change it can happen is in `onChangeDefault` which reads from input attribute.

Meaning that, you can remove `disabled` state completely,  and just use `disabled` prop

src/components/adslot-ui/CheckboxSingle/index.jsx
nimishjha: Good to know, thanks.

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+class CheckboxGroup extends React.Component {
+  static getDerivedStateFromProps(newProps) {
+    if (newProps.value) return { value: newProps.value };
+    return false;
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: this.props.value || [],
```
xiaofan2406: use default props. and remove the `||` here

src/components/adslot-ui/CheckboxSingle/index.jsx
nimishjha: Not sure what you mean...

src/components/adslot-ui/CheckboxGroup/index.jsx
nimishjha: Actually, have to keep it this way, or the defaultProps override the props when you pass them in with value undefined.

docs/examples/CheckboxGroupExample.jsx
```diff
@@ -0,0 +1,66 @@
...
+const exampleProps = {
+  componentName: 'CheckboxGroup',
+  notes: 'Contains individual checkboxes. The state of child checkboxes is held in an array.',
+  exampleCodeSnippet: `
+      <CheckboxGroup name="movies" value={['terminator', 'predator']} onChange={onChangeGroup}>
+        <CheckboxSingle label="The Terminator" value="terminator" onChange={onChangeIndividual} />
+        <CheckboxSingle label="Predator" value="predator" />
+        <CheckboxSingle label="The Sound of Music" value="soundofmusic" />
+      </CheckboxGroup>
+  `,
```
omgaz: Kinda good kinda bad way in which back-ticks work across multi-line. They're great because they allow multi-line strings but they also mean we get string literal line breaks for the start and end (as well as spaces).

Can we update so that it'll render correctly as:

```
  exampleCodeSnippet: `<CheckboxGroup name="movies" value={['terminator', 'predator']} onChange={onChangeGroup}>
  <CheckboxSingle label="The Terminator" value="terminator" onChange={onChangeIndividual} />
  <CheckboxSingle label="Predator" value="predator" />
  <CheckboxSingle label="The Sound of Music" value="soundofmusic" />
</CheckboxGroup>`
```

so the html rendering looks like:


```jsx
<CheckboxGroup name="movies" value={['terminator', 'predator']} onChange={onChangeGroup}>
  <CheckboxSingle label="The Terminator" value="terminator" onChange={onChangeIndividual} />
  <CheckboxSingle label="Predator" value="predator" />
  <CheckboxSingle label="The Sound of Music" value="soundofmusic" />
</CheckboxGroup>
```

docs/examples/CheckboxGroupExample.jsx
```diff
@@ -0,0 +1,66 @@
...
+    {
+      label: '',
+      propTypes: [
+        {
+          propType: 'name',
+          type: 'string',
+        },
+        {
+          propType: 'value',
+          type: 'Array of strings',
```
omgaz: It looks like we use the `arrayOf` terminology. e.g. `arrayOf(string)` although this doesn't seem to be consistent.

docs/examples/CheckboxSingleExample.jsx
```diff
@@ -0,0 +1,79 @@
...
+  exampleCodeSnippet: `
+    <CheckboxSingle
+      name="Name goes here"
+      label="Label goes here"
+      value="Value goes here"
+      checked={true}
+      disabled={true}
+      dts="data-test-selector-goes-here"
+    />
+  `,
```
omgaz: ditto with the example snippet indentation.

src/components/adslot-ui/CheckboxGroup/index.jsx
omgaz: Can you leave that in a comment so someone doesn't unintentionally refactor this out.

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+    };
+    this.renderChildren = this.renderChildren.bind(this);
+    this.onChangeDefault = this.onChangeDefault.bind(this);
+  }
+
+  onChangeDefault(checkboxParams, event) {
```
omgaz: Why do we need checkboxParams here, doesn't the event/prevState/state give us everything we need? We have onChangeDefault in the single checkbox only accept event which seems more straightforward.

src/components/adslot-ui/CheckboxGroup/index.jsx
omgaz: I disagree and your setState below proves array.indexOf isn't usable in the solution (i.e having to filter by value rather than knowing the index and having an if/else block). Tuan's suggestion of an associative array creates a consistent way of handling the data through look-up rather than having to rely later on to do a filter based on value.

We'd need a `getter` to allow us to return as an array, and have the constructor be a little bit smarter at setting the internal state from the provided value but it'd really clean up the setState function below.

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+    let newValue;
+
+    this.setState(prevState => {
+      if (checkboxParams.checked) {
+        newValue = prevState.value.concat([checkboxParams.value]);
+      } else {
+        newValue = prevState.value.filter(currentValue => currentValue !== checkboxParams.value);
```
omgaz: as we can't rely on array.indexOf an associative array with lookup would greatly help

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+    };
+    this.renderChildren = this.renderChildren.bind(this);
+    this.onChangeDefault = this.onChangeDefault.bind(this);
+  }
+
+  onChangeDefault(checkboxParams, event) {
+    let newValue;
+
+    this.setState(prevState => {
+      if (checkboxParams.checked) {
```
omgaz: shouldn't need this at all if we can just toggle the existing value. Also, I think this creates a disparity, why do we need checkbox params AND prevState, surely prevState should be our source of truth - if it's not reliable as such there's probably a bigger issue somewhere.

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+        newValue = prevState.value.concat([checkboxParams.value]);
+      } else {
+        newValue = prevState.value.filter(currentValue => currentValue !== checkboxParams.value);
+      }
+      if (this.props.onChange) {
+        this.props.onChange(newValue, event, this.props.name);
```
omgaz: Can this onChange introduce side effects? What happens if this fails? I think, because this is passed from the user we shouldn't put potentially unsafe (or asyncronous stuff) in the state transformation but move it to the callback of setState.

@lightbringer1991 @xiaofan2406 thoughts?

src/components/adslot-ui/CheckboxSingle/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+    }));
+    if (this.onChange) {
+      this.onChange({ value: this.props.value, checked: isChecked }, event, this.props.name);
+    }
+  }
+
+  render() {
+    const { name, value, label, dts } = this.props;
+    return (
+      <div className="checkbox-single">
```
omgaz: Is this div necessary? Can we just not export the root as label?

src/components/adslot-ui/CheckboxGroup/index.jsx
nimishjha: But this is the checkboxGroup -- its state is ['value1','value2'...], not like a single checkbox.

src/components/adslot-ui/CheckboxGroup/index.jsx
nimishjha: True, removed checkboxParams and now only using event.

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { value: {} };
+    React.Children.map(this.props.children, child => {
+      this.state.value[child.props.value] = false;
+    });
+    if (this.props.value) {
+      this.state.value = mapPropsToState(this.props, this.state.value);
```
omgaz: We should probably call getDerivedStateFromProps?

nimishjha: Wouldn't it get called automatically when the props change? It's a lifecycle method, I'm using it in place of the deprecated `componentWillReceiveProps`.

lightbringer1991: `constructor()` is only called once when the component initializes, while `getDerivedStateFromProps()` is called every time props or state changes.

I think this block should be in `getDerivedStateFromProps()`, but I haven't tested it out yet. I assume that `getDerivedStateFromProps()` is called when component initializes.

nimishjha: Yes, it's called when the component is instantiated, as well as when props change. Why do you think that block of code needs to be in `getDerivedStateFromProps()`? Do we envision the child checkboxes changing during the life of a CheckboxGroup?

lightbringer1991: @nimishjha I think we need the state to be updated whenever props are changed, so it can be controlled from outside via props. With the current implementation, we can only pass in the props on initialization, and once it's done, there is no control over the values of the checkboxes inside `CheckboxGroup`.

I might be missing something here, this is just my thoughts when reading the PR. I may need to pull your branch to investigate further, just to play around with the scenarios.

nimishjha: That's exactly what's happening in `getDerivedStateFromProps()` here. Check the tests, I have a couple of them that test that scenario.

lightbringer1991: ahh yep, I missed the function before `constructor()`, so I guess we don't need this block here then, as it will be handled in `getDerivedStateFromProps()` anyway.

nimishjha: We do need the block here, because this is where the state object gets initialized - we loop over the children checkboxes, get each child's `value` and attach a property of that name to the state object. Then if props have a `value` array, we map that to the state object. `getDerivedStateFromProps` handles subsequent props changes.

src/components/adslot-ui/CheckboxSingle/index.jsx
nimishjha: One problem with removing the divs is that labels are inline by default, so the checkboxes/labels display in a row instead of in a column. Setting display: block on the labels solves that problem, but introduces a new one, namely that clicking on the whitespace beyond the text still toggles the checkbox, which doesn't seem desirable.

src/components/adslot-ui/Checkbox/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+        {label ? <span>{label}</span> : null}
+      </label>
+    );
+  }
+}
+
+Checkbox.propTypes = {
+  id: PropTypes.string,
+  className: PropTypes.string,
+  'data-name': PropTypes.string,
```
nimishjha: `data-name` is for backwards compatibility with Symphony.

src/components/adslot-ui/Checkbox/styles.scss
```diff
@@ -0,0 +1,14 @@
+.checkbox-single {
```
lightbringer1991: should it be just `.checkbox`?

src/components/adslot-ui/CheckboxGroup/index.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+CheckboxGroup.propTypes = {
+  name: PropTypes.string.isRequired,
+  value: PropTypes.arrayOf(PropTypes.string),
+  children: PropTypes.arrayOf(PropTypes.element).isRequired,
+  className: PropTypes.string,
+  onChange: PropTypes.func,
+};
+
+CheckboxGroup.defaultProps = {};
```
lightbringer1991: just nitpicking here, you can remove this line.

nimishjha: Nitpicks are important... fixed.

src/components/adslot-ui/Checkbox/styles.scss
```diff
@@ -0,0 +1,14 @@
+.checkbox {
```
omgaz: As a global CSS class I can see this is going to cause us issues when we come to put this into direct.  For example, this namespace is used by bootstrap and I believe it's used across our angular/non-angular/react code base poorly.

A reason why grid work as prefixes and why other components here suffix with -component. Not ideal, but I can see problems.

nimishjha:
Closing this PR as this page keeps crashing. Stay tuned for new PR.

PR #719, opened 2018-05-15
Fix IE breaking and Update react 

ChaoLiangSuper:
Fix IE breaking and Update react
## Description

- Update React to `16.3.2`
- Add polyfill to the entry in each `entry file`
- Add `isomorphic-fetch`, because IE 11 does not support ES6 `fetch`

- [ ] Yes
- [x] No

## Check-list:

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,6 +126,7 @@
   },
   "dependencies": {
     "diff-match-patch": "^1.0.0",
+    "isomorphic-fetch": "^2.2.1",
     "lodash": "^4.17.5",
     "react": "^16.3.2",
     "react-dom": "^16.3.2"
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "babel-plugin-transform-react-jsx": "^6.24.1",
     "babel-plugin-transform-strict-mode": "^6.24.1",
     "babel-polyfill": "^6.26.0",
-    "babel-preset-env": "^1.6.1",
+    "babel-preset-env": "^1.7.0",
     "babel-preset-react": "^6.24.1",
     "bootstrap-sass": "^3.3.7",
     "chai": "^4.1.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -126,9 +126,10 @@
   },
   "dependencies": {
     "diff-match-patch": "^1.0.0",
+    "isomorphic-fetch": "^2.2.1",
     "lodash": "^4.17.5",
-    "react": "^15.6.2 || ^16.3.1",
-    "react-dom": "^15.6.2 || ^16.3.1"
```
omgaz: We still need to support React 15 as a dependency here.

xiaofan2406:
This is my bad. Its caused by my PR for the babel change ages ago.

What caused it is that: The version of `babel-preset-env` we are using right now does not read `browserslist` from package.json. Only the next version which is for babel-7 and renamed to `@babel/preset-env` does it.

Essentially, the fix is: Copy the `browserslist` configuration from `package.json` back to `babelrc.js` file. See https://babeljs.io/docs/plugins/preset-env/#targetsbrowsers. Then babel should automatically provide poly-fill based on browser targets.

src/index.js
```diff
@@ -1,3 +1,5 @@
+// Polyfill for supporting IE 11
+import 'babel-polyfill';
```
xiaofan2406: I don't think its a library's responsibility to provider polyfills.

This file is huge and this would includes it in the .min.js file. Is that a desired outcome?

cc/ @omgaz

ChaoLiangSuper: Or alternatively, as talked with @xiaofan2406, maybe just polyfill the `object.assign`

omgaz: So - https://babeljs.io/docs/plugins/transform-object-assign/?

Or manually as a prototype on Object?

ChaoLiangSuper:
Update: 
- Add `babel-polyfill` to docs for IE 11 compatibility
- Not add polyfill to `dist`
- Keep the commit separately as they are isolated things

PR #723, opened 2018-05-29
Breaking: Radio and RadioGroup components

lightbringer1991:
- Creates a new `Radio` and `RadioGroup` components
- Consolidates prop types between `Radio` and `Checkbox`; `RadioGroup` and `CheckboxGroup`

TODO:

- [x] tests
- [x] IE11 compatibility check
- [ ] uninstall `react-icheck` (may need to go as a separate PR since it affects Checkbox and CheckboxGroup usage as well)

Resolves https://github.com/Adslot/adslot-ui/issues/700

- [x] Yes
- [ ] No

- Radio buttons API has been changed from `react-icheck` API to `adslot-ui` (see [documentation](https://adslot.github.io/adslot-ui/))
- `onClick` is no longer supported, `onChange` should be used instead
- `data-name` is no longer supported, please use `dts` instead

screen shot 2018-06-05 at 3 31 56 pm

## Check-list:

### DIFF
--- a/docs/examples/RadioExample.jsx
+++ b/docs/examples/RadioExample.jsx
@@ -1,47 +1,73 @@
 import React from 'react';
 import Example from '../components/Example';
-import { Radio, RadioGroup } from '../../src';
+import Radio from 'adslot-ui/Radio';
 
 class RadioExample extends React.PureComponent {
+  onChange(event) {
+    _.noop();
+  }
+
   render() {
     return (
-      <RadioGroup name="yesNo" className="radiogroup-stacked">
-        <Radio label="Yes" value="true" />
-        <Radio label="No" value="false" />
-      </RadioGroup>
+      <Radio
+        name="Radio button name"
+        label="Radio button label"
+        dts="radio-button-data-test-selector"
+        value="Radio button value"
+        onChange={this.onChange}
+      />
     );
   }
 }
 
 const exampleProps = {
   componentName: 'Radio',
-  designNotes: (
-    <p>
-      <span className="text-bold">Radio buttons</span> used for making a single selection from multiple options. Only
-      one selection can ever be made from the radio button group at a time.
-    </p>
-  ),
-  notes: (
-    <p>
-      See <a href="https://github.com/luqin/react-icheck">React iCheck Documentation</a>
-    </p>
-  ),
-  exampleCodeSnippet: `
-  <RadioGroup name="yesNo" className="radiogroup-stacked">
-    <Radio label="Yes" value="true" />
-    <Radio label="No" value="false" />
-  </RadioGroup>`,
+  notes: '',
+  exampleCodeSnippet: `<Radio
+  name="Radio button name"
+  label="Radio button label"
+  dts="radio-button-data-test-selector"
+  value="Radio button value"
+  onChange={this.onChange}
+/>`,
   propTypeSectionArray: [
     {
       propTypes: [
         {
+          propType: 'id',
+          type: 'string',
+        },
+        {
+          propType: 'className',
+          type: 'string',
+          note: 'This class will be applied to the input element',
+        },
+        {
+          propType: 'name',
+          type: 'string',
+        },
+        {
           propType: 'label',
-          type: 'node',
-          note: 'Usually fine to rely on a string but can pass HTML e.g. for a url.',
+          type: 'string',
         },
         {
           propType: 'value',
           type: 'string',
+        },
+        {
+          propType: 'dts',
+          type: 'string',
+          note: 'render `data-test-selector` onto the component. It can be useful for testing.',
+        },
+        {
+          propType: 'disabled',
+          type: 'bool',
+          defaultValue: <code>false</code>,
+        },
+        {
+          propType: 'checked',
+          type: 'bool',
+          defaultValue: <code>false</code>,
         },
         {
           propType: 'onChange',
--- a/src/components/adslot-ui/Radio/index.jsx
+++ b/src/components/adslot-ui/Radio/index.jsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { expandDts } from '../../../lib/utils';
+import { radioButtonPropTypes } from '../../prop-types/inputPropTypes';
+import './styles.scss';
+
+class RadioButton extends React.Component {
+  static getDerivedStateFromProps(newProps, prevState) {
+    return newProps.checked === prevState.checked ? null : { checked: newProps.checked };
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { checked: props.checked };
+
+    this.onChangeDefault = this.onChangeDefault.bind(this);
+  }
+
+  onChangeDefault(event) {
+    this.setState({ checked: Boolean(event.target.checked) });
+    if (this.props.onChange) {
+      this.props.onChange(event);
+    }
+  }
+
+  render() {
+    const { name, className, label, dts, disabled, id, value } = this.props;
+    const radioInputProps = {
+      type: 'radio',
+      name,
+      checked: this.state.checked,
+      disabled,
+      onChange: this.onChangeDefault,
+      value,
+      id,
+      className,
+    };
+
+    return (
+      <div className="radio-component" {...expandDts(dts)}>
+        <span className={`selection-component-icon iradio${this.state.checked ? ' checked' : ''}`} />
+        <label>
+          <div className="radio-component-input-container">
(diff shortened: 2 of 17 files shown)


docs/examples/RadioExample.jsx
```diff
@@ -1,52 +1,78 @@
...
     {
       propTypes: [
+        {
+          propType: 'id',
+          type: 'string',
+        },
+        {
+          propType: 'className',
+          type: 'string',
+          note: 'This class will go into the input element',
```
nimishjha: will be applied to

nimishjha:
Will the radio component ever be used by itself, i.e. not within a radioGroup? If not, maybe add a check for that and throw an error if someone tries to do so.

lightbringer1991:
we actually export both Radio and RadioGroup from `react-icheck` atm, I would prefer to do the same to avoid having it as a breaking change.

src/components/adslot-ui/Radio/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+    const radioInputProps = {
+      type: 'radio',
+      name,
+      checked: this.state.checked,
+      disabled,
+      onChange: this.onChangeDefault,
+      value,
+    };
+
+    const optionalProps = _.pickBy({
```
mdotwills: is it possible to grab these out without the `pickBy`. I was reading somewhere that it is expensive.

ChaoLiangSuper:
Had to pull your branch and ran it locally. 
I dont think the style is the same as compared to the previous. Can you check with UX team? Or @omgaz 

Used to be:
screen shot 2018-05-30 at 10 46 16 am

New style: 
screen shot 2018-05-30 at 10 46 21 am

lightbringer1991:
just investigated `symphony` code, this change and the Checkbox change will break `symphony` quite a bit, since the syntax of `onChange` function changes from `(event, value) => {...}` to `(event) => {...}`

omgaz:
Echo @ChaoLiangSuper's design comment. The padding and spacing is off and needs to be resolved. Make sure the label is 5px from the input and 20px (10px each) between them. This should also be the case for the checkboxes.

screen shot 2018-05-31 at 13 50 53

src/components/adslot-ui/Radio/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+  disabled: PropTypes.bool,
+  checked: PropTypes.bool,
+  onChange: PropTypes.func,
+  value: PropTypes.string,
+};
+
+RadioButton.defaultProps = {
+  disabled: false,
+  checked: false,
+};
```
omgaz: I'm wondering if you think it might be worthwhile to share the prop types between radios and checkboxes to ensure they have a shared API. I think it will help ensure consistency between the two components that make sense separated as components.

lightbringer1991: interesting, I'll investigate that possiblity.

src/components/adslot-ui/Radio/styles.scss
```diff
@@ -0,0 +1,14 @@
...
+    vertical-align: middle;
+    cursor: pointer;
+  }
+
+  .radio-label {
+    vertical-align: middle;
+    padding-left: 5px;
+  }
+
+  padding-top: 10px;
```
ChaoLiangSuper: Combine L12 and L13 together? `padding: 10px 0` ?

src/components/adslot-ui/Radio/index.jsx
lightbringer1991: I think we should still export them separately, as per icheck.
Combining checkbox and radio into 1 component is a bit over complicated imo.

omgaz: It makes sense to keep them separate, I think they get grouped together too often because they feel similar. I think sharing a prop type API is a good way of having consistency without coupling their logic and forcing branches in the code to differentiate logic.

src/components/adslot-ui/Radio/styles.scss
lightbringer1991: I think it's easier to read this way so I separate them out.

src/components/adslot-ui/Radio/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+      type: 'radio',
+      name,
+      checked: this.state.checked,
+      disabled,
+      onChange: this.onChangeDefault,
+      value,
+      id,
+      className,
+      'data-name': this.props['data-name'],
+    };
```
lteacher: Is this the new standard for setting props on the input? I usually like to just see them on the input so wondering why we need the extra var and then later spread?

src/components/adslot-ui/Radio/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+  constructor(props) {
+    super(props);
+    this.state = { checked: props.checked };
+
+    this.onChangeDefault = this.onChangeDefault.bind(this);
+  }
+
+  onChangeDefault(event) {
+    this.setState({ checked: Boolean(event.target.checked) });
+    if (this.props.onChange) {
```
lteacher: Why don't we just use `defaultProps` with `_.noop` to set these functions? then won't have to keep doing these conditional checks.

src/components/adslot-ui/RadioGroup/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+        },
+      });
+
+      return React.cloneElement(child, childProps);
+    });
+  }
+
+  render() {
+    const { dts, className, id } = this.props;
+    const componentProps = _.pickBy({
```
lteacher: Whats the `pickBy` for? isnt this just the object given since I dont see a predicate.

src/components/adslot-ui/Radio/index.jsx
nimishjha: One problem with using `defaultProps` like that here is that if you change the props during the lifetime of the component and don't specify `onChange` when you do so, it gets reset to to the `defaultProps` value, i.e. `_.noop`.

src/components/adslot-ui/Radio/index.jsx
lightbringer1991: this was done on `Checkbox` to accommodate symphony's usage of that particular `data-*` attribute. I think I need to discuss with Gary about this, and potentially remove `data-*` props and let symphony changes their code.

lightbringer1991: also it looks cleaner to me as well

src/components/adslot-ui/Radio/index.jsx
lteacher: Im not sure i understand how that works, if it changes with no `onChange` shouldn't it reset? what else should it be

nimishjha: Well, let's say you create a `radioButton` with an initial set of props, including an `onChange` function. Later on, you want to change the props, say, to set the radiobutton to `disabled` without changing anything else. I think it's better to be able to do that by just passing in a `disabled` prop without having to also pass in all the other props we passed in earlier.

lightbringer1991: most of the time we're actually just modify the `props` object and pass all of them back into the component, so this is generally not a problem. You can set it as `_.noop`.

Maybe it will help reducing the branches while testing as well.

lteacher: Im still struggling to get that implementation, is there somewhere to refer to that has an example? Sounds like some dynamic props definition which is passed to the component like:

```javascript
const initialProps = {
  checked: false,
  disabled: false,
  onChange: doAmazingStuff,
}

getRadioProps = (props = initialProps) => props;

// Somewhere out there in the wild
<Radio {...getRadioProps({ disabled: true })} />
```

Which doesn't sound like the right implementation to me?

lteacher: I think its the right move to pass all the props back from a consumer perspective. There might be some custom implementation such as defaults described, or `shouldComponentUpdate` check which depends on the props, or even the `getDerivedStateFromProps` so generally i would figure its not the best practice to do anything like my example unless sure its what is desired, i.e a totally new component with totally new props not just some kind of update.

nimishjha: Well, yeah, we could just pass in all the props every time, I really don't prefer one way over another.

ChaoLiangSuper:
Merging....🌚

Issue #727, opened 2018-06-06
Deprecate iCheck dependency


ChaoLiangSuper:
As the master branch has already replaced checkbox and radio components, will need to remove iCheck library.

## Types of issue

- [ ] Bug (High impact or critical to a production build)
- [ ] Bug (Low impact or not working as expected)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Something else (idea, build tool, documentation improvement)

ChaoLiangSuper:
Outdated

PR #728, opened 2018-06-06
Breaking: deprecate react-icheck

lightbringer1991:
- Remove `react-icheck` from dependency list.
- Remove `react-icheck` related css styles
- Introduce hover state for `Checkbox` component (since the svg is already there)
after

TODO:

- [ ] check with related projects

## Description

`react-icheck` is no longer needed.
The SVGs are actually ours, there is no style coming from `react-icheck`.

Resolves https://github.com/Adslot/adslot-ui/issues/727

- [x] Yes
- [ ] No

## Check-list:


ChaoLiangSuper:
Just realized one thing. Do we need to change the iCheck class name (the class name for the SVG) for some else class name in this PR?

omgaz:
screen shot 2018-06-07 at 09 55 30
screen shot 2018-06-07 at 09 55 40
screen shot 2018-06-07 at 09 56 04


Just some renames to clean-up and move the icheck custom stuff to the radio/checkbox scss stuff

lightbringer1991:
yea that would turn this PR into a breaking change. I'll do that then.

src/components/adslot-ui/Checkbox/index.jsx
```diff
@@ -37,11 +51,21 @@ class Checkbox extends React.Component {
...
       id,
       className,
+      onMouseEnter: this.onMouseEnter,
+      onMouseLeave: this.onMouseLeave,
     };
 
+    const iconClassNames = _([
+      'selection-component-icon',
+      this.state.checked ? 'checked' : null,
+      this.state.isHover ? 'hover' : null,
```
mdotwills: _suggestion:_ if we want to reduce our dependency on lodash we could

```javascript
const iconClassNames = ['selection-component-icon'];
if (this.state.checked) iconClassNames.push('checked');
if (this.state.isHover) iconClassNames.push('hover');
```

lightbringer1991:
@mdotwills this is already a breaking change PR, combining with the previous merged PR (which is also a breaking change), we're looking at releasing a major version of `adslot-ui`

src/components/adslot-ui/Checkbox/index.jsx
omgaz: We already have classNames library that's also part of React that does this:

```es6
import cx from 'classNames';

const iconClassNames = cx([
  'selection-component-icon',
  {
    'checked': this.state.checked,
    'hover': this.state.isHover,
  },
]
```

src/styles/mixins/enhancedSwitch.scss
```diff
@@ -0,0 +1,39 @@
...
+
+@mixin enhanced-switch-icon {
+  background-repeat: no-repeat;
+  border: 0;
+  cursor: pointer;
+  display: inline-block;
+  height: 16px;
+  position: relative;
+  vertical-align: middle;
+  width: 16px;
```
omgaz: Do you think it makes sense to use `$svg-icon-size` for width/height or do you feel we want to keep them separate?

omgaz: Can you also keep width & height together?

lightbringer1991: I didn't know we already have a variable, will switch over

lightbringer1991:
stale, closing

PR #753, opened 2018-09-03
Fix: list picker pure component after checkbox updated

ChaoLiangSuper:
## Description

After checkbox changed, the list picker pure component need to mange its own state.

Currenttly the component is broken.
Should be a patch fix, it is not changing the usage. 

- [ ] Yes
- [x] No

## Check-list:

### DIFF
--- a/docs/examples/ListPickerExample.jsx
+++ b/docs/examples/ListPickerExample.jsx
@@ -14,12 +14,30 @@ const teamMember3 = { givenName: 'Jack', id: 3, surname: 'White' };
 const listPickerItems = [teamMember1, teamMember2, teamMember3];
 const listPickerInitialSelection = [teamMember2];
 const labelFormatter = item => `${item.givenName} ${item.surname}`;
-const addonFormatter = () => <Checkbox />;
+const addonFormatter = () => <UncontrolledCheckbox />;
 const listPickerItemHeaders = {
   label: 'Team',
   toggle: 'Primary',
   addon: 'Secondary',
 };
+
+class UncontrolledCheckbox extends React.PureComponent {
+  constructor() {
+    super();
+    this.state = {
+      checked: false,
+    };
+    this.handleCheck = this.handleCheck.bind(this);
+  }
+
+  handleCheck() {
+    this.setState(prevState => ({ checked: !prevState.checked }));
+  }
+
+  render() {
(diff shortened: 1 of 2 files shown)


docs/examples/ListPickerExample.jsx
```diff
@@ -14,13 +14,31 @@ const teamMember3 = { givenName: 'Jack', id: 3, surname: 'White' };
...
 const labelFormatter = item => `${item.givenName} ${item.surname}`;
-const addonFormatter = () => <Checkbox />;
+const addonFormatter = () => <UnControlledCheckbox />;
 const listPickerItemHeaders = {
   label: 'Team',
   toggle: 'Primary',
   addon: 'Secondary',
 };
 
+class UnControlledCheckbox extends React.PureComponent {
```
lteacher: Uncontrolled is actually a single word so the casing here should be `UncontrolledCheckbox`

nimishjha: @lteacher Unless, of course, he means a checkbox controlled by the United Nations... :)

docs/examples/ListPickerExample.jsx
```diff
@@ -14,13 +14,31 @@ const teamMember3 = { givenName: 'Jack', id: 3, surname: 'White' };
...
+  constructor() {
+    super();
+    this.state = {
+      checked: false,
+    };
+    this.handleCheck = this.handleCheck.bind(this);
+  }
+
+  handleCheck() {
+    this.setState({ checked: !this.state.checked });
```
nimishjha: Shouldn't this be using the updater function pattern? I.e., `setState((prevState, props) => ...)` ?

ChaoLiangSuper: Because there isn't complex logic here. And I think I saw this expression in other places. So should be fine?

docs/examples/ListPickerExample.jsx
ChaoLiangSuper: `AUControlled` and `USControlled`

docs/examples/ListPickerExample.jsx
ChaoLiangSuper: Fixed

ChaoLiangSuper:
Tested in the main project

Issue #756, opened 2018-09-05
SvgSymbol component is not working under IE 11


ChaoLiangSuper:
MDN source: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use#Browser_compatibility

The `<use>` tag does not support external source. 

So in order to update the `SvgSymbol` the component should detect whether the browser accept use tag, if not using img tag.

## Types of issue

- [x] Bug (High impact or critical to a production build)
- [ ] Bug (Low impact or not working as expected)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Something else (idea, build tool, documentation improvement)

ChaoLiangSuper:
As discussed will add polyfill for this.

PR #757, opened 2018-09-09
Update SVG component to support based64

ChaoLiangSuper:
Resolves: #756 

- Update svg component to support base 64 encoded
- Add polyfill for svg component
- [x] Fix test
- [x] Update docs

- [ ] Yes
- [x] No

## Check-list:

### DIFF
--- a/src/components/alexandria/SvgSymbol/index.jsx
+++ b/src/components/alexandria/SvgSymbol/index.jsx
@@ -3,7 +3,10 @@ import PropTypes from 'prop-types';
 import { classSuffixHelper } from 'lib/utils';
 import './styles.scss';
 
+const base64UrlPrefix = 'data:image/svg+xml;base64,';
+
 const SvgSymbol = props => {
+  let isEncoded = false;
   const { classSuffixes, href, onClick } = props;
   const componentClass = 'svg-symbol-component';
   const suffixOptions = { clickable: props.onClick };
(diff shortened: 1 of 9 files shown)


src/components/alexandria/SvgSymbol/index.jsx
```diff
@@ -3,8 +3,15 @@ import PropTypes from 'prop-types';
 import { classSuffixHelper } from 'lib/utils';
 import './styles.scss';
 
+const componentName = 'AlexandriaSvgSymbolComponent';
+
 const SvgSymbol = props => {
-  const { classSuffixes, href, onClick } = props;
+  if (!props.href && !props.based64Data) {
+    console.error(`One of props 'href' or 'based64Data' must be specified in '${componentName}'.`);
```
nimishjha: I don't think we need to check for this and show an error. If they provide nothing, it renders nothing and it should be obvious what the problem is -- and we have example code for every component.

src/components/alexandria/SvgSymbol/index.jsx
```diff
@@ -13,18 +20,29 @@ const SvgSymbol = props => {
...
   );
 };
 
-SvgSymbol.displayName = 'AlexandriaSvgSymbolComponent';
+SvgSymbol.displayName = componentName;
 
 SvgSymbol.propTypes = {
   classSuffixes: PropTypes.arrayOf(PropTypes.string.isRequired),
   href: PropTypes.string,
+  based64Data: PropTypes.string,
```
nimishjha: Didn't we discuss using custom function `propTypes` to check for the situation where at least one of two props is required?

src/components/alexandria/SvgSymbol/index.jsx
```diff
@@ -3,7 +3,10 @@ import PropTypes from 'prop-types';
 import { classSuffixHelper } from 'lib/utils';
 import './styles.scss';
 
+const base64Reg = /data:image\/svg\+xml;base64,.+/g;
```
nimishjha: This regex is too lenient. Base64 data only contains @, =, and 0-9A-Za-z. I'm searching for a better regex, will let you know.

nimishjha: Use this instead: 

`^@[a-zA-Z0-9+/]+={0,2}$`

nimishjha: in place of the `.+`

ChaoLiangSuper: As discussed offline, will use `indexOf`

PR #778, opened 2018-11-29
fix: use own implementation of Tabs&Tab

xiaofan2406:
## Description

This PR re-implements the same API with the current `Tabs` & `Tab` component.
but missing animation related props.

react-bootstrap `Tabs` and `Tab` components are causing issue https://reactjs.org/docs/error-decoder.html/?invariant=254&args[]=inner

I do not know why. I have also seen some other people on react-bootstrap repo with the same error for Tabs

So my conclusion was: we gotta write our own Tabs.

- [ ] Yes
- [x] No

## Check-list:

### DIFF
--- a/src/components/adslot-ui/Tab/index.jsx
+++ b/src/components/adslot-ui/Tab/index.jsx
@@ -1,0 +1,21 @@
+/* eslint-disable react/no-unused-prop-types */
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+const Tab = ({ children, show }) => (
+  <div role="tabpanel" aria-hidden={show} className={classnames(['tab-pane', 'fade', { active: show, in: show }])}>
+    {children}
+  </div>
+);
+
+Tab.propTypes = {
+  children: PropTypes.node.isRequired,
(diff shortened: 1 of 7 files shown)


src/components/adslot-ui/Tab/index.jsx
```diff
@@ -0,0 +1,20 @@
...
+const Tab = ({ children, show }) => (
+  <div role="tabpanel" aria-hidden={show} className={classnames(['tab-pane', 'fade', { active: show, in: show }])}>
+    {children}
+  </div>
+);
+
+Tab.propTypes = {
+  children: PropTypes.node.isRequired,
+  eventKey: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
+  title: PropTypes.node.isRequired,
```
ChaoLiangSuper: Any reason for those `no-used` props? Like this `title`?

xiaofan2406: they are used by `Tabs`

xiaofan2406:
@lteacher docs updated with some notes and proptypes

PR #783, opened 2018-12-13
Update babel

ChaoLiangSuper:
- Update babel from 6 to 7

- [ ] Yes
- [x] No

## Check-list:


package.json
```diff
@@ -137,6 +138,7 @@
     "webpack-merge": "^4.1.4"
   },
   "dependencies": {
+    "@babel/polyfill": "^7.0.0",
```
lightbringer1991: shouldn't it be devDependencies?

ChaoLiangSuper: No the doc says need to be under `dependencies` 
https://babeljs.io/docs/en/babel-polyfill

src/components/adslot-ui/AlertInput/index.jsx
```diff
@@ -14,33 +14,25 @@ export default class AlertInput extends Component {
...
-    this.handleMouseEnter = this.handleMouseEnter.bind(this);
-    this.handleMouseLeave = this.handleMouseLeave.bind(this);
-    this.handleInputFocus = this.handleInputFocus.bind(this);
-    this.handleInputBlur = this.handleInputBlur.bind(this);
   }
 
-  getRef() {
-    return this.root;
-  }
+  getRef = () => this.root;
```
ChaoLiangSuper: This file is a example of using class properties, will change other files in later PRs.

package.json
```diff
@@ -89,15 +92,16 @@
     "eslint-config-prettier": "^2.9.0",
     "eslint-import-resolver-webpack": "^0.8.4",
     "eslint-loader": "^1.9.0",
+    "eslint-plugin-babel": "^5.3.0",
     "eslint-plugin-import": "^2.8.0",
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-lodash": "^2.6.1",
+    "eslint-plugin-prettier": "^3.0.0",
```
xiaofan2406: dont add this. we already have config pretteir extended.

.eslintrc
```diff
@@ -14,7 +15,7 @@
     "expect": false
   },
   "extends": ["airbnb", "prettier"],
-  "plugins": ["lodash", "react"],
+  "plugins": ["babel", "lodash", "react", "prettier"],
```
xiaofan2406: same here.
  this file change is not needed and L30 L31.

package.json
ChaoLiangSuper: OK sure

.eslintrc
ChaoLiangSuper: I think L30 is needed because it is in the `babel-plugin-eslint` pack

.babelrc.js
```diff
@@ -1,10 +1,15 @@
...
+const plugins = [
+  '@babel/plugin-transform-strict-mode',
+  '@babel/plugin-proposal-class-properties',
+  '@babel/plugin-proposal-object-rest-spread',
+  isDevelopment ? 'react-hot-loader/babel' : '',
+  isTest ? 'istanbul' : '',
+].filter(Boolean);
 
-const presets = process.env.TYPE === 'development' ? [['env', { modules: false }], 'react'] : ['env', 'react'];
+const presets = [['@babel/preset-env', { useBuiltIns: 'entry' }], '@babel/preset-react'];
```
xiaofan2406: you can actually copy set up here: https://github.com/Adslot/direct-web/pull/9238

the strict mode is not needed, hot loader is buggy, not sure about istanbul one tho.

config/polyfill.js
```diff
@@ -1,3 +1,5 @@
+require('@babel/polyfill');
```
xiaofan2406: use `useBuiltIns: usage` to remove this file completely. it only requires you to have this package installed

.eslintrc
xiaofan2406: babel-parser line and install that package will do.

config/polyfill.js
ChaoLiangSuper: `usage` is under experimental. So as a library, I think combining the polyfill in it is better.

.babelrc.js
ChaoLiangSuper: I will remove the strict mode 
I would say leave Hot-loader as it is, because it already using it. 
Then the `istanbul` is about the coverage.

.eslintrc
ChaoLiangSuper: OK sure

config/polyfill.js
xiaofan2406: polyfill should only be used for the app, not the library distribution?

and `usage` will include only minimal amount polyfill actually used rather than everything.

ChaoLiangSuper: But think this, as a library, we mark this lib support for IE 10, then if the use want to use it, it should be ready to go. Isn't it ?

PR #798, opened 2019-01-21
feat: refactor Accordion component

lightbringer1991:
- Refactor Accordion component to take in children as the Panels to be rendered.
- Allow uncontrolled behaviour to collapse/expand Panels.
- Resolves https://github.com/Adslot/adslot-ui/issues/799

## Description

Accordion component has been taking in Panel's props array to render children internally, which can cause some performance issue when dealing with a dynamic list of Panels. By allowing user to define the Panels themselves, we improve the performance on the Accordion, and enhance the readability.

```jsx
<Accordion defaultActivePanelIds={["filter-by-region"]} onPanelClick={this.onPanelClick} maxExpand={1}>
    <Accordion.Panel id="filter-by-region" title="Filter by region">
      <ul className="list-unstyled">
        <li>
          <Checkbox label="Australia" />
        </li>
        <li>
          <Checkbox label="New Zealand" />
        </li>
      </ul>
    </Accordion.Panel>
    <Accordion.Panel id="filter-by-device" title="Filter by device">
      <ul className="list-unstyled">
        <li>
          <Checkbox label="Desktop" />
        </li>
        <li>
          <Checkbox label="Mobile" />
        </li>
        <li>
          <Checkbox label="Tablet" />
        </li>
      </ul>
    </Accordion.Panel>
  </Accordion>
```

- [x] Yes
- [ ] No

The API for Accordion is changed, therefore we need to update all places that use the Accordion in the related projects.
Changing from using `props.panels` to actually render `Panel` component in `Accordion` children (as shown in the example)

## Check-list:

### DIFF
--- a/docs/components/Layout/index.jsx
+++ b/docs/components/Layout/index.jsx
@@ -39,6 +39,7 @@ import SvgSymbolExample from '../../examples/SvgSymbolExample';
 import SvgSymbolCircleExample from '../../examples/SvgSymbolCircleExample';
 import TileGridExample from '../../examples/TileGridExample';
 import AccordionExample from '../../examples/AccordionExample';
+import AccordionPanelExample from '../../examples/AccordionPanelExample';
 import CarouselExample from '../../examples/CarouselExample';
 import ConfirmModalExample from '../../examples/ConfirmModalExample';
 import HelpIconPopoverExample from '../../examples/HelpIconPopoverExample';
@@ -227,6 +228,7 @@ class PageLayout extends React.Component {
             <CardExample />
             <PanelExample />
             <AccordionExample />
+            <AccordionPanelExample />
             <BorderedWellExample />
             <CarouselExample />
             <GridExample />
--- a/src/components/adslot-ui/Accordion/index.jsx
+++ b/src/components/adslot-ui/Accordion/index.jsx
@@ -1,35 +1,91 @@
 import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
-import Panel from 'adslot-ui/Panel';
 import Card from 'alexandria/Card';
+import AccordionPanel from './AccordionPanel';
 
-const AccordionComponent = ({ dts, panels, onPanelClick }) => (
-  <Card.Container>
-    <Card.Content fill>
-      {_.map(panels, panel => {
-        const panelDts = dts ? `panel-${panel.id}` : undefined;
-        const panelProps = {
-          key: panel.id,
-          onClick: onPanelClick,
-          dts: panelDts,
-          ..._.omit(panel, ['content', 'onClick']),
-        };
+class AccordionComponent extends React.PureComponent {
+  static propTypes = {
+    dts: PropTypes.string,
+    onPanelClick: PropTypes.func,
+    children: PropTypes.node,
+    defaultActivePanelIds: PropTypes.arrayOf(PropTypes.string),
+    maxExpand: PropTypes.oneOfType([PropTypes.number, PropTypes.oneOf(['max'])]),
+  };
 
-        return <Panel {...panelProps}>{panel.content}</Panel>;
-      })}
-    </Card.Content>
-  </Card.Container>
-);
+  static defaultProps = {
+    maxExpand: 'max',
+  };
 
-const accordionPanelPropTypes = _.pick(Panel.propTypes, ['id', 'icon', 'title', 'isCollapsed']);
+  state = {
+    activePanelIds:
+      this.props.maxExpand === 'max'
+        ? this.props.defaultActivePanelIds
+        : _.slice(this.props.defaultActivePanelIds, 0, this.props.maxExpand),
+  };
 
-accordionPanelPropTypes.content = PropTypes.node;
+  onPanelClick = panelId => {
+    const { maxExpand } = this.props;
 
-AccordionComponent.propTypes = {
-  dts: PropTypes.string,
-  panels: PropTypes.arrayOf(PropTypes.shape(accordionPanelPropTypes)).isRequired,
-  onPanelClick: PropTypes.func.isRequired,
-};
+    if (_.includes(this.state.activePanelIds, panelId)) {
+      // remove panelId out of the active list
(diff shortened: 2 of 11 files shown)


src/components/adslot-ui/Accordion/index.jsx
```diff
@@ -1,35 +1,66 @@
...
+    } else {
+      this.setState({
+        activePanelIds: [...this.state.activePanelIds, panelId],
+      });
+    }
+
+    if (this.props.onPanelClick) {
+      this.props.onPanelClick(panelId);
+    }
+  };
```
lightbringer1991: allows uncontrolled collapse/expand in the Accordion

src/components/adslot-ui/Accordion/index.jsx
```diff
@@ -1,35 +1,67 @@
...
+  static propTypes = {
+    dts: PropTypes.string,
+    onPanelClick: PropTypes.func,
+    children: PropTypes.node,
+    defaultActivePanelIds: PropTypes.arrayOf(PropTypes.string),
+    isSingleExpand: PropTypes.bool,
+  };
+
+  static defaultProps = {
+    isSingleExpand: false,
```
lteacher: Dunno what other examples are like but id rather see something like `expandMode` or similar as `single|multi` its a little more useful if some magical new one is needed, whereas if that happens with `isSingleExpand` then its immediately breaking and needs to change. Sure probably there wont be another but thats the thought I get from seeing a specifically named boolean sitting here.

src/components/adslot-ui/Accordion/index.jsx
```diff
@@ -1,35 +1,67 @@
...
+    }
+  };
+
+  renderPanelFromChildren = child => {
+    const { id, isCollapsed } = child.props;
+
+    // respects child.props.isCollapsed for controlled behaviour
+    return React.cloneElement(child, {
+      ...child.props,
+      onClick: this.onPanelClick,
```
lteacher: What about panel `onClick` here, the user's `onClick` is overridden is that correct?

lightbringer1991: yes it's overridden in the current implementation as well.

src/components/adslot-ui/Accordion/index.jsx
lightbringer1991: I'll change it to `multi: boolean`

ChaoLiangSuper:
As discussed offline, maybe can add this https://github.com/Adslot/adslot-ui/issues/799 in this PR, as we are doing the refactor.

ChaoLiangSuper:
Did some local testing.
I think whether we need to restrict the children type to be a `Panel` or we need to render every children as a collapsable tab
I tried this code
```jsx
<Accordion defaultActivePanelIds={['filter-by-region']} onPanelClick={this.onPanelClick} isSingleExpand>
            <Panel id="filter-by-region" title="Filter by region">
              <ul className="list-unstyled">
                <li>
                  <Checkbox label="Australia" />
                </li>
                <li>
                  <Checkbox label="New Zealand" />
                </li>
              </ul>
            </Panel>
            <div>test</div>
            <div>test</div>
            <div>test</div>
            <div>test</div>
            <Panel id="filter-by-device" title="Filter by device">
              <ul className="list-unstyled">
                <li>
                  <Checkbox label="Desktop" />
                </li>
                <li>
                  <Checkbox label="Mobile" />
                </li>
                <li>
                  <Checkbox label="Tablet" />
                </li>
              </ul>
            </Panel>
          </Accordion>
```
It renders this 
screen shot 2019-01-22 at 11 30 10 am
Firstly, I don't think those texts should be rendered into the previous tab (Or seems like it's part of the previous tab)

Second, when click on the text (the text is `div` and the div accepted `onClick` function)
Then it runs into this
screen shot 2019-01-22 at 11 27 21 am

ChaoLiangSuper:
Some suggestions, maybe we can export `Accordion` and another component `Accordion.Tab`
The `Tab` doesn't do anything. Just render the `panel`, and the `panel` accepts the onClick and id.
Also the `According` component can detect if the children is not the type `Accordion.Tab`, it will not render this child or throw a warning. 
The reason for this is before we used to using an array of panels, as currently if we want to accept `children`, you cannot predict what is user's input. So maybe some restriction is needed.
@lightbringer1991

src/components/adslot-ui/Accordion/index.jsx
veegandhi: Wouldn't those behavior be on the panel itself rather then the accordion rendering the panel?

ChaoLiangSuper: Also as discussed, if we want to allow the `maximum number of expanded tab` we may don't need this props.
For example:  (Ignore the naming) 
```jsx
<Accordion
  maxExpandable={1} // Means single tab only, this could be a default option
  maxExpandable={3} // Means multiple tabs but with limitation 
  maxExpandable={'max'} // Means multiple without limitation
>
</Accordion>
```

lightbringer1991: updated to `maxExpand` to accept either a number or `max`

lightbringer1991: default is 'max'

docs/components/Layout/index.jsx
```diff
@@ -39,6 +39,7 @@ import SvgSymbolExample from '../../examples/SvgSymbolExample';
 import SvgSymbolCircleExample from '../../examples/SvgSymbolCircleExample';
 import TileGridExample from '../../examples/TileGridExample';
 import AccordionExample from '../../examples/AccordionExample';
+import AccordionPanelExample from '../../examples/AccordionPanelExample';
```
ChaoLiangSuper: This is internal use only. I think it is no need to have an example. @lightbringer1991

lightbringer1991: it's not really an example, more like a place for people to know what kind of component we expect in `Accordion` children.

lightbringer1991:
@ChaoLiangSuper tested against our main project, looking good. Will need buddy test tomorrow.

ChaoLiangSuper:
Tested LGTM

Issue #796, opened 2019-01-11
Partial selected checkbox component


lightbringer1991:
As discussed
Prefer this approach
```
<Checkbox
  name="Name"
  label="Label"
  value="value"
  onChange={onChangeFunc}
  checked={true|false|'partial'}
/>
```

Depends on the implementation of the checkbox
if goes for a state component the onChange function will be
``` onChange={(true|false) => onChangeFunc}```
If the checkbox will be a stateless component 
``` onChange={(true|false|'partial') => onChangeFunc}```

PR #802, opened 2019-01-25
Partial select checkbox

ChaoLiangSuper:
## Description
Fixes: https://github.com/Adslot/adslot-ui/issues/796
- Rewrite checkbox component
- Deprecate `iCheck` classes
- Added new svgs for checkbox icons
- Added partial state
- Change onChange API arguments (Because we used to get value, but never used the value)
   Change from `onChange(event, name)` to `onChange(nextState, name, value)`
- State cycle for the checkbox will be 
   true -> false,
   false -> true,
   'partial' -> false,
- Add `size` prop to set the size of the checkbox (default: 16)
- ~Reduce DOM elements for only using checkbox without label~ 
(Revert this change, cause if there is not `label` the input will not catch the `onChange` event)

- [x] Test and coverage
- [x] Update checkbox group 

- [x] Yes
- [ ] No

screen shot 2019-02-06 at 12 35 12 pm

## Check-list:


config/dev-cold.js
```diff
@@ -18,17 +18,6 @@ module.exports = merge(commonConfig, {
     filename: 'app.js',
     publicPath,
   },
-  module: {
```
ChaoLiangSuper: Using `url-loader` to handle images

src/components/adslot-ui/Checkbox/styles.scss
```diff
@@ -1,3 +1,84 @@
...
+  }
+
+  .checkbox-component-icon {
+    display: inline-block;
+    background-repeat: no-repeat;
+    background-size: contain;
+    background-image: url('~styles/icons/checkbox/empty.svg');
+  }
+
+  &:hover:not(.disabled):not(.checked):not(.partial-checked) .checkbox-component-icon {
```
nimishjha: If we're using SVGs, can the disabled states have a little more contrast? Right now I can barely see the difference between `checked disabled` and `partial disabled`.

ChaoLiangSuper: Those SVGs are from UX, so need to check with UX team.

ChaoLiangSuper:
@lightbringer1991 I don't think so, we have several PR which already marked `breaking` and I also checked with other people they are not block by this. So will do a 27 release right after this. So no need to have a feature branch.

ChaoLiangSuper:
Design reviewed, screenshot updated.

PR #806, opened 2019-01-31
Remove 'react-hot-loader'

ChaoLiangSuper:
Remove `react-hot-loader` library, and using `webpack-dev-server` to handle hot load.

`react-hot-loader` was modifying the babel compiler. 
Which give this `<Component />.type === Component` false


vinteo:
Merging...

Issue #814, opened 2019-02-21
A non-disappear popover


ChaoLiangSuper:
Per requirement: 
Need a popover which can be triggered to show by clicking on it, and when the mouse is away it will not disappear. So the popover will stay there, and if the popover has some link in it, the link should be clickable as well.

veegandhi:
Wondering if this whole thing can just be a Popover instead of thinking it as an controlled or uncontrolled? Supported props looks like
- Placement: left, right etc
- Theme
- Trigger: focus, click, alwaysShow ??

veegandhi:
```
<Popover popoverContent={<div>content</div>} trigger="hover" placement="foo" theme="dark" >
   <cell>
</Popover>
```

PR #815, opened 2019-02-22
feat: Introduce Popper.js and rewrite Popover component

lightbringer1991:
- Add `Popper.js`
- Rewrite `Popover`

Resolves https://github.com/Adslot/adslot-ui/issues/814

- [x] Yes
- [ ] No

popover-example

## Check-list:


package-lock.json
```diff
@@ -817,16 +817,14 @@
       "version": "7.2.0",
```
veegandhi: Walking through this lock file a lot  `https` has changed to `http`. You will need to be revert this.

src/components/third-party/Popover/Uncontrolled/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+import themes from '../constants';
+import './styles.scss';
+
+const basePlacements = ['auto', 'top', 'right', 'bottom', 'left'];
+const popoverPlacements = _.flatMap(basePlacements, trigger => [trigger, `${trigger}-start`, `${trigger}-end`]);
+
+class UncontrolledPopover extends React.PureComponent {
+  static propTypes = {
+    theme: PropTypes.oneOf(themes),
+    title: PropTypes.node.isRequired,
```
veegandhi: title can be optional

src/components/third-party/Popover/Uncontrolled/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+    return (
+      <Manager>
+        <Reference innerRef={this.referenceRef}>
+          {({ ref }) => (
+            <span
+              ref={ref}
+              onClick={this.onClick}
+              onMouseOver={this.onMouseOver}
+              onFocus={_.noop}
```
veegandhi: If these are required by Popper. We might as well support it. It's harmless, IMO

src/components/third-party/Popover/Uncontrolled/popoverTheme.scss
```diff
@@ -0,0 +1,41 @@
+@mixin theme-popover($theme-background-color, $theme-border-color) {
```
veegandhi: `popover-theme.scss`

src/components/third-party/Popover/Uncontrolled/index.jsx
lightbringer1991: it throws React warnings in general, so `mouseover` is defined, then the element should have `focus` event as well

src/components/third-party/Popover/styles.scss
```diff
@@ -0,0 +1,220 @@
...
+      height: $arrow-inner-size;
+      display: block;
+      position: relative;
+      border: 0 solid $color-background;
+    }
+
+    &[data-placement*='top'] {
+      margin-bottom: -$arrow-outer-size;
+
+      border-top-width: $arrow-outer-size;
```
veegandhi: I think you can convert the below 6 lines into a `mixin`

src/styles/variable.scss
```diff
@@ -127,6 +128,7 @@ $popover-border-color: $color-border-lighter;
 $popover-arrow-width: 5px;
 $popover-arrow-outer-color: $popover-border-color;
 $popover-title-bg: $color-background;
+$popover-zindex: 1060;
```
nimishjha: Looking through the codebase, the convention for naming z-index variables seems to be `$zindex-popover`.

src/styles/_bootstrap-custom.scss
```diff
@@ -40,6 +40,7 @@
 // @import '../../node_modules/bootstrap-sass/assets/stylesheets/bootstrap/thumbnails';
 @import '../../node_modules/bootstrap-sass/assets/stylesheets/bootstrap/alerts';
 @import '../../node_modules/bootstrap-sass/assets/stylesheets/bootstrap/progress-bars';
+@import '../../node_modules/bootstrap-sass/assets/stylesheets/bootstrap/progress-bars';
```
veegandhi: duplicate import?

src/styles/variable.scss
```diff
@@ -33,6 +33,7 @@ $font-family-sans-serif: Roboto, Helvetica, sans-serif;
 $font-family-serif: Georgia, serif;
 $font-family-monospace: Monaco, Consolas, 'Courier New', monospace;
 
+$font-family-base: $font-family-sans-serif;
```
veegandhi: Unused var? 😕

docs/examples/PopoverExample.jsx
```diff
@@ -1,39 +1,84 @@
...
       <>
-        <label>Placement (left, right, bottom, top)</label>
+        <label>Placement (left(-start, -end), top(-start, -end), bottom(-start, -end), right(-start, -end))</label>
         <div style={{ display: 'flex' }}>
-          <Popover id="popover-example-left" placement="left" title="Popover Title">
-            LEFT
+          <Popover placement="left" title="Popover Title" popoverContent="Popover Left">
+            <Button>Left</Button>
+          </Popover>
+          <div style={{ width: 12 }} />
```
veegandhi: This should be controlled via style sheet.

src/components/third-party/Popover/index.jsx
```diff
@@ -1,34 +1,107 @@
...
+import classnames from 'classnames';
 import React from 'react';
 import PropTypes from 'prop-types';
-import BootstrapPopover from 'react-bootstrap/lib/Popover';
+import { Manager, Reference, Popper } from 'react-popper';
+import themes from './constants';
+import './styles.scss';
 
-export const themes = ['light', 'dark', 'warn', 'error'];
+const basePlacements = ['auto', 'top', 'right', 'bottom', 'left'];
```
veegandhi: If we are using `Popover/constants.js`. This should be moved there too.

src/components/third-party/Popover/index.jsx
```diff
@@ -1,34 +1,107 @@
...
+    theme: 'light',
+    placement: 'auto',
+    trigger: 'hover',
+    isOpen: false,
+  };
 
-    const themeClass = _.includes(themes, theme) ? `popover-${theme}` : 'popover-light';
+  constructor(props) {
+    super(props);
+    this.referenceRef = React.createRef();
```
veegandhi: Optional change: Class variables should work. We don't need constructor.

src/components/third-party/Popover/index.jsx
```diff
@@ -1,34 +1,107 @@
...
+    this.popperRef = props.popperRef || React.createRef();
+  }
+
+  state = {
+    isPopoverOpen: this.props.isOpen,
+  };
+
+  static getDerivedStateFromProps(props, state) {
```
veegandhi: Looking at this, I am confused on what is the intent behind this? If the `trigger === disabled` then only respect `props.isOpen`? What happens to the changes for theme and placement when `trigger === disabled`?

lightbringer1991: this is only for state changes, other props, if passed in, will not change the internal state

veegandhi: Ohh sorry I was confused by the `defaultProps` 😖 .

src/components/third-party/Popover/styles.scss
lightbringer1991: not really, since each placement has different configuration, so each of those blocks are duplicated twice only

src/components/third-party/Popover/index.jsx
```diff
@@ -1,34 +1,119 @@
...
   static propTypes = {
-    id: PropTypes.string.isRequired,
     theme: PropTypes.oneOf(themes),
+    title: PropTypes.node,
     className: PropTypes.string,
+    wrapperClassName: PropTypes.string,
+    placement: PropTypes.oneOf(popoverPlacements),
+    popoverContent: PropTypes.node.isRequired,
+    children: PropTypes.node.isRequired,
+    trigger: PropTypes.oneOf(['click', 'hover', 'disabled']),
```
veegandhi: We may need multiple triggers. I suggest this should be an array. But I am not entirely sure if `Popper` supports it?

lightbringer1991: popper doesn't have any trigger, we define the triggers ourselves.
I assume that we're going to have triggers like `focus`?

PR #818, opened 2019-03-08
Update react-select

ChaoLiangSuper:
## Description

- Update `react-select` to version v2
- New default styles
- Update docs provided two common scenarios `single selection` and `multiple selections`
- Add tests for coverage

- [x] Yes
- [ ] No

Screen Shot 2019-03-14 at 3 11 59 pm

## Check-list:


src/components/adslot-ui/Select/index.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+const componentBaseClass = 'select-component';
+
+const DropdownIndicator = props => (
+  <components.DropdownIndicator {...props}>
+    <img src={caretIcon} />
+  </components.DropdownIndicator>
+);
+
+const ClearIndicator = props => <components.ClearIndicator {...props}>✕</components.ClearIndicator>;
```
lightbringer1991: this x is a special character right? I remember we had a discussion to have it just a lowercase x...

ChaoLiangSuper: Yea it is the special one

src/components/adslot-ui/Select/index.spec.jsx
```diff
@@ -0,0 +1,67 @@
...
+  { value: 'uk', label: 'United Kingdom' },
+];
+
+describe('SelectComponent', () => {
+  it('should has base class', () => {
+    const wrapper = mount(<SelectComponent />);
+    expect(wrapper.find('div.select-component')).to.have.length(1);
+  });
+
+  it('should has custom option background', () => {
```
veegandhi: should 'have'

src/components/adslot-ui/Select/index.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+// The propType for this component should be the same as 'React-Select` lib
+const SelectComponent = props => (
+  <Select
+    components={{ DropdownIndicator, ClearIndicator }}
+    className={componentBaseClass}
+    classNamePrefix={componentBaseClass}
```
veegandhi: wondering why is the `classNamePrefix` and `className` is the same?

src/components/adslot-ui/Select/styles.js
```diff
@@ -0,0 +1,63 @@
+const defaultStyle = {
```
veegandhi: Though I agree passing styles this way is a good practice. But we don't do this yet in the UI library. I think we should keep it consistent and handle styles via sass and only pass `classNames` here where required.

docs/examples/SelectExample.jsx
```diff
@@ -46,19 +62,36 @@ const exampleProps = {
...
+      { value: 'jp', label: 'Japan' },
       { value: 'uk', label: 'United Kingdom' },
     ]}
-    placeholder="Countries"
+    closeMenuOnSelect={false}
+    placeholder="Select multiple country"
     value={this.state.selected}
     onChange={this.onChangeHandler}
+    isClearable
+    isMulti
```
veegandhi: Can you add some minimal propType defnation? Atleast for the properties stated in the example.

docs/examples/SelectExample.jsx
```diff
@@ -46,19 +62,36 @@ const exampleProps = {
...
+      { value: 'jp', label: 'Japan' },
+      { value: 'uk', label: 'United Kingdom' },
+    ]}
+    isOptionDisabled={({ value }) => value === 'jp'}
+    placeholder="Select single country"
+    value={this.state.selected}
+    onChange={this.onChangeHandler}
+    isClearable
+  />
+
```
veegandhi: Probably a header here, stating the below is code for multi select

src/components/adslot-ui/Select/styles.js
ChaoLiangSuper: The reason is this lib is already using the StyledComponent. I just consume their API to avoid css override. 

Also as you see the classnames above, those are provided for further css override in de main project if custom style required.

Overall, we won't introduce styled component into this lib.

src/components/adslot-ui/Select/index.jsx
ChaoLiangSuper: `className ` for wrapper.
`classNamePrefix` for all child node inside `react-select`

docs/examples/SelectExample.jsx
ChaoLiangSuper: It is specified inside placeholder

docs/examples/SelectExample.jsx
ChaoLiangSuper: We have the link just below this example pointing to the `react-select` page.
I think it is enough, and the prop name is understandable.

docs/examples/SelectExample.jsx
veegandhi: Ofcourse its specified in the palceholder, if we were to read the placeholders, we might as well remove all the headings from this doc. Strongly urge adding a header here.

ChaoLiangSuper: Sure

docs/examples/SelectExample.jsx
veegandhi: This example is specified in Adslot-ui doc and not in react-select. You can either remove the PropTypes block from below or add the proptype def.

ChaoLiangSuper: Because it is a third party lib. We added the link there from the beginning. 
In another word, the remove proptypes block is coming from the structure. 
if you want remove it, sure, but should be inside this update PR.

lteacher: Yeah since the props are passed straight through it doesn't seem worthwhile to redefine them here. But I did notice other examples as well that have the empty proptypes block, like the `DatePicker` for example.

Personally I would be ok with not removing for this PR else should be consistent and remove all empty blocks.

PR #824, opened 2019-03-22
[experimental] feat: transpiling

knilink:
## Description
Provide a transpiled version of the source code, like most of the popular component libraries do. 
`npm run build:es` to run the traspiling

Currently the library contain only bundle minified, which make debugging difficult.  
Developers can also apply and test the fix in the transpiled files if they found any issues.

The bundle minified file is the same as before, but importing the transpiled version may break some unit test because the component name is restored.

- [x] Yes
- [ ] No

## Check-list:


src/style.js
```diff
@@ -0,0 +1,77 @@
+/* eslint-disable id-length */
+/* eslint-disable no-console */
+
+import 'styles/_bootstrap-custom.scss';
```
ChaoLiangSuper: Any reason for removing styles outside each project?

src/style.js
```diff
@@ -0,0 +1,77 @@
...
+import 'styles/_icheck-custom.scss';
+import 'styles/_react-datepicker-custom.scss';
+import 'styles/_react-select-custom.scss';
+
+const req = require.context('./components', true, /^\.\/.*\/styles\.scss$/);
```
veegandhi: This is definitely good change 💥

src/style.js
```diff
@@ -0,0 +1,77 @@
...
+
+function camelCase(name) {
```
veegandhi: loadash has a utility for this

src/style.js
knilink: have separate the module code and from the style.
otherwise it will break if user didn't setup plugin for scss, it will break when hiting  `import '*.scss'; `

src/component.js
```diff
@@ -0,0 +1,140 @@
+/* eslint-disable no-console */
```
veegandhi: `components.js`

src/style.js
```diff
@@ -0,0 +1,77 @@
...
+  return name.charAt(0).toUpperCase() + name.slice(1).replace(/-(\w)/g, (m, n) => n.toUpperCase());
+}
+req.keys().forEach(mod => {
+  let v = req(mod);
```
veegandhi: definitely needs a better variable name 😉

src/style.js
knilink: I copied this piece of code from somewhere else, I don't even know what `exports[camelCase(match[1])] = v;` do yet. So, not sure whether it's necessary. :trollface:

.babelrc.js
```diff
@@ -5,12 +5,11 @@ const plugins = [
   '@babel/plugin-proposal-class-properties',
   '@babel/plugin-proposal-object-rest-spread',
   isTest ? 'istanbul' : '',
+  ['babel-plugin-webpack-alias', { config: './webpack.config.js' }],
```
veegandhi: is this alias used anywhere?

src/component.js
knilink: Probably better to have a `index.js` in components folder.
I tried to achieve this PR as quick as I can so I didn't consider too much.
Probably the file name and location of `style.js` need to be reconsider as well.

.babelrc.js
knilink: https://github.com/Adslot/adslot-ui/blob/21454650ab7d3014abace786c6c8b5fffba6f151/config/common.js#L11-L17
https://github.com/Adslot/adslot-ui/blob/21454650ab7d3014abace786c6c8b5fffba6f151/src/index.js#L78
need to turn the alias to the actual path, otherwise it will be resolved as a module.

ChaoLiangSuper:
Just wondering what's the final `dist/` structure after this change.

knilink:
@ChaoLiangSuper The `dist` fold remain unchanged.

src/style.js
knilink: unnecessary code, removed

src/component.js
knilink: the `component.js` was moved to `src/components/index.js`
the `style.js` was merged to `src/index.js`

src/style.js
knilink: unnecessary code, removed.

package-lock.json
```diff
@@ -13374,21 +13429,33 @@
       }
     },
     "react": {
-      "version": "16.6.3",
-      "resolved": "https://registry.npmjs.org/react/-/react-16.6.3.tgz",
-      "integrity": "sha512-zCvmH2vbEolgKxtqXL2wmGCUxUyNheYn/C+PD1YAjfxHC54+MhdruyhO7QieQrYsYeTxrn93PM2y0jRH1zEExw==",
+      "version": "16.8.4",
```
ChaoLiangSuper: React is still get updated here.

knilink: removed nodule_modules and ran a clean installed.
This time should fix.

lightbringer1991:
hmm I'm not sure if this is necessary, since `adslot-ui` exposes development version and production version in the build.
Does this mean all dependant third party libraries are also shipped with development mode?

src/index.js
```diff
@@ -1,144 +1,8 @@
...
-  TextEllipsis,
-  TreePickerGrid,
-  TreePickerNav,
-  TreePickerNode,
-  TreePickerSimplePure,
-  UserListPicker,
-  InformationBox,
-  HoverDropdownMenu,
-} from 'adslot-ui';
+require.context('./components', true, /^.*\.scss$/);
```
xiaofan2406: if we are able to 
- leave out all the `import ***.scss` changes
- let the transpiler ignore the `scss` files

will be lot better.

Also this PR should deprecate the dev bundle js that exposes.

knilink:
@lightbringer1991 
It just a better practice to me. 

`adslot-ui` wouldn't ship any 3rd party npm modules in this scenario. All dependent 3rd party libraries will be installed by users via `npm` base on the `dependencies` in `package.json`. So if users run their project in development environment, all the 3rd party libraries will be running in development environment as well.

knilink:
outdated

PR #825, opened 2019-03-25
feat: new button group component

veegandhi:
## Description

Introduce a new button group

Provides a button group for similar styled buttons but do not have same events per say. They may represent a group of actions for a similar task.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Needs local testing

image

## Check-list:


src/components/adslot-ui/ButtonGroup/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+  injectProps(children) {
+    return React.Children.map(children, child => {
+      if (React.isValidElement(child)) {
+        const buttonProps = {
+          ...(this.props.bsStyle ? { bsStyle: this.props.bsStyle } : {}),
+          ...(this.props.inverse ? { inverse: this.props.inverse } : {}),
+          ...(this.props.disabled ? { disabled: this.props.disabled } : {}),
+        };
+
+        const childNodes = this.injectProps(child.props.children);
```
veegandhi: Inject props into children recursively

docs/examples/styles.scss
```diff
@@ -82,6 +82,28 @@
       width: 200px;
     }
   }
+
+  &.button-group-example {
+    .adslot-ui-example {
+      display: flex;
+      .aui__button-group {
+        .btn {
+          border-color: $color-white;
```
ChaoLiangSuper: Feel this border color should be part of the `group component`

veegandhi: IMO, I think as a button group we should not update the base style of the button. However, having said that, when we update with your new button set, we will look into it so that there is a consistency in border moving forward.

ChaoLiangSuper: Screen Shot 2019-03-26 at 9 49 59 am

Button group should have at least some styles for example the "|" separator between each buttons. This style need to be part of the group component as default. Otherwise, by default the button group will show as above. We cannot expect the user to overwrite styles every single time. 

Also margin should go into the group component as well. For the situation, when `group button` + `group button` and `group button` + `button`

veegandhi: What you are mentioning in the above comment is a very specific use case. As a UI we should support generic stuff then individual use case. All the cases you have mentioned above also have a flip side. For eg. What if they are expecting the button without the border override (original button style we have)? What if the button group are placed in a flex div and spaced evenly? I personally don't agree on providing the exact use case because that will change in no time. However a couple of more opinions can help us take a decision on this.

CC: @nimishjha , @lteacher - Need you opinion on this case. Cheers!

nimishjha: I'm not across the spec for this component, but I think it should look like the screenshot in the PR description by default, without the user having to apply any styles.

lteacher: Hmm isnt this thing like locking the group to have some white border somewhere? havent pulled the pr to check it out

veegandhi: The white border is for the primary button, and for the inverse its grey border. Having said that, its not the default behavior of the button we ship. I think that is what we are trying to make a decision on, whether we change the default behavior of `Button` inside the button group?

lteacher: Ok sorry, the example only has the border. Yeah I guess i might have thought someone could change the group border and background or something to achieve it with some kind of spacing on the buttons?

I wouldnt think that there would be some automatic borders though like the example (which matches the PR descr) because what if the background isnt white and they dont want white borders, i mean the white is a pretty hard color.

nimishjha: We're already modifying certain things like border-radius on the child buttons, so I think it's fine to also add the separator and margins as a default. 

It also seems to me that `ButtonGroup` is a misleading name for this component and is causing some of the confusion. It should be 'DropdownButton' or something. A 'button group' is just a group of vanilla buttons; this is a very specific component with a button that has a dropdown (or popover, whatever).

lteacher: Well the radius does seem to make some sense it rounds the group of buttons into like a bar of buttons.

The border thing, I think the spacer makes the most sense but I suppose that it could also make sense to change the border but I would think that it would be something set on the group (which may flow to the buttons)

The dropdown thing, i mean thats just the example given here there is no need for it to be a dropdown and rather it could just be a group of buttons.

docs/examples/ButtonGroupExample.jsx
```diff
@@ -0,0 +1,137 @@
...
+import { Button, ButtonGroup, OverlayTrigger, Popover, SvgSymbol } from '../../src';
+
+class ButtonGroupExample extends React.PureComponent {
+  render() {
+    return (
+      <React.Fragment>
+        <ButtonGroup bsStyle="success">
+          <Button>Approve</Button>
+          <OverlayTrigger
+            trigger={['click']}
```
ChaoLiangSuper: I think we said not using the overlayTrigger, instead will add an alert or something to indicate the user has clicked on the button. 
Also We said to add some examples for buttongroup like `leftbutton, centerbutton, rightbutton` ?

src/components/adslot-ui/ButtonGroup/styles.scss
```diff
@@ -0,0 +1,38 @@
...
+
+    &:last-child,
+    &:last-child:hover {
+      border-right-width: 1px;
+      border-top-right-radius: 2px;
+      border-bottom-right-radius: 2px;
+    }
+
+    + .button-component,
+    + button {
```
ChaoLiangSuper: Just note here.
3 situations should be 
`Group + Group`, `Group + Button` and `Button + Group`

src/components/adslot-ui/ButtonGroup/index.jsx
```diff
@@ -0,0 +1,52 @@
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import './styles.scss';
+import { Button } from '../../third-party';
+
+class ButtonGroup extends React.PureComponent {
```
xiaofan2406: ui library should not contain any `PureComponent` unless every prop is of primitive type.

src/components/adslot-ui/ButtonGroup/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  injectProps(children) {
+    return React.Children.map(children, child => {
+      if (React.isValidElement(child)) {
+        const buttonProps = {
+          ...(this.props.bsStyle ? { bsStyle: this.props.bsStyle } : {}),
+          ...(this.props.inverse ? { inverse: this.props.inverse } : {}),
+          ...(this.props.disabled ? { disabled: this.props.disabled } : {}),
+          ...(this.props.bsSize ? { bsSize: this.props.bsSize } : {}),
```
xiaofan2406: inverse, and disabled are boolean.

imo, just spread them regardless and make sure the default values are correct.

src/components/adslot-ui/ButtonGroup/index.jsx
xiaofan2406: don't believe this is a good thing to do.

I can understand 1 level extra nesting due to potential React.Fragment usage, but this is a recursion without termination.

ChaoLiangSuper:
There is some movement around the `hover`, so please solve that before merging.

src/components/adslot-ui/ButtonGroup/index.jsx
veegandhi: changed to `isNil`

src/components/adslot-ui/ButtonGroup/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+          ...(!_.isNil(this.props.inverse) ? { inverse: this.props.inverse } : {}),
+          ...(!_.isNil(this.props.disabled) ? { disabled: this.props.disabled } : {}),
+          ...(this.props.bsSize ? { bsSize: this.props.bsSize } : {}),
+        };
+
+        const childNodes = React.Children.map(child.props.children, childNode =>
+          React.isValidElement(childNode)
+            ? React.cloneElement(childNode, {
+                ...childNode.props,
+                ...(childNode.type.name === Button.name ? buttonProps : {}),
```
xiaofan2406: should just use `childNode.type === Button` rather than a string comparison.

veegandhi: Results the same 😃 Prefer it this way!

src/components/adslot-ui/ButtonGroup/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+      return child;
+    });
+  }
+
+  render() {
+    const { children, dts } = this.props;
+    const content = this.injectProps(children);
+
+    return (
+      <div dts={dts} className="aui--button-group">
```
ChaoLiangSuper: should not be `dts` on the `div`, using the `expandDts` from the lib

ChaoLiangSuper:
LGTM Merging....

PR #829, opened 2019-03-27
New: info icon popover

ChaoLiangSuper:
## Description

Add info icon popover component
`HelpIconPopover` and `InfoIconPopover` will share most of their code.
Add new icons

- [ ] Test and coverage

- [ ] Yes
- [x] No

Screen Shot 2019-03-27 at 5 01 45 pm

## Check-list:


lightbringer1991:
what's the difference between `HelpIconPopover` and `InfoPopover`?

ChaoLiangSuper:
@veegandhi It is a feature work request.

ChaoLiangSuper:
@lightbringer1991 the only difference is the classname which contains the icon.
And the `InfoPopover` is not exported.

lightbringer1991:
I think if the difference is just the class name, ideally we can expose a component called `IconPopover`, and pass in the `className` as one of the props...
I think in this PR you've already done that, but not exposing `IconPopover` component.
@veegandhi @ChaoLiangSuper any chance that we can do that instead of exposing 2 similar components?

veegandhi:
@lightbringer1991 - I agree that should be the way to go and that is the reason this should have been in issue list to clarify on the implementation details. If we go by this approach, for every icon we have a popover, we will need another component all together.

@ChaoLiangSuper - Can you please add this item on the issue list regardless. Cheers!

ChaoLiangSuper:
@lightbringer1991 @veegandhi The reason for exposing two components. Because the these two components are for certain case only. And I try to not make a breaking change to the PR.
Also I don't mind to pass classnames. But for a normal case, we should not encourage people to pass a certain classname for changing the background.
As a normal case, people should use <Popover /> + <SvgComponent />

PR #842, opened 2019-05-02
chore: fix package-lock and add git diff check

bchew:
## Description
- fix package-lock
- switch to use npm install instead of npm ci so that we can validate changes
- add git diff check to fail CI to catch incorrect package-lock changes

Although `npm ci` is faster, changing travis-ci to use `npm install` instead so that we can catch incorrectly pushed package-lock changes.

- [ ] Yes
- [x] No

## Check-list:


lightbringer1991:
merging...

PR #850, opened 2019-05-22
Merge Search and SearchBar

### LINKED ISSUE #822: Merge Search and Search Bar
## Types of issue

- [ ] New feature (non-breaking change which adds functionality)
- [x] Something else (idea, build tool, documentation improvement)

## Description
It looks like `Search` and `SearchBar` has a lot of similarities and `SearchBar` is more enhanced version of `Search`. 

This could be simplified to single component. From recent usage it seems that we need to revisit the logic in `Search` component this could be a good opportunity to execute this cleanup task.

superpower1:
## Description

This PR has a new Search component that can cover both use cases of old Search and SearchBar component.

Fixed #822

- [x] Yes
- [ ] No

## How Has This Been Tested?

Manually tested it myself, need to do some buddy testing.

new_search

## Check-list:


docs/components/Layout/index.jsx
```diff
@@ -227,8 +226,7 @@ class PageLayout extends React.Component {
             <ConfirmModalExample />
 
             <PageTitle title="Search" />
-            <SearchExample />
-            <SearchBarExample />
+            <NewSearchExample />
```
vinteo: I think use replace it with `SearchExample`

docs/examples/NewSearchExample.jsx
```diff
@@ -0,0 +1,163 @@
+import React from 'react';
+import Example from '../components/Example';
+import { Search } from '../../src';
+
+class NewSearchExample extends React.PureComponent {
```
lightbringer1991: I think just `SearchExample` will suffice.

docs/examples/NewSearchExample.jsx
```diff
@@ -0,0 +1,163 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: '',
+    };
+  }
```
lightbringer1991: can also use 
```jsx
state = {
  value: '',
};
```
So we don't have to create a constructor. See https://github.com/Adslot/adslot-ui/blob/master/docs/examples/CheckboxGroupExample.jsx#L7 for reference

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,131 @@
...
-    const value = _.get(event, 'target.value');
-    onChange(value);
-    if (searchOnChange) this.debounceOnSearch(value);
-  }
+import './styles.scss';
 
-  onKeyPress(event) {
-    const { disabled, searchOnEnterKey, onSearch } = this.props;
-    if (disabled) return;
+function Search({
```
lightbringer1991: I think we prefer ES6 syntax
```jsx
const Search = ({ ... }) => {...};
```

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,131 @@
...
+  onSearch: PropTypes.func.isRequired,
   placeholder: PropTypes.string,
-  svgSymbolCancel: PropTypes.shape(SvgSymbol.propTypes),
-  svgSymbolSearch: PropTypes.shape(SvgSymbol.propTypes),
-  value: PropTypes.string,
   searchOnChange: PropTypes.bool,
-  searchOnEnterKey: PropTypes.bool,
-  debounceInterval: PropTypes.number,
+  showSearchBtn: PropTypes.bool,
+  value: PropTypes.string,
```
lightbringer1991: I think this list of props is longer than the proposed one here https://github.com/Adslot/adslot-ui/issues/822#issuecomment-492040864. Not sure if @veegandhi  has already discussed the matter.

src/components/adslot-ui/Search/index.jsx
lteacher: I dont think we have `function` anywhere so probably just use arrow syntax _(unless its needed for some reason but prob it isnt in this case)_

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,131 @@
...
-    this.onChange = this.onChange.bind(this);
-    this.onClear = this.onClear.bind(this);
-    this.onKeyPress = this.onKeyPress.bind(this);
-  }
-
-  onChange(event) {
-    const { disabled, searchOnChange, onChange } = this.props;
-    if (disabled) return;
+import classnames from 'classnames';
+import Button from 'react-bootstrap/lib/Button';
```
lightbringer1991: I think we should use `src/components/third-party/Button` instead.

src/components/adslot-ui/Search/index.spec.jsx
```diff
@@ -1,166 +1,173 @@
...
-  let props = {};
+  const props = {
+    className: 'additional-class',
+    dts: 'test-dts',
+    placeholder: 'search',
+    value: 'abc',
+  };
+
+  it('should render with defaults', () => {
+    const component = shallow(<Search />);
```
lightbringer1991: I think we follow the naming `const wrapper = shallow(....)`, still need to change into the new naming convention in quite a few places though 🛠

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,131 @@
...
+  dts,
+  icons,
+  isLoading,
+  onChange,
+  onClear,
+  onSearch,
+  placeholder,
+  searchOnChange,
+  showSearchBtn,
+  value,
```
lteacher: I think it was supposed to be possible to use as uncontrolled input and respond only to the `onSearch` since otherwise we were seeing a lot of pointless state wrapping just to manage the search input.

src/components/adslot-ui/TreePicker/Nav/index.spec.jsx
```diff
@@ -55,6 +57,21 @@ describe('TreePickerNavComponent', () => {
...
+    };
+    const svgSymbolSearch = {
+      href: '/assets/svg-symbols.svg#search',
+    };
+    const component = shallow(
+      <TreePickerNavComponent {...props} svgSymbolSearch={svgSymbolSearch} svgSymbolCancel={svgSymbolCancel} />
+    );
+    const searchElement = component.find(Search);
+    expect(_.has(searchElement.prop('icons'), 'search')).to.equal(true);
+    expect(_.has(searchElement.prop('icons'), 'close')).to.equal(true);
```
lightbringer1991: I think to simplify we can use
```jsx
expect(searchElement.prop('icons')).to.include.members(['search', 'close']);
```

src/components/adslot-ui/Search/index.jsx
superpower1: I was just following the coding style from airbnb `And if you don’t have state or refs, prefer normal functions (not arrow functions) over classes:

// bad
class Listing extends React.Component {
  render() {
    return <div>{this.props.hello}</div>;
  }
}

// bad (relying on function name inference is discouraged)
const Listing = ({ hello }) => (
  <div>{hello}</div>
);

// good
function Listing({ hello }) {
  return <div>{hello}</div>;
}`
Seems that arrow function is not a preference.
As I am going to use an internal 'value' state, I will change it to class.

src/components/adslot-ui/TreePicker/Nav/index.spec.jsx
superpower1: As icons is an object, it should be
`expect(searchElement.prop('icons')).to.have.keys(['search', 'close']);`

src/components/adslot-ui/Search/index.jsx
superpower1: As per discussion, I will remove showSearchBtn prop and replace searchOnChange with searchOnEnter. Search button will be shown if searchOnEnter is true.

docs/examples/SearchExample.jsx
```diff
@@ -2,154 +2,141 @@ import React from 'react';
...
-    this.onChange = this.onChange.bind(this);
-    this.onSearch = this.onSearch.bind(this);
-  }
+class SearchExample extends React.PureComponent {
+  state = {
+    value: '',
+  };
 
-  onChange(value) {
+  onChange = value => {
```
adityasabnis: `onChange = (value) => {`

I can't recollect of places we skip ()

docs/examples/SearchExample.jsx
```diff
@@ -2,154 +2,141 @@ import React from 'react';
...
-        isLoading={this.state.isLoading}
-      />
+      <>
+        Search without search button
+        <Search
+          placeholder="Search"
+          value={this.state.value}
+          onChange={this.onChange}
+          onSearch={this.onSearch}
+          debounceInterval={500}
```
adityasabnis: is there a reason to increase the debounceInterval?

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,148 @@
...
 };
 
 Search.defaultProps = {
+  debounceInterval: 0,
   disabled: false,
   isLoading: false,
-  onChange: _.noop,
-  onClear: _.noop,
-  onSearch: _.noop,
+  searchOnChange: false,
```
adityasabnis: can't locate defaultValue for `showSearchBtn`

docs/examples/SearchExample.jsx
lightbringer1991: I think that's the prettier doing the work

src/components/adslot-ui/Search/index.jsx
superpower1: I will remove this prop as we will use searchOnEnter to determine whether we show the button or not.

docs/examples/SearchExample.jsx
superpower1: Yeah, it's prettier

src/components/adslot-ui/TreePicker/Nav/index.jsx
```diff
@@ -21,24 +22,31 @@ const TreePickerNavComponent = ({
...
+  return (
+    <div className={`treepickernav-component ${disabled ? 'disabled' : ''}`} data-test-selector="treepicker-nav-search">
+      <Search
+        disabled={disabled}
+        debounceInterval={debounceInterval}
+        isLoading={isLoading}
+        onClear={onClear}
+        onChange={onChange}
+        onSearch={onSearch}
+        searchOnChange={searchOnChange}
```
veegandhi: `searchOnChange` and `searchOnEnter` are not used any more. They should be removed.

src/components/adslot-ui/TreePicker/Nav/index.jsx
```diff
@@ -21,24 +22,31 @@ const TreePickerNavComponent = ({
...
-      searchOnChange={searchOnChange}
-      searchOnEnterKey={searchOnEnterKey}
-      svgSymbolCancel={svgSymbolCancel}
-      svgSymbolSearch={svgSymbolSearch}
-      value={searchValue}
-    />
-    <Breadcrumb disabled={disabled} nodes={breadcrumbNodes} onClick={breadcrumbOnClick} />
-  </div>
-);
+}) => {
```
veegandhi: I feel this `TreePickerNavComponent` should be refactored with this too, because its plainly passing props to the `Search` which the new component does not accept anymore, so should be removed from NavBar component too.

src/components/adslot-ui/Search/styles.scss
```diff
@@ -1,31 +1,19 @@
-@import '~styles/border';
-@import '~styles/color';
-@import '~styles/font-size';
-@import '~styles/font-weight';
+@import '~styles/variable';
 
-.search-component {
+.aui--new-search {
```
veegandhi: .aui--search-component

src/components/adslot-ui/Search/styles.scss
```diff
@@ -38,25 +26,32 @@
...
-
-    &:disabled {
-      border: $border-lighter;
-      color: $color-text-disabled;
-    }
   }
 
-  .svg-symbol-component {
-    box-sizing: content-box;
+  &-icon {
```
veegandhi: This works but because when we search the classes it is hard to find class with `&-` as a team, a general agreement was made to move away from `&-` practice. So this should be `.aui--search-component-icon`.

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,147 @@
...
-          value={value}
+          value={inputValue}
         />
-        {isLoading ? <Spinner size="small" /> : null}
-        {_.isEmpty(value) ? (
-          <SvgSymbol href={svgSymbolSearch.href} classSuffixes={searchClassSuffixes} />
-        ) : (
-          <SvgSymbol href={svgSymbolCancel.href} classSuffixes={cancelClassSuffixes} onClick={this.onClear} />
+        {isLoading && (
+          <span className={classnames('aui--new-search-spinner', { 'with-btn': searchOnEnter })}>{loaderIcon}</span>
```
veegandhi: I think we have a `Spinner` component which you can use here

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,147 @@
...
-    const { disabled, isLoading, placeholder, svgSymbolCancel, svgSymbolSearch, value } = this.props;
-    const searchClassSuffixes = disabled ? ['color-disabled'] : svgSymbolSearch.classSuffixes;
-    const cancelClassSuffixes = disabled ? ['color-disabled'] : svgSymbolCancel.classSuffixes;
+    const { className, disabled, dts, icons, isLoading, onChange, placeholder, searchOnEnter, value } = this.props;
+
+    const inputValue = value || this.state.value;
+
+    if (value && !onChange)
+      console.warn(
+        'Failed prop type: You provided a `value` prop to Search Component without an `onChange` handler. This will render a read-only field.'
```
veegandhi: You **have** provided a `value`

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,147 @@
...
         />
-        {isLoading ? <Spinner size="small" /> : null}
-        {_.isEmpty(value) ? (
-          <SvgSymbol href={svgSymbolSearch.href} classSuffixes={searchClassSuffixes} />
-        ) : (
-          <SvgSymbol href={svgSymbolCancel.href} classSuffixes={cancelClassSuffixes} onClick={this.onClear} />
+        {isLoading && (
+          <span className={classnames('aui--new-search-spinner', { 'with-btn': searchOnEnter })}>{loaderIcon}</span>
+        )}
+        {searchOnEnter && (
```
veegandhi: I feel ternary is better option here rather then accessing value of `searchOnEnter` twice. Optional change however

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,147 @@
...
     this.onChange = this.onChange.bind(this);
     this.onClear = this.onClear.bind(this);
     this.onKeyPress = this.onKeyPress.bind(this);
+    this.onSearch = this.onSearch.bind(this);
   }
 
   onChange(event) {
-    const { disabled, searchOnChange, onChange } = this.props;
-    if (disabled) return;
+    const eventValue = _.get(event, 'target.value');
```
lightbringer1991: hmm just wondering when it doesn't have the event object

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,147 @@
...
-    onChange(value);
-    if (searchOnChange) {
-      onSearch(value);
+  onSearch(searchValue) {
+    const { onSearch, debounceInterval, value } = this.props;
+    const debounceOnSearch = debounceInterval ? _.debounce(onSearch, debounceInterval) : onSearch;
+    if (value) {
+      debounceOnSearch(searchValue);
+    } else {
+      debounceOnSearch(this.state.value);
```
lightbringer1991: I'm a bit lost here.
so why do we have `searchValue`, `props.value` and `state.value` all affecting the search?

shouldn't it just be either `state.value` or `props.value`, is there a use case where we need to perform searching on a custom string passing from the outside?

src/components/adslot-ui/Search/index.jsx
superpower1: Linter is not happy with ternary

lightbringer1991: linter doesn't like nested ternary.

in this case I think it should be
```jsx
{searchOnEnter ? (...) : null }
```

src/components/adslot-ui/Search/index.jsx
superpower1: Do you mean that the internal onSearch should not take any parameters? Like this:
```
onSearch() {
    const { onSearch, debounceInterval, value } = this.props;
    const debounceOnSearch = debounceInterval ? _.debounce(onSearch, debounceInterval) : onSearch;
    if (value) {
      debounceOnSearch(value);
    } else {
      debounceOnSearch(this.state.value);
    }
  }
```

src/components/adslot-ui/Search/index.jsx
superpower1: I don't know if there is a case when event object will be missing from `<input onChange={onChange} />` binding?

src/components/adslot-ui/Search/index.jsx
lightbringer1991: yea that's what I'm expecting, since it's an internal function

docs/examples/SearchExample.jsx
```diff
@@ -2,154 +2,133 @@ import React from 'react';
...
-      begin typing.
-    </p>
-  ),
+  notes:
+    'Search with search button is commonly used above the pages designed to manage filter pills and search while Search without search button is more commonly used within the pickers and modals.',
   exampleCodeSnippet: `
   <Search
-    value={this.state.value}
-    onChange={this.onChange}
+    placeholder="Search"
```
veegandhi: Make sure the `exampleCodeSnippet` matches the code you have in the actual example

docs/examples/SearchExample.jsx
```diff
@@ -2,154 +2,133 @@ import React from 'react';
...
+            search: React.Node,
+            loader: React.Node,
+            close: React.Node
+          }`,
         },
         {
           propType: 'isLoading',
           type: 'bool',
           defaultValue: 'false',
+          note: '',
```
veegandhi: remove `note` all together

docs/examples/SearchExample.jsx
```diff
@@ -2,154 +2,133 @@ import React from 'react';
...
-          note: 'determine if onSearch() will be fired upon text changes',
-        },
-        {
-          propType: 'searchOnEnterKey',
+          propType: 'searchOnEnter',
           type: 'bool',
           defaultValue: 'false',
-          note: 'determine if onSearch() will be fired upon pressing Enter key',
+          note:
+            'Determine whether onSearch() will be fired use ENTER key (Default behavior is to fire onSearch() when input change)',
```
lteacher: _'Determine**s** whether onSearch() will be fired **on** ENTER key **press** (Default behaviour is to fire onSearch() when **the** input change**s**)'_

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,145 @@
...
+    const search = debounceInterval ? this.debounceOnSearch : onSearch;
+    search(searchValue);
+  }
 
-    onChange(value);
-    if (searchOnChange) {
-      onSearch(value);
-    }
-    onClear(value);
+  onBtnClick() {
```
veegandhi: this is more like `onSearchButtonClick`

src/components/adslot-ui/Search/index.jsx
```diff
@@ -1,116 +1,145 @@
...
+    this.state = {
+      value: '',
+    };
 
     this.onChange = this.onChange.bind(this);
     this.onClear = this.onClear.bind(this);
     this.onKeyPress = this.onKeyPress.bind(this);
+    this.onSearch = this.onSearch.bind(this);
+    this.onBtnClick = this.onBtnClick.bind(this);
+    this.debounceOnSearch = _.debounce(props.onSearch, props.debounceInterval);
```
lteacher: Dont remember if we have class properties but probably wouldn't need all these binds if they were simply arrow functions

src/components/adslot-ui/Search/styles.scss
```diff
@@ -1,31 +1,19 @@
...
-  ::-webkit-search-cancel-button,
-  ::-webkit-search-decoration {
-    appearance: none;
-  }
-
-  &-input {
-    -webkit-appearance: textfield; // For Safari.
+  .aui--search-component-input {
+    flex: 1;
+    height: $search-height + 2;
```
veegandhi: Why is the search input height > the actual component, specially when the actual component is `display:flex`

src/components/adslot-ui/Search/index.jsx
superpower1: I was following the coding style from [https://github.com/airbnb/javascript/tree/master/react](https://github.com/airbnb/javascript/tree/master/react)

> Why? A bind call in the render path creates a brand new function on every single render. Do not use arrow functions in class fields, because it makes them [challenging to test and debug, and can negatively impact performance](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1), and because conceptually, class fields are for data, not logic.

```jsx
// bad
class extends React.Component {
  onClickDiv() {
    // do stuff
  }

  render() {
    return <div onClick={this.onClickDiv.bind(this)} />;
  }
}

// very bad
class extends React.Component {
  onClickDiv = () => {
    // do stuff
  }

  render() {
    return <div onClick={this.onClickDiv} />
  }
}

// good
class extends React.Component {
  constructor(props) {
    super(props);

    this.onClickDiv = this.onClickDiv.bind(this);
  }

  onClickDiv() {
    // do stuff
  }

  render() {
    return <div onClick={this.onClickDiv} />;
  }
}
```

lightbringer1991: hmm I agree with the article in general OOP principles.
However, they raised 3 main points:
- mockability: we can still mock the method via created object, i.e. not through the class object
```jsx
class MyComponent extends React.PureComponent {
  method1 = () => console.log('x');

  render() {
    return null;
  }
}

const wrapper = shallow(<MyComponent />);
sinon.stub(wrapper.instance(), 'method1').callsFake(() => console.log('y'));
```

- Inheritance: Facebook actually discourages inherit react components, and use composition instead. We're following this guideline. https://reactjs.org/docs/composition-vs-inheritance.html

- performance: I think this is the only thing that encourages me to use binding method instead. However I would like to have a live benchmarking to clarify.

superpower1: I think the performance issue will happen in two cases:

1. We have a lot of this component in a same page (e.g. Cell in a large grid) 

1. Those arrow functions are passed as props to its many children (children will re-render because those functions are recreated every time when the parent re-render)

Apparently, these two cases are not applicable to the search component here. But I think we need to consider this in other components. And so for code consistency, maybe we should follow the coding style. I will look into making a showcase for performance comparison in my spear time.

docs/examples/SearchExample.jsx
```diff
@@ -13,11 +13,18 @@ class SearchExample extends React.PureComponent {
     });
   };
 
-  onSearch = () => {
+  onSearch1 = () => {
```
lteacher: Why is there like 1 / 2 on the end of these? looks like you can put some context word in place there.

docs/examples/SearchExample.jsx
```diff
@@ -13,11 +13,18 @@ class SearchExample extends React.PureComponent {
...
     });
-    setTimeout(() => this.setState({ isLoading: false }), 950);
+    setTimeout(() => this.setState({ isLoading1: false }), 950);
+  };
+
+  onSearch2 = value => {
+    this.setState({
+      isLoading2: true,
+    });
+    setTimeout(() => this.setState({ isLoading2: false }, () => alert(`You search for '${value}'`)), 950);
```
lteacher: searched

docs/components/MigrationNote/index.jsx
```diff
@@ -30,6 +30,45 @@ class MigrationNote extends React.Component {
           <div>
             <h2>Migration Guide</h2>
             <hr />
+            <h3>Search Component</h3>
+            <p>
+              The new {`<Search />`} component will replace the old {`<Search />`} amd {`<SearchBar />`} component.
```
veegandhi: The new {`<Search />`} component will merge old {`<Search />`} and {`<SearchBar />`} component.

docs/components/MigrationNote/index.jsx
```diff
@@ -30,6 +30,45 @@ class MigrationNote extends React.Component {
...
+            </p>
+            <p>
+              1. <b>onSearch</b> function is the only required prop. It can work as an uncontrolled component without{' '}
```
veegandhi: You editor has appended `{' '}` at the end.

docs/components/MigrationNote/index.jsx
```diff
@@ -30,6 +30,45 @@ class MigrationNote extends React.Component {
...
+            <SyntaxHighlighter language="jsx" style={coy}>
+              {`
+                {
+                  search: React.Node,
+                  close: React.Node,
+                  loader: React.Node
+                }
+              `}
+            </SyntaxHighlighter>
+            <p>New Search component will use its default icons if none or some of the icons are not provided.</p>
```
veegandhi: New Search component will use ~~its~~ default icons if none or some of the icons are not provided

docs/components/MigrationNote/index.jsx
superpower1: I think this is done by the prettier when I commit the change.

PR #855, opened 2019-05-31
Upgrade node to v10.16

lteacher:
## Description

Upgrades node js to version 10


lteacher:
@xiaofan2406 I did actually put 12 in but there were issues apparently so leaving it off till next upgrade

pphminions:
Merging 🚀

PR #863, opened 2019-06-14
fix: add external to dev build

veegandhi:
## Description

Include external config in dev build webpack to align with dist config as dev builds are used locally.


pphminions:
Merging 🚀

PR #869, opened 2019-06-25
test: remove console warnings in Search tests

superpower1:
1. Add onChange function when value is provided in some tests
2. Remove unnecessary value in some tests
3. Add tests for console warning itself
4. Add tests to cover cases when onChange is not provided

## Description

This PR is to get rid of some annoying warning when running tests on Search component. Warning was showing because onChange is not provided with value in some tests.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Run `npm run test` and see all the warnings are gone.

## Check-list:

### DIFF
--- a/src/components/adslot-ui/Search/index.spec.jsx
+++ b/src/components/adslot-ui/Search/index.spec.jsx
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import React from 'react';
 import { shallow } from 'enzyme';
 import sinon from 'sinon';
@@ -6,12 +7,13 @@ import SvgSymbol from 'alexandria/SvgSymbol';
 import Spinner from 'alexandria/Spinner';
 
 describe('Search', () => {
-  const onSearch = sinon.spy();
+  const onSearch = _.noop;
   const props = {
     className: 'additional-class',
     dts: 'test-dts',
     placeholder: 'search',
     value: 'abc',
+    onChange: _.noop,
     onSearch,
   };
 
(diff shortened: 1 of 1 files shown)


src/components/adslot-ui/Search/index.spec.jsx
```diff
@@ -48,6 +50,21 @@ describe('Search', () => {
     expect(inputEle.prop('value')).to.equal('abc');
   });
 
+  it('should throw warning if value is provided without onChange', () => {
+    const originalConsoleWarn = console.warn;
+    console.warn = sinon.spy();
```
lteacher: For some reason there is no `sinon.restore` anywhere so we should add that to the `afterEach`. Then I am wondering why its not simply `sinon.spy(console, 'warn')` over here?

Shouldnt be any of this reset original stuff (hopefully)

lightbringer1991: I think we can do the following
```js
sandbox.stub(console, 'warn');
....

afterEach(() => sandbox.restore());
```

src/components/adslot-ui/Search/index.spec.jsx
```diff
@@ -6,15 +7,23 @@ import SvgSymbol from 'alexandria/SvgSymbol';
...
     className: 'additional-class',
     dts: 'test-dts',
     placeholder: 'search',
     value: 'abc',
+    onChange: _.noop,
     onSearch,
   };
 
+  before(() => {
+    sandbox = sinon.sandbox.create();
```
lteacher: Super picky comment, but the docs never reference this so maybe if we are creating an explicit sandbox better use the `createSandbox` function

src/components/adslot-ui/Search/index.spec.jsx
```diff
@@ -6,15 +7,23 @@ import SvgSymbol from 'alexandria/SvgSymbol';
 import Spinner from 'alexandria/Spinner';
 
 describe('Search', () => {
-  const onSearch = sinon.spy();
+  let sandbox = null;
+  const onSearch = _.noop;
```
lightbringer1991: imo it's redundant, since you can refer to this as `props.onSearch`

src/components/adslot-ui/Search/index.spec.jsx
superpower1: You're right, I am referring to old version docs. Since we are using v4 sinon I should use `sinon.createSandbox` instead. Just fyi, if we upgrade sinon to v5+, `sinon` object itself will be a sandbox so we don't need to create sandbox anymore, just use it like `sinon.stub()` and `sinon.restore()`

Issue #849, opened 2019-05-13
Conditionally wrap bootstrap button with popover


veegandhi:
## Types of issue

- [ ] Bug (High impact or critical to a production build)
- [x] Bug (Low impact or not working as expected)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Something else (idea, build tool, documentation improvement)

## Description

With the update of popover component, bootstrap button is wrapper with `Popover`  even when `reason` property is not provided.

veegandhi:
The `reason` property should be dropped all together which will remove the usage of `Popover` from the button component.

PR #885, opened 2019-07-17
refactor: deprecate button reason

veegandhi:
## Description

Internally in `Bootstrapbutton`, we wrap `Popover` so user can show a message when button is displayed. This introduces an extra layer of wrapping for the component. Popover are display `inline-flex`, this also makes button styling break intermittently based on parent container's display properties. And putting icons within buttons require more customized styles for each button. This makes is difficult for general use case (which is priority). 

Originally it was decided to conditionally wrap this with popover, however, we now provide an extensive api of Popover so it would be easy for user to implement this.

Hence deprecating `reason` prop from `Button`.

Fixes: #849

- [ ] Yes
- [X] No

### DIFF
--- a/src/components/Button/index.jsx
+++ b/src/components/Button/index.jsx
@@ -7,79 +7,40 @@ import PropTypes from 'prop-types';
 import BootstrapButton from 'react-bootstrap/lib/Button';
 import Spinner from '../Spinner';
 import { expandDts } from '../../lib/utils';
-import Popover from '../Popover';
 import './styles.scss';
 
 const adslotButtonPropTypes = {
   inverse: PropTypes.bool,
-  reason: PropTypes.node,
   dts: PropTypes.string,
   isLoading: PropTypes.bool,
 };
 
-class Button extends React.PureComponent {
-  constructor(props) {
-    super(props);
-    this.buttonRef = React.createRef();
-    this.state = {
-      show: false,
-    };
-    this.handleReasonPopover = this.handleReasonPopover.bind(this);
-  }
+const Button = props => {
+  const { inverse, children, dts, className, isLoading, disabled } = props;
+  const baseClass = 'aui--button';
+  const classes = classNames(baseClass, className, {
+    'btn-inverse': inverse && !/btn-inverse/.test(className),
+  });
 
-  handleReasonPopover(value) {
-    this.setState({ show: value });
-  }
+  const renderSpinner = () =>
+    isLoading ? (
+      <div className="spinner-container">
+        <Spinner size={_.includes(['lg', 'large'], props.bsSize) ? 'medium' : 'small'} />
+      </div>
+    ) : null;
 
-  renderSpinner() {
-    if (this.props.isLoading) {
-      return (
-        <div className="spinner-container">
-          <Spinner size={_.includes(['lg', 'large'], this.props.bsSize) ? 'medium' : 'small'} />
-        </div>
-      );
-    }
-
-    return null;
-  }
-
-  render() {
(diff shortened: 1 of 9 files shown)


src/components/third-party/Button/index.jsx
```diff
@@ -21,14 +19,6 @@ class Button extends React.PureComponent {
   constructor(props) {
     super(props);
     this.buttonRef = React.createRef();
```
lightbringer1991: do we still need this ref?

www/examples/ButtonExample.jsx
```diff
@@ -13,14 +13,20 @@ class ButtonExample extends React.PureComponent {
...
+        >
+          <Button bsStyle="link" disabled={!this.state.canUndo} onClick={this.onClick}>
+            Undo
+          </Button>
+        </Popover>
         <Button bsStyle="primary" disabled={this.state.canUndo} onClick={this.onClick}>
           Apply
         </Button>
-      </div>
+      </React.Fragment>
```
lightbringer1991: I think we can even remove this example. There's no need to show usage of Popover with button

src/components/third-party/Button/index.jsx
omgaz: If not, do we still need the constructor function: that can probably go as well (also hi)

www/mdexamples/TreePicker.mdx
```diff
@@ -83,7 +83,7 @@ const Example = () => {
         onChange={setPickerSearchValue}
         searchOnClear={() => setPickerSearchValue('')}
         includeNode={node => setSelectedNodes(_.concat([], selectedNodes, node))}
-        removeNode={node => setSelectedNodes(_.reject(this.state.selectedNodes, { id: node.id }))}
+        removeNode={node => setSelectedNodes(_.reject(selectedNodes, { id: node.id }))}
```
Harris-YWang: fix a tiny bug on the ui.adslot.com page

src/components/Button/index.jsx
```diff
@@ -7,79 +7,40 @@ import PropTypes from 'prop-types';
...
-          {this.renderSpinner()}
-          <div className={isLoading ? 'button-component-children-container' : null}>{children}</div>
-        </Popover>
-      </BootstrapButton>
-    );
-  }
-}
+const Button = props => {
+  const { inverse, children, dts, className, isLoading, disabled } = props;
+  const baseClass = 'button-component';
```
superpower1: I think we are going to give all ui components a `aui--` prefix, so maybe change it to `aui--button`?

lightbringer1991: this change will lead to some style breaking in other libraries as well. But I'd say let's do it.

Harris-YWang: Fixed and checked relevant files in the project. Any further idea about other libraries? Further changes would be made if we need.

src/components/Button/index.jsx
```diff
@@ -7,79 +7,40 @@ import PropTypes from 'prop-types';
...
-        </Popover>
-      </BootstrapButton>
-    );
-  }
-}
+const Button = props => {
+  const { inverse, children, dts, className, isLoading, disabled } = props;
+  const baseClass = 'aui--button';
+  const classes = classNames(baseClass, className, {
+    'btn-inverse': inverse && !/btn-inverse/.test(className),
```
lightbringer1991: @veegandhi just wondering if we want to go all BEM, e.g. `aui--button__inverse`, `aui--button-spinner`, etc.

veegandhi: Not really. For not lets just hyphenate the class names.

PR #888, opened 2019-07-19
[ADS-2794] feat: add new component PopoverWithRef

### LINKED ISSUE #852: Allow popover to render based on a external reference
Post the migration of Popover to react-popover, it would be nice to extend the component to accept a external ref and render based on the ref position. There are a few usages available on the library itself.

## Types of issue

superpower1:
## Description

Popover.WithRef will accept `refElement` and render the popover based on the position of that element.

Fixes #852 

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/third-party/Popover/WithRef.jsx
+++ b/src/components/third-party/Popover/WithRef.jsx
@@ -1,0 +1,70 @@
+import _ from 'lodash';
+import classnames from 'classnames';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import Popper from './Popper';
+import { themes, popoverPlacements } from './constants';
+import './styles.scss';
+
+const WithRef = ({
+  theme,
+  title,
+  dts,
+  popoverClassNames,
+  popoverContent,
+  refElement,
+  modifiers,
+  wrapperStyles,
+  placement,
+  isOpen,
+  arrowStyles,
+  getContainer,
+  popperRef,
+}) => {
+  const themeClass = _.includes(themes, theme) ? `popover-${theme}` : 'popover-light';
+  const popoverClass = classnames('aui--popover-wrapper', themeClass, popoverClassNames);
+  const boundariesElement = getContainer ? getContainer() : document.body;
+
+  if (!isOpen) return null;
+
+  return ReactDOM.createPortal(
+    <Popper
+      refElement={refElement}
+      placement={placement}
+      modifiers={modifiers}
+      boundariesElement={boundariesElement}
+      popoverClass={popoverClass}
+      wrapperStyles={wrapperStyles}
+      dts={dts}
+      title={title}
+      popoverContent={popoverContent}
+      arrowStyles={arrowStyles}
+      innerRef={popperRef}
+    />,
+    boundariesElement
+  );
+};
+
+WithRef.propTypes = {
+  refElement: PropTypes.instanceOf(Element).isRequired,
+  title: PropTypes.string,
+  theme: PropTypes.oneOf(themes),
+  popoverClassNames: PropTypes.string,
+  getContainer: PropTypes.func,
+  arrowStyles: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+  wrapperStyles: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+  modifiers: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+  placement: PropTypes.oneOf(popoverPlacements),
+  popoverContent: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+  isOpen: PropTypes.bool,
+  dts: PropTypes.string,
+  popperRef: PropTypes.func,
+};
+
+WithRef.defaultProps = {
+  theme: 'light',
+  placement: 'auto',
+};
+
+export default WithRef;
--- a/src/components/third-party/Popover/index.jsx
+++ b/src/components/third-party/Popover/index.jsx
@@ -4,8 +4,10 @@ import classnames from 'classnames';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
-import { Manager, Reference, Popper } from 'react-popper';
+import { Manager, Reference } from 'react-popper';
 import { themes, popoverPlacements } from './constants';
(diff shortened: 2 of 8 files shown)


src/components/third-party/PopoverWithRef/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+class PopoverWithRef extends React.PureComponent {
+  render() {
+    const { theme, title, dts, popoverClassNames, popoverContent, refElement, modifiers, wrapperStyles } = this.props;
+    const themeClass = _.includes(themes, theme) ? `popover-${theme}` : 'popover-light';
+    const popoverClass = classnames('aui--popover-wrapper', themeClass, popoverClassNames);
+
+    if (!this.props.isOpen) return null;
+
+    return ReactDOM.createPortal(
```
lightbringer1991: this will always render Popover at the end of `body`. Is this a desired output?

superpower1: Yes, popover will be rendered at a separate DOM and use absolute position of the `refElement` to decide where to place.

lightbringer1991: you seem to not using any state, so I guess it can be changed to a functional component

src/components/third-party/PopoverWithRef/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+            <div
+              className="aui--popover-arrow"
+              data-placement={placement}
+              ref={arrowProps.ref}
+              style={{ ...arrowProps.style, ...this.props.arrowStyles }}
+            />
+          </div>
+        )}
+      </Popper>,
+      document.body
```
veegandhi: Like `Popover`, this should receive the container from the user.

src/components/third-party/Popover/WithRef.jsx
```diff
@@ -0,0 +1,71 @@
...
+  wrapperStyles,
+  placement,
+  isOpen,
+  arrowStyles,
+  getContainer,
+  popperRef,
+}) => {
+  const themeClass = _.includes(themes, theme) ? `popover-${theme}` : 'popover-light';
+  const popoverClass = classnames('aui--popover-wrapper', themeClass, popoverClassNames);
+  const boundariesElement = getContainer ? getContainer() : document.body;
```
xiaofan2406: Please avoid using `document.body` as the root target for `createPortal` API.

grid_ repo does it for other reasons (should be re-visited too tbh).

You can create a `Portal` component that creates a div on mount, and cleans up on unmount.

xiaofan2406:
Not sure how this decision of creating yet another component that does popover was arrived, maybe its worth thinking about the most common usages we have, and check against the current API to see if the two component make sense to be separate or if there are any other pieces are missing

veegandhi:
> Not sure how this decision of creating yet another component that does popover was arrived, maybe its worth thinking about the most common usages we have, and check against the current API to see if the two component make sense to be separate or if there are any other pieces are missing

We do that in the review session. Would encourage you to attend them

xiaofan2406:
all good if this the api we decided to go with, was just curious.

src/components/third-party/Popover/WithRef.jsx
veegandhi: Resolving this per offline discussion

veegandhi:
Will need an example in the documentation explaining the usage difference in the api.

src/components/third-party/Popover/index.jsx
```diff
@@ -94,39 +96,17 @@ class Popover extends React.PureComponent {
...
-                <div
-                  className="aui--popover-arrow"
-                  data-placement={placement}
-                  ref={arrowProps.ref}
-                  style={{ ...arrowProps.style, ...arrowStyles }}
-                />
-              </div>
-            )}
-          </Popper>,
+            modifiers={this.props.modifiers}
```
veegandhi: `modifiers` implementation should not change here. On existing api, it should still `preventOverflow` and get `boundariesElement` and then override modifiers by props.

src/components/third-party/Popover/index.jsx
```diff
@@ -94,39 +96,17 @@ class Popover extends React.PureComponent {
...
+            boundariesElement={this.getBoundedContainer()}
+            popoverClass={popoverClass}
+            wrapperStyles={this.props.wrapperStyles}
+            dts={dts}
+            title={title}
+            popoverContent={popoverContent}
+            arrowStyles={arrowStyles}
```
veegandhi: `arrowStyles` configuration should not different between both api.

veegandhi: The above applies to `popoverClass`  too.

veegandhi:
@superpower1 - Running local throws a couple of console warnings. Can you resolve it?

src/components/third-party/Popover/index.jsx
superpower1: > `modifiers` implementation should not change here. On existing api, it should still `preventOverflow` and get `boundariesElement` and then override modifiers by props.

No, it doesn't change, I just move the implementation to `Popper.jsx`

src/components/third-party/Popover/index.jsx
superpower1: Same as the `modifier` above, I did not change them, just move them to a separate file `Popper.jsx`.

superpower1:
> @superpower1 - Running local throws a couple of console warnings related to `Popover` . Can you resolve it?

Yeah, I've resolved it.

www/examples/PopoverWithRefExample.jsx
```diff
@@ -0,0 +1,144 @@
...
+
+  togglePopover = () =>
+    this.setState(prevState => ({
+      isOpen: !prevState.isOpen,
+    }));
+
+  render() {
+    return (
+      <>
+        <h3>Popover with Ref</h3>
```
veegandhi: remove this, component name get the title.

www/examples/PopoverWithRefExample.jsx
```diff
@@ -0,0 +1,144 @@
...
+            isOpen={this.state.isOpen}
+          />
+        )}
+      </>
+    );
+  }
+}
+
+const exampleProps = {
+  componentName: 'Popover With Ref',
```
veegandhi: Popover with external ref

PR #891, opened 2019-07-22
chore: upgrade non breaking deps

veegandhi:
## Description

Upgrade all non breaking deps, needs local testing for :
- React 16.8 (deprecation warnings from `src`) 
- Demo Page

- [ ] Yes
- [x] No

## Manual testing step?

Run demo


.eslintrc
```diff
@@ -41,6 +41,7 @@
     "import/no-extraneous-dependencies": 0,
     "react/prefer-stateless-function": 0,
     "react/self-closing-comp": ["error", { "component": true, "html": false }],
+    "react/default-props-match-prop-types":0,
```
veegandhi: We initialize third party component props internally in the extended component.

lightbringer1991: need a space after colon here

package.json
```diff
@@ -69,97 +69,97 @@
...
-    "diff-match-patch": "^1.0.0",
+    "@babel/polyfill": "^7.4.4",
+    "diff-match-patch": "^1.0.4",
     "isomorphic-fetch": "^2.2.1",
-    "lodash": "^4.17.11",
-    "react": "^16.4.2",
-    "react-dom": "^16.4.2",
+    "lodash": "^4.17.15",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
```
knilink: `react` and `react-dom` should be better stay consist with the main project's version, otherwise additional instance will be installed, since it not `peerDependencies`.
Currently the "single react" instance is guaranteed by `dedupe`

veegandhi: @knilink - That's the plan

PR #898, opened 2019-07-25
chore: make class name convention to bem format

lightbringer1991:
- Make sass lint throw error if class name is not in BEM format.
- Added more examples for rendering VerticalNav menu.

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/adslot-ui/VerticalNavigation/index.jsx
+++ b/src/components/adslot-ui/VerticalNavigation/index.jsx
@@ -81,7 +81,7 @@ class VerticalNavigation extends React.Component {
         { 'aui--vertical-navigation-component__menu-item-is-active': index === activeTabIndex },
       ]);
       menuList.push(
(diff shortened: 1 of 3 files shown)


src/components/adslot-ui/VerticalNavigation/index.jsx
```diff
@@ -81,6 +81,7 @@ class VerticalNavigation extends React.Component {
         { 'aui--vertical-navigation-component__menu-item-is-active': index === activeTabIndex },
       ]);
       menuList.push(
+        // eslint-disable-next-line react/no-array-index-key
```
lightbringer1991: we have no other way to distinguish the menu item, therefore we need to use index here

veegandhi: If conditionally rendering is involved. This will confuse react's internal ecosystem. 

But you are right, we don't have much of an option. Think about it now, menu list re-renders regardless, you are better off using `_.uniqueId` to generate. We'll have a think about a long term solution for this later.

PR #14, opened 2015-12-22
Add toggle component

robmunro:
Adding toggle component to library

screen shot 2015-12-22 at 3 18 28 pm

PR #900, opened 2019-07-30
feat: Switch component

adityasabnis:
New toggle component.

## Description
Component that acts like a checkbox however, looks like a switch.

- [ ] Yes
- [X] No

## Manual testing step?
- Use the new Switch component.
- Test the controlled and uncontrolled state workflows.

image

Implements #14


src/components/adslot-ui/ToggleSwitch/index.jsx
```diff
@@ -0,0 +1,38 @@
...
+  state = {
+    isSwitchedOn: this.props.isSwitchedOn,
+  };
+
+  handleSwitch = async () => {
+    const { isSwitchedOn } = this.state;
+
+    await this.props.onSwitch(!isSwitchedOn);
+    this.setState({ isSwitchedOn: !isSwitchedOn });
+  };
```
lightbringer1991: this async seems to be not as `async` as we think, as we all know `this.setState` is an asynchronous operation with a callback option. I think to ensure that the logic is completely resolved, we need to return a promise on `setState`.

```jsx
return new Promise((resolve) => this.setState({ isSwitchedOn: !isSwitchedOn }, () => resolve()));
```

adityasabnis: Refactored the implementation.
Comment is redundant.

adityasabnis:
Tested with:
- Chrome
- Firefox
- Edge
- Safari

src/components/adslot-ui/ToggleSwitch/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+      </span>
+    );
+  }
+}
+
+ToggleSwitch.propTypes = {
+  value: PropTypes.bool,
+  onChange: PropTypes.func,
+  onSwitch: PropTypes.func,
+};
```
lightbringer1991: might need to add `dts` and `className` for future usage.

src/components/adslot-ui/ToggleSwitch/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+    if (onChange) {
+      onChange(eventValue);
+    } else {
+      this.setState({ value: eventValue });
+    }
+
+    if (_.isFunction(onSwitch)) {
+      onSwitch(eventValue);
+    }
```
lightbringer1991: I may need to check with you on the difference between `onChange` and `onSwitch`

src/styles/variable.scss
```diff
@@ -156,3 +156,9 @@ $grid-float-breakpoint: 0; // Stops the navigation responding to breakpoints.
 // === Popover
 $popover-vertical-padding: $spacing-etalon / 6;
 $popover-horizontal-padding: $spacing-etalon / 3;
+
+// Toggle Switch
+$color-switch-background-off: $color-switch-light-red;
+$color-switch-background-on: $color-switch-light-green;
+$color-switch-circle-off: $color-switch-dark-red;
+$color-switch-circle-on: $color-switch-dark-green;
```
lightbringer1991: I think these can be inside `ToggleSwitch/styles.scss`, since it seems to be very specific.

src/components/adslot-ui/ToggleSwitch/index.jsx
knilink: Same question, I though we decided to use `checked/value`, `defaultChecked/Value` in the meeting. 
Probably I misunderstood.

src/components/adslot-ui/ToggleSwitch/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  onChangeHandler = event => {
+    const eventValue = _.get(event, 'target.checked');
+    const { onChange, onSwitch } = this.props;
+
+    if (onChange) {
+      onChange(eventValue);
+    } else {
+      this.setState({ value: eventValue });
```
knilink: Be aware of that `<ToggleSwitch value={value}/>` is not read only(uncontrolled).

src/components/adslot-ui/ToggleSwitch/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import './style.scss';
+
+class ToggleSwitch extends React.Component {
+  state = {
+    value: false,
+  };
+
+  onChangeHandler = event => {
```
knilink: Just FYI, about the naming convention. 
Usually, `onXxxx` is for the component props and `handleXxxx` is for the function instance.
for example `<Component onChange={this.handleChange} onSubmit={this.handleSumit) />`

adityasabnis:
@lightbringer1991 / @superpower1 / @lteacher / @veegandhi / @knilink 

Please give your views/preference on the impl of this component, it has been in the hanging for way too long.
I am ok with either/ or.

thanks

veegandhi:
@adityasabnis - For now you can move ahead with `value`/`defaulValue`/`onChange` setup to handle controlled and uncontrolled behavior. If we face issues while implementing side effect we can address it then. For your use case, simple implementation will suffice.

knilink:
Following `material-ui`'s implementation, it should be.
```jsx
class ToggleSwitch extends React.Component {
  // https://github.com/mui-org/material-ui/blob/5301c24c3be1023efcec7e33bd7e7f030a87c0cb/packages/material-ui/src/internal/SwitchBase.js#L54
  isControlled = 'value' in this.props; // or !_.isNil(this.props.value)

  handleChange = event => {
    // https://github.com/mui-org/material-ui/blob/5301c24c3be1023efcec7e33bd7e7f030a87c0cb/packages/material-ui/src/internal/SwitchBase.js#L82-L88
    // L15 - L19
    if(!this.isControlled) {
      this.setState({ value: eventValue });
    }
    if (onChange) {
      onChange(eventValue);
    }
  }

  render() {
    // https://github.com/mui-org/material-ui/blob/5301c24c3be1023efcec7e33bd7e7f030a87c0cb/packages/material-ui/src/internal/SwitchBase.js#L99
    // L35
    const toggleInputValue = this.isControlled ? value : this.state.value;
  }
}
```

Personally prefer the naming to be `Switch`, `checked`, `defaultChecked` which's following the general naming convention in most of the popular rc libs, but ok with the current naming for now.

src/components/adslot-ui/ToggleSwitch/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+    if (value && !onChange)
+      console.warn(
+        'Failed prop type: You have provided a `value` prop to ToggleSwitch Component without an `onChange` handler. This will render a read-only field.'
+      );
+
+    const toggleInputValue = value || this.state.value;
+
+    return (
+      <span className="toggle-switch">
```
veegandhi: Wondering why is this wrapped in a span? Won't the label be enough? also classname convention is to prefix with `aui--`

adityasabnis: fixed

src/components/adslot-ui/ToggleSwitch/index.jsx
```diff
@@ -0,0 +1,65 @@
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import './style.scss';
+
+class ToggleSwitch extends React.Component {
+  state = {
+    value: _.get(this.props, 'defaultValue', false),
```
vinteo: maybe should be using `defaultProps` for `defaultValue`?

veegandhi: `input` in react accepts `defaultChecked`

adityasabnis: setting `defaultValue` via defaultProps would make the component un-controlled.

src/components/adslot-ui/ToggleSwitch/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  handleChange = event => {
+    const eventValue = _.get(event, 'target.checked');
+    const { onChange } = this.props;
+
+    if (onChange) {
```
veegandhi: You are checking if `onChange` exists or not. You should check  `_.isNil(this.props.value)` which will then throw a console error on if `onChange` does not exists. Which will in turn serve as a guard on top of warning. 

```suggestion
    if (_.isNil(this.props.value)) {
      this.setState({ value: eventValue });
    } else {
      onChange(eventValue);
    }
```

src/components/adslot-ui/Switch/index.jsx
```diff
@@ -0,0 +1,67 @@
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import './style.scss';
+
+class Switch extends React.Component {
```
adityasabnis: Confirmed with Pedro about name change from `ToggleSwitch` -> `Switch`

adityasabnis:
Impl altered as per discussion. thanks

www/examples/SwitchExample.jsx
```diff
@@ -0,0 +1,97 @@
...
+          defaultValue: 'toggle-switch-component',
+        },
+      ],
+    },
+  ],
+};
+
+export default () => (
+  <Example {...exampleProps}>
+    <>
```
veegandhi: Can you move this under the `SwitchExample`'s render class and control the `br` with  styling in `examples/styles.scss`?

src/components/adslot-ui/Switch/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+        <span className="aui--slider round"></span>
+      </label>
+    );
+  }
+}
+
+Switch.defaultProps = {
+  value: '',
+  dts: 'switch-component',
+  className: '',
```
lightbringer1991: I think you don't need to provide an empty class name

src/styles/color.scss
```diff
@@ -119,3 +119,9 @@ $color-carousel-button: rgba($color-gray-darkest, .5);
 
 // Help
 $color-help: $color-gray;
+
+// Switch
+$color-switch-light-green: #def1de;
```
veegandhi: They are generic colors. Should remove name spacing  `switch` from it

www/examples/SwitchExample.jsx
adityasabnis: sure 👍

www/examples/SwitchExample.jsx
```diff
@@ -0,0 +1,97 @@
...
+    <Switch checked={true} onChange={(nextState) => func(nextState)} />
+  `,
+  propTypeSectionArray: [
+    {
+      propTypes: [
+        {
+          propType: 'defaultChecked',
+          type: 'boolean',
+          defaultValue: null,
+          note: 'switch value, if the value is non-controlled',
```
veegandhi: `non-controlled` -> `uncontrolled`

www/examples/SwitchExample.jsx
```diff
@@ -0,0 +1,97 @@
...
+        },
+        {
+          propType: 'className',
+          type: 'string',
+          defaultValue: null,
+        },
+        {
+          propType: 'dts',
+          type: 'string',
+          defaultValue: 'toggle-switch-component',
```
veegandhi: `switch-component`

www/examples/SwitchExample.jsx
```diff
@@ -0,0 +1,97 @@
...
+          type: 'string',
+          defaultValue: '',
+        },
+        {
+          propType: 'onChange',
+          type: 'func',
+          defaultValue: null,
+          note: (
+            <div>
+              This Func is called which Switch value is changed <br />
```
veegandhi: `which Switch`  to just -> `when`

src/components/adslot-ui/Switch/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+    if (onChange) {
+      onChange(targetCheckedValue);
+    }
+  };
+
+  render() {
+    const { defaultChecked, checked, value, onChange, className, dts } = this.props;
+
+    if (!_.isNil(checked) && !_.isNil(defaultChecked))
+      console.warn(
```
knilink: I think simply forwarding `defaultChecked`, `checked` to `input` and let `react` do the warning should be fine.
```jsx
<input
  defaultChecked={defaultChecked}
  checked={toggleInputChecked}
  value={value}
  ...
/>
```

www/examples/SwitchExample.jsx
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  render() {
+    return <Switch checked={this.state.isToggleOn} onChange={this.onChange} />;
+  }
+}
+
+const exampleProps = {
+  componentName: 'Switch',
+  exampleCodeSnippet: `
```
veegandhi: the example snippet should just exactly match with the component's render, except for internal state. `propTypes` below will already explain the function signature

src/components/adslot-ui/Switch/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+    const toggleInputChecked = !_.isNil(checked) ? checked : this.state.checked;
+    return (
+      <label className="aui--switch">
+        <input
+          type="checkbox"
+          checked={toggleInputChecked}
+          value={value}
+          onChange={this.handleChange}
+          className={className}
```
superpower1: I think maybe it's better to put custom class names in `<label/>`.

src/components/adslot-ui/Switch/index.jsx
```diff
@@ -0,0 +1,67 @@
...
+          dts={dts}
```
knilink: I think the `dst` should be simply passed in via "rest", since it's our internal rather than a general `Switch` component feature. [ref](https://github.com/mui-org/material-ui/blob/5301c24c3be1023efcec7e33bd7e7f030a87c0cb/packages/material-ui/src/internal/SwitchBase.js#L137)
```jsx
const { ..., ...inputProps } = this.props;
<input
   ...
  {...inputProps}
/>
```

www/examples/SwitchExample.jsx
adityasabnis: Yes, the `exampleCodeSnippet` does match the renders of the 4 switches being rendered in the example.

The only difference is wrt the onChange prop, where it is defined as `onChange={(nextState) => func(nextState)}` in the snippet to depict the signature, but passed as `_.noop` instead in the examples.

but that doesn't deviate the snippet behavior from the rendered example.

src/components/adslot-ui/Switch/index.jsx
adityasabnis: is there any benefit of doing that?

src/components/adslot-ui/Switch/index.jsx
adityasabnis: can u please explain this @superpower1 , probably with an example?

src/components/adslot-ui/Switch/index.jsx
knilink: @adityasabnis so that you don't have to manually do the warning.
Here is an example about if I do `<Switch defaultValue={false} value={true} />` in `material-ui` https://codesandbox.io/s/material-demo-yxp7c
You can see the react built in warning message in browser console after the component rendered.

adityasabnis: since, it is already impl, i don't see a reason to go back just to remove it, if there is no other benefit.

src/components/adslot-ui/Switch/index.jsx
adityasabnis: @superpower1 : do u mean `aui--switch` is too generic class name for label and needs to be more specific?

src/components/adslot-ui/Switch/index.jsx
adityasabnis: also passing both,
```
<input
  defaultChecked={defaultChecked}
  checked={toggleInputChecked}
  value={value}
  ...
/>```
give warning even only when just `defaultChecked` is passed for un-controlled state.
as, `const toggleInputChecked = !_.isNil(checked) ? checked : this.state.checked;` will evaluate to `false` coz of  `state = { checked: this.props.defaultChecked || false };` and then eventually 

input will be rendered as,
```
<input
  defaultChecked=true
  checked=true
   ...
/>
```
which react will complain about.

File: src/components/adslot-ui/Switch/index.jsx (hunk 19 above)
Comment (adityasabnis) [reply src/components/adslot-ui/Switch/index.jsx]: class name changed to be something more explicit.

File: src/components/adslot-ui/Switch/index.jsx (hunk 17 above)
Comment (knilink) [reply src/components/adslot-ui/Switch/index.jsx]: @adityasabnis sorry, my mistake. confused by the `material-ui` code.
it should be 
```jsx
<input
  defaultChecked={this.props.defaultChecked}
  checked={this.props.checked}
  value={value}
  ...
/>
```
Since the major api feature decided. It's good to merge for me if it's urgent. The reset are internal improvements which can be done latter, in my opinion.

adityasabnis:
Tested with @lightbringer1991

adityasabnis:
Please merge. thanks

Issue #916, opened 2019-10-02
Don't auto close ListPicker modal on apply


romi-h:
## Types of issue
- [ ] Bug (High impact or critical to a production build)
- [ ] Bug (Low impact or not working as expected)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Something else (idea, build tool, documentation improvement)

## Description

Currently clicking `apply` in `ListPicker` calls both handler below that ere specified in props:
- `this.props.modalApply()`
- `this.props.modalClose()`. 

[See this code reference](https://github.com/Adslot/adslot-ui/blob/master/src/components/adslot-ui/ListPicker/index.jsx#L75-L78)

However there can be situation when we don't want to close the modal after clicking `apply`, for example when validations failed or when additional confirmation action is needed.

Hence the proposed change is to not 'auto' close `ListPicker` modal on apply action.

PR #918, opened 2019-10-02
Don't auto close ListPicker modal on apply

romi-h:
## Description
Resolves https://github.com/Adslot/adslot-ui/issues/916

Currently clicking `apply` in `ListPicker` calls both handler below that are specified in props:
- `this.props.modalApply()`
- `this.props.modalClose()`. 

[See this code reference](https://github.com/Adslot/adslot-ui/blob/master/src/components/adslot-ui/ListPicker/index.jsx#L75-L78)

However there can be situation when we don't want to close the modal after clicking `apply`, for example when validations failed or when additional confirmation action is needed.

Hence the proposed change is to not 'auto' close `ListPicker` modal on apply action. 

Note: If we require a non-breaking change, one approach is to add a config prop `closeOnApply` which defaults to `true`.  And if specified `false`, will not close modal on apply. 

- [x] Yes, usage of ListPicker must now call modalClose explicitly on apply. 
- [ ] No

### DIFF
--- a/src/components/adslot-ui/ListPicker/index.jsx
+++ b/src/components/adslot-ui/ListPicker/index.jsx
@@ -74,7 +74,6 @@ class ListPickerComponent extends React.Component {
 
   applyAction() {
     this.props.modalApply(this.state.selectedItems);
-    this.props.modalClose();
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


adityasabnis:
Merging 🚀 🚀

PR #919, opened 2019-10-03
fix: react select border colour

lightbringer1991:
Default react select border colour is blue, which does not align with our existing design.
Update the default code to our light gray colour (see screenshot)

- [ ] Yes
- [x] No

before:
before

after:
after

### DIFF
--- a/src/components/adslot-ui/Select/styles.js
+++ b/src/components/adslot-ui/Select/styles.js
@@ -25,8 +25,15 @@ const defaultStyle = {
     ...styles,
     display: 'none',
   }),
-  control: styles => ({
+  control: (styles, state) => ({
     ...styles,
+    ...(state.isFocused
+      ? {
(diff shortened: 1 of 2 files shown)


src/components/adslot-ui/Select/styles.js
```diff
@@ -25,8 +25,15 @@ const defaultStyle = {
...
     display: 'none',
   }),
-  control: styles => ({
+  control: (styles, state) => ({
     ...styles,
+    ...(state.isFocused
+      ? {
+          boxShadow: '0 0 0 1px #cccccc',
+          ':hover': { borderColor: '#cccccc' },
+          borderColor: '#cccccc',
```
superpower1: `#ccc` will suffice

lightbringer1991: I'd prefer to keep it as is, just for clarity

Issue #897, opened 2019-07-24
Missing dependency `moment`


superpower1:
Cannot resolve dependency `moment`.

## Types of issue

- [x] Bug (High impact or critical to a production build)
- [ ] Bug (Low impact or not working as expected)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Something else (idea, build tool, documentation improvement)

## Description

Steps to reproduce: 

1. Start a new project
2. Run `npm i adslot-ui` 
3. Run `npm start`
4. See error on the screenshot

This is resolved after I manually install moment by running `npm i moment`.

## Screenshots

<img width="786" alt="Screen Shot 2019-07-24 at 4 41 49 pm" src="https://user-images.githubusercontent.com/8103605/61771651-b97c3180-ae33-11e9-8c21-6c56d8dff404.png">

lightbringer1991:
can you link an example repository please.

veegandhi:
This is expected. Moment, lodash, react, react-dom, react-redux and redux are not bundled. They are specified as `externals` in webpack build. However, we should add them as peer dependencies so it will warn the user during install if the listed packages do not exist.

veegandhi:
Keeping this open to add these as peer dependency.

PR #921, opened 2019-10-03
chore: add peer dependencies

lightbringer1991:
add `moment` as a peer dependency
Resolves https://github.com/Adslot/adslot-ui/issues/897

- [ ] Yes
- [x] No

Screen Shot 2019-10-04 at 10 46 25 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -164,6 +164,12 @@
     "react-dom": "^16.8.6",
     "react-popper": "^1.3.3"
   },
+  "peerDependencies": {
+    "moment": ">= 1.0",
+    "react": ">= 16.0",
+    "react-dom": ">= 16.0",
+    "lodash": ">= 4.0"
+  },
   "engines": {
     "node": ">=10"
   },


package.json
```diff
@@ -164,6 +164,9 @@
     "react-dom": "^16.8.6",
     "react-popper": "^1.3.3"
   },
+  "peerDependencies": {
```
lteacher: If we are going to use `peerDependencies` then prob want to be particular about the version range that our codebase is compatible with... well obviously future versions we cant know lol but if our code is compatible with older versions we should include them in the range otherwise it can be pretty offensive as the deps wont install.

I assume our moment usage is compatible with earlier versions than the one listed

lightbringer1991: @lteacher updated the PR to accommodate all moment version >= 1

package.json
```diff
@@ -164,6 +164,9 @@
...
+    "moment": ">= 1.0"
```
veegandhi: I think we have more peed dependencies then just moment.

lightbringer1991: some library that are in the issue are actually in dependency list.
also there is only 1 warning about `moment` when starting the app.

veegandhi: which are still ok to be convert them into peer dependencies. If you look at `index.html` we load them via cdn

lightbringer1991: I checked the cdn that we load.
We use `react`, `react-dom` and `lodash` in the components. Therefore these libraries are required for development.
The remaining library is `moment`, which is included in peer dependency.

Please let me know how should we go forward.

vinteo:
Merging...

PR #922, opened 2019-10-04
test: action panel button tests

NamrataTrivedi:
## Description
Adding buttons test for action panel

- [ ] Yes
- [ ] No


src/components/adslot-ui/ActionPanel/index.spec.jsx
```diff
@@ -22,24 +22,47 @@ describe('ActionPanelComponent', () => {
...
   });
 
-  it('should render as a modal', () => {
-    const wrapper = shallow(
-      <ActionPanel {...makeProps({ isModal: true, size: 'large', actionButton: <Button>Action</Button> })} />
-    );
-
+  it('should render as a modal with close icon', () => {
+    const wrapper = shallow(<ActionPanel {...makeProps({ isModal: true, size: 'large' })} />);
+    console.log(wrapper.debug());
```
vinteo: remove

NamrataTrivedi: done 🤦‍♀

src/components/adslot-ui/ActionPanel/index.spec.jsx
```diff
@@ -22,24 +22,46 @@ describe('ActionPanelComponent', () => {
...
+  it('should render as a modal with close icon', () => {
+    const wrapper = shallow(<ActionPanel {...makeProps({ isModal: true, size: 'large' })} />);
     expect(document.body.classList.contains('modal-open')).to.equal(true);
 
     expect(wrapper.find('.aui--action-panel-modal-wrapper')).to.have.length(1);
     const actionPanelElement = wrapper.find('.aui--action-panel');
     expect(actionPanelElement.prop('className')).to.equal('aui--action-panel is-large action-modal');
 
+    const closeIcon = wrapper.find('.close-button div.close-icon');
+    expect(closeIcon).to.have.length(1);
```
vinteo: should we check if there is no cancel button too ?

src/components/adslot-ui/ActionPanel/index.spec.jsx
```diff
@@ -22,24 +22,46 @@ describe('ActionPanelComponent', () => {
     expect(headerElement).to.have.length(1);
     expect(headerElement.find('.title').text()).to.equal('Action Panel');
 
+    const closeIcon = wrapper.find('.close-button div.close-icon');
```
vinteo: we don't usually include `div` class is enough

src/components/adslot-ui/ActionPanel/index.spec.jsx
NamrataTrivedi: i added a check for text in `should render with defaults` should i add here as well?

vinteo: yes i usually try to check both buttons on both scenarios, one to show and one not to show to make sure both are not showing at the same time

vinteo (dismissed):

PR #937, opened 2019-10-23
fix: add placement support for AlertInput

adityasabnis:
Current AlertInput Popover has a hadcoded bottom placement. we need to allow placement in all directions.

Resolves: #938 

## Description

Allow AlertInput dynamic placement support 

- [ ] Yes
- [x] No

## Manual testing step?

Screenshot included

image

### DIFF
--- a/src/components/adslot-ui/AlertInput/index.jsx
+++ b/src/components/adslot-ui/AlertInput/index.jsx
@@ -54,6 +54,7 @@ export default class AlertInput extends Component {
       className: customClass,
       dts,
       defaultValue,
+      popoverPlacement,
       disabled,
       value,
       type,
@@ -83,7 +84,7 @@ export default class AlertInput extends Component {
           isOpen={this.state.isPopoverVisible && !_.isEmpty(alertMessage)}
           triggers={['disabled']}
           popoverContent={<strong>{alertMessage}</strong>}
-          placement="bottom"
+          placement={popoverPlacement}
           popoverClassNames={popoverClassName}
           theme={theme}
         >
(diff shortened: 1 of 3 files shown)


src/components/adslot-ui/AlertInput/index.jsx
```diff
@@ -128,6 +129,7 @@ AlertInput.propTypes = {
   prefixAddon: PropTypes.node,
   suffixAddon: PropTypes.node,
   alertStatus: PropTypes.oneOf(['success', 'info', 'warning', 'error']),
+  placement: PropTypes.oneOf(['left', 'top', 'top-start', 'top-end', 'bottom-start', 'bottom', 'bottom-end', 'right']),
```
knilink: not sure but feel it should be `popoverPlacement` or `popoverProps: {placement}`.
looking forward for more opinions.

adityasabnis:
Buddy tested with @superpower1

src/components/adslot-ui/AlertInput/index.jsx
superpower1: Yeah, I think maybe `tooltipPlacement` or `popoverPlacement` will give more context.

adityasabnis: I would say makes sense to change it to `popoverPlacement` 👍 as it gives more context to the placement.

probably not `popoverProps: {placement}` atm, as there are other props being straight passed to to popover too. refactoring all popover props will make it a breaking change.

probably needs more discussion in the adslot-ui catchup, if we intend to go down that path.

lightbringer1991: I'd say `popoverPlacement` is the better option here.

adityasabnis: Changed. Please re-review thanks @knilink 
Resolving this issue

PR #940, opened 2019-10-28
refactor: change header of ActionPanel

superpower1:
1. Changed the left padding of header
2. Changed the color of cancel button when hovering

- [ ] Yes
- [x] No

Before:
Screen Shot 2019-10-28 at 11 17 20 am

After:
Screen Shot 2019-10-28 at 11 15 58 am

### DIFF
--- a/src/components/adslot-ui/ActionPanel/styles.scss
+++ b/src/components/adslot-ui/ActionPanel/styles.scss
@@ -36,6 +36,7 @@
     height: 48px;
     color: $color-gray-darkest;
     padding: 12px;
+    padding-left: 18px;
     border-bottom: $border-lighter;
     display: flex;
     justify-content: space-between;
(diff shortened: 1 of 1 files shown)


src/components/adslot-ui/ActionPanel/styles.scss
```diff
@@ -97,7 +98,7 @@
       .actions {
         .btn {
           &:hover {
-            color: $color-blue;
+            color: $color-gray-darker;
```
adityasabnis: won't this change the color of Action button too on hover, instead of just the cancel button?

superpower1: Yeah, you're right, will change that.

superpower1: Changed it to apply on only cancel button.

PR #942, opened 2019-11-04
[ADS-3423] style: remove font-weight-light

superpower1:
## Description

As our design changed, we don't need the font-weight-light (300) any more. All light font-weight will be defaulted to 400.

JIRA: [ADS-3423](https://adslot.atlassian.net/browse/ADS-3423)

- [x] Yes
- [ ] No


superpower1:
> LGTM. Maybe check if in some components that you removed font-weight, is there a chance that the parent has `font-weight: bold`?

Checked all styles for bold font-weight. Also replace `font-weight-light` with `font-weight-normal` in bootstrapOverrides folder in case bootstrap has bold font-weight in some component.

superpower1:
I marked this as breaking change because adslot-grid import and use variable `$font-weight-light` from ui but this should not break direct-web or any other project that is not importing scss variable from ui.

PR #944, opened 2019-11-19
Add fixes & enhancements to tree picker components

lteacher:
## Description

Apparently the tree picker has various defective elements (_highlighted in comments_). This PR aims to introduces changes to resolve some of these epic trolls that still exist. Such as:

- Remove 'fastStatelessWrapper' so that nodes update correctly for regular changes
- Remove constructor bound functions that **include** a captured prop reference
- Adds prop to determine if the search should show or not

- [ ] Yes
- [X] No

### DIFF
--- a/src/components/adslot-ui/TreePicker/Grid/index.jsx
+++ b/src/components/adslot-ui/TreePicker/Grid/index.jsx
@@ -1,7 +1,6 @@
 import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
-import fastStatelessWrapper from 'adslot-ui/fastStatelessWrapper';
 import TreePickerNode from 'adslot-ui/TreePicker/Node';
 import Empty from 'alexandria/Empty';
 import Grid from 'alexandria/Grid';
@@ -11,8 +10,6 @@ import Spinner from 'alexandria/Spinner';
 import TreePickerPropTypes from '../../../prop-types/TreePickerPropTypes';
 
 require('./styles.scss');
-
-export const TreePickerNodeFast = fastStatelessWrapper(TreePickerNode, ['node.id', 'disabled', 'selected']);
 
 const TreePickerGridComponent = ({
   disabled,
@@ -48,7 +45,7 @@ const TreePickerGridComponent = ({
               </div>
             ) : null}
             {_.map(groupedNodes, node => (
-              <TreePickerNodeFast
+              <TreePickerNode
                 key={node.id}
                 {...{
                   disabled,
--- a/src/components/adslot-ui/TreePicker/Nav/index.jsx
+++ b/src/components/adslot-ui/TreePicker/Nav/index.jsx
@@ -21,6 +21,7 @@ const TreePickerNavComponent = ({
   searchOnEnter,
   searchPlaceholder,
   searchValue,
+  showSearch,
   svgSymbolCancel,
   svgSymbolSearch,
 }) => {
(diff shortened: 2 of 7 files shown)


src/components/adslot-ui/TreePicker/Grid/index.jsx
```diff
@@ -12,8 +11,6 @@ import TreePickerPropTypes from '../../../prop-types/TreePickerPropTypes';
 
 require('./styles.scss');
 
-export const TreePickerNodeFast = fastStatelessWrapper(TreePickerNode, ['node.id', 'disabled', 'selected']);
```
lteacher: This didnt really do anything, it implemented a custom `shouldComponentUpdate` on the props listed here. Unfortunately that sucks because if the node is changing, for instance other values on the node are changing they wont update if the reference is changed.

There isn't even anything faster about this anyway since the `PureComponent` checks are just fine and simply ref check the props.

src/components/adslot-ui/TreePicker/Nav/index.jsx
```diff
@@ -39,18 +40,20 @@ const TreePickerNavComponent = ({
...
-        isLoading={isLoading}
-        onClear={onClear}
-        onChange={onChange}
-        onSearch={onSearch}
-        searchOnEnter={searchOnEnter}
-        icons={icons}
-        value={searchValue}
-        placeholder={searchPlaceholder || 'Search'}
-      />
+      {showSearch && (
```
lteacher: Added to hide search when its not desired

src/components/adslot-ui/TreePicker/Node/index.jsx
```diff
@@ -26,30 +26,24 @@ const printAncestorText = node =>
 
 const pathPrefix = ({ type }) => (_.isEmpty(type) ? '' : `${type} in `);
 
-class TreePickerNodeComponent extends React.Component {
-  constructor(props) {
-    if (_.isUndefined(props.node.path) && _.isUndefined(props.node.ancestors)) {
-      throw new Error(`AdslotUi TreePickerNode needs property 'path' or property 'ancestors' for ${props.node}`);
```
lteacher: Moved to render, since undefined checks low cost, once everything was removed dont see the point in leaving it here. Furthermore, in the render it will catch scenarios if the node changes and suddenly is missing these.

Ok with moving if we want to put somewhere else.

src/components/adslot-ui/TreePicker/Node/index.jsx
```diff
@@ -26,30 +26,24 @@ const printAncestorText = node =>
...
+    this.setState({ isLoading: true }, () => this.props.expandNode(this.props.node));
+  };
 
-    this.state = {
-      isLoading: false,
-    };
-    this.setLoadingAndExpandNode = this.setLoadingAndExpandNode.bind(this);
-    this.includeNodeBound = this.props.includeNode.bind(this, this.props.node);
-    this.removeNodeBound = this.props.removeNode.bind(this, this.props.node);
-    if (this.props.expandNode) this.expandNodeBound = this.props.expandNode.bind(this, this.props.node);
```
lteacher: Stuff like this above meant that stale node references were passed through to functions in some cases such as when the nodes have been rendered already and the node prop changes.

src/components/adslot-ui/TreePicker/Node/index.jsx
```diff
@@ -26,30 +26,24 @@ const printAncestorText = node =>
 
 const pathPrefix = ({ type }) => (_.isEmpty(type) ? '' : `${type} in `);
 
-class TreePickerNodeComponent extends React.Component {
-  constructor(props) {
-    if (_.isUndefined(props.node.path) && _.isUndefined(props.node.ancestors)) {
-      throw new Error(`AdslotUi TreePickerNode needs property 'path' or property 'ancestors' for ${props.node}`);
-    }
+class TreePickerNodeComponent extends React.PureComponent {
```
lteacher: Seems fine here as it will check a bunch of props that the user has control over by reference

src/components/adslot-ui/TreePicker/index.jsx
```diff
@@ -156,6 +159,7 @@ TreePickerSimplePureComponent.propTypes = {
   searchPlaceholder: PropTypes.string,
   searchValue: PropTypes.string,
   selectedNodes: PropTypes.arrayOf(TreePickerPropTypes.node.isRequired).isRequired,
+  showSelectionSearch: PropTypes.bool,
```
lteacher: Added to map to the Grid `showSearch` prop

superpower1: Just wondering why not just use `showSearch`?

lteacher: Hmm yeah apparently there are two searches now, but since the other props that refer to the selection search just say search then maybe should change it to `showSearch`

lteacher: Changed it to `showSearch`

lteacher:
Added prop to docs, other comments resolved

adityasabnis:
Does it need to be tested?

lteacher:
@adityasabnis only new feature is the `showSearch` thing which has some unit testing so I don't think that is needed to merge this unless anyone wants to checkout the changes to take a look.

adityasabnis:
Merging 🚀 🚀

Issue #946, opened 2019-11-26
Popover-WithRef arrow position is incorrect


lightbringer1991:
Arrow position from `Popover.WithRef` component is incorrect for `-start` and `-end` placements.

`top-start`
![Screen Shot 2019-11-27 at 9 58 22 am](https://user-images.githubusercontent.com/2588669/69679541-7a6d2e00-10fc-11ea-9ccc-7e285d760fa5.png)

`top-end`
![Screen Shot 2019-11-27 at 9 58 53 am](https://user-images.githubusercontent.com/2588669/69679559-89ec7700-10fc-11ea-85fd-93b73aede266.png)

`bottom-start`
![Screen Shot 2019-11-27 at 9 59 32 am](https://user-images.githubusercontent.com/2588669/69679609-a092ce00-10fc-11ea-983a-660d2b1c4991.png)

`bottom-end`
![Screen Shot 2019-11-27 at 9 59 59 am](https://user-images.githubusercontent.com/2588669/69679626-b0aaad80-10fc-11ea-80c2-e10fb59e35c5.png)

## Types of issue

- [ ] Bug (High impact or critical to a production build)
- [x] Bug (Low impact or not working as expected)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Something else (idea, build tool, documentation improvement)

PR #947, opened 2019-11-27
fix: fix popover with ref showing incorrect arrow position

lightbringer1991:
Resolves https://github.com/Adslot/adslot-ui/issues/946
Fix Popover incorrect arrow position (see issue for screenshots)

- [ ] Yes
- [x] No

after

### DIFF
--- a/src/components/third-party/Popover/Popper.jsx
+++ b/src/components/third-party/Popover/Popper.jsx
@@ -1,8 +1,47 @@
-import React from 'react';
-import PropTypes from 'prop-types';
 import _ from 'lodash';
+import PropTypes from 'prop-types';
+import React from 'react';
 import { Popper as ReactPopper } from 'react-popper';
 import { popoverPlacements } from './constants';
+
+// default arrow position in pixel from the corresponding popover edge
+const DEFAULT_ARROW_POSITION = 12;
+
+/**
+ *
+ * @param {string} placement
+ * @param {object} arrowStyles            user-defined arrow styles
+ * @param {Element} container             anchor container
+ */
+export const renderArrowStyles = (placement, arrowStyles, container = null) => {
+  const horizontalPosition =
+    _.get(container, 'clientWidth') >= DEFAULT_ARROW_POSITION * 2 ? DEFAULT_ARROW_POSITION : null;
+  const verticalPosition =
+    _.get(container, 'clientHeight') >= DEFAULT_ARROW_POSITION * 2 ? DEFAULT_ARROW_POSITION : null;
+
+  let calculatedArrowStyles = {};
+  switch (true) {
+    case _.includes(['bottom-start', 'top-start'], placement) && !_.isNull(horizontalPosition):
+      calculatedArrowStyles = { left: horizontalPosition };
+      break;
+    case _.includes(['bottom-end', 'top-end'], placement) && !_.isNull(horizontalPosition):
+      calculatedArrowStyles = { left: 'auto', right: horizontalPosition };
+      break;
+    case _.includes(['left-start', 'right-start'], placement) && !_.isNull(verticalPosition):
+      calculatedArrowStyles = { top: verticalPosition };
+      break;
+    case _.includes(['left-end', 'right-end'], placement) && !_.isNull(verticalPosition):
+      calculatedArrowStyles = { top: 'auto', bottom: verticalPosition };
+      break;
+    default:
+      calculatedArrowStyles = {};
+  }
+
+  return {
+    ...calculatedArrowStyles,
+    ...arrowStyles, // let user override default configuration
+  };
+};
 
 const Popper = ({
   arrowStyles,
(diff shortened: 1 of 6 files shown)


src/components/third-party/Popover/constants.js
```diff
@@ -2,5 +2,36 @@ import _ from 'lodash';
...
+      break;
+    case _.includes(['left-end', 'right-end'], placement) && _.get(container, 'clientHeight') > 12:
+      refinedArrowStyles = { top: 'auto', bottom: 12 };
+      break;
+    default:
+      refinedArrowStyles = {};
+  }
+
+  return { ...refinedArrowStyles, ...arrowStyles }; // let user override default configuration
+};
```
superpower1: IMO this function should be moved to `index` cos it is not a constant.

src/components/third-party/Popover/constants.js
```diff
@@ -2,5 +2,36 @@ import _ from 'lodash';
...
+/**
+ *
+ * @param {string} placement
+ * @param {object} arrowStyles            user-defined arrow styles
+ * @param {Element} container             anchor container
+ */
+export const renderArrowStyles = (placement, arrowStyles, container = null) => {
+  let refinedArrowStyles = {};
+  switch (true) {
+    case _.includes(['bottom-start', 'top-start'], placement):
```
superpower1: Although it's rare to have width less than 12 but since you've checked container `clientHeight` maybe add a check for `clientWidth`?

src/components/third-party/Popover/constants.js
lightbringer1991: yea but then it'll create a circular dependency. I think it'll be either `helpers.js` or still in `constants.js`

src/components/third-party/Popover/constants.js
lightbringer1991: updated the code to check for width as well, so the custom arrow will be ignored if `size < 12px`

src/components/third-party/Popover/constants.js
superpower1: How come circular dependency?

lightbringer1991: if we add it to `index.jsx`.
`WithRef.jsx` imports `index.js` to use the function.
then `index.jsx` imports `WithRef.jsx` to re-export, hence causing circular dependency.

superpower1: Sorry, my bad, I mean it should be put into `Popper.jsx` and don't need to be exported. Both index and WithRef are using Popper.

src/components/third-party/Popover/WithRef.jsx
```diff
@@ -28,6 +28,8 @@ const WithRef = ({
 
   if (!isOpen) return null;
 
+  const refinedArrowStyles = renderArrowStyles(placement, arrowStyles, refElement);
```
superpower1: I think there is no need to export `renderArrowStyles` from Popper and get the refinedArrowStyles then pass it into Popper again. Can just do calculate refinedArrowStyles in Popper.

src/components/third-party/Popover/index.jsx
```diff
@@ -79,18 +79,7 @@ class Popover extends React.PureComponent {
...
-        arrowStyles = { left: 12 };
-        break;
-      case _.includes(['bottom-end', 'top-end'], this.props.placement):
-        arrowStyles = { left: 'auto', right: 12 };
-        break;
-      default:
-        arrowStyles = {};
-    }
-    arrowStyles = { ...arrowStyles, ...this.props.arrowStyles }; // let user override default configuration
+    const arrowStyles = renderArrowStyles(this.props.placement, this.props.arrowStyles);
```
superpower1: Same as above, just do the calculation inside Popper.

src/components/third-party/Popover/Popper.jsx
```diff
@@ -16,43 +44,47 @@ const Popper = ({
...
-        data-test-selector={dts}
-      >
-        <div className="aui--popover-container">
-          {title ? <div className="popover-title">{title}</div> : null}
-          <div className="popover-content">
-            {_.isFunction(popoverContent) ? popoverContent({ scheduleUpdate }) : popoverContent}
-          </div>
-        </div>
+}) => {
+  const refinedArrowStyles = renderArrowStyles(popperPlacement, arrowStyles, refElement);
```
veegandhi: Little skeptical here on the names, this is more like `calculatedArrowStyles`. Moreover, if you want use the override the arrow style then we probably don't need to send it in function to circle it back, you could just override locally when you pass it on the popper.

src/components/third-party/Popover/constants.js
```diff
@@ -2,5 +2,8 @@ import _ from 'lodash';
 
 export const themes = ['light', 'dark', 'warn', 'error', 'info', 'success'];
 
-const basePlacements = ['auto', 'top', 'right', 'bottom', 'left'];
-export const popoverPlacements = _.flatMap(basePlacements, trigger => [trigger, `${trigger}-start`, `${trigger}-end`]);
+const basePlacements = ['top', 'right', 'bottom', 'left'];
+export const popoverPlacements = [
+  'auto',
```
veegandhi: `auto` is still base placement.

src/components/third-party/Popover/Popper.jsx
lightbringer1991: the logic is a bit complicated so I decided to pull it out so it's easier to test. If the consensus is just put the block into the component then I'll switch over.

PR #949, opened 2019-11-28
style: add body font color

superpower1:
Override bootstrap body text colour from `#333333` to `#5a5a5a`

- [ ] Yes
- [x] No

### DIFF
--- a/src/styles/_bootstrap-custom.scss
+++ b/src/styles/_bootstrap-custom.scss
@@ -64,6 +64,10 @@
 @import './bootstrapOverrides/Modal';
 @import './bootstrapOverrides/Nav';
 
+body {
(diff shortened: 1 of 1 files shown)


src/styles/_bootstrap-custom.scss
```diff
@@ -64,6 +64,10 @@
 @import './bootstrapOverrides/Modal';
 @import './bootstrapOverrides/Nav';
 
+body {
+  color: $color-gray-darker;
```
lightbringer1991: can't we override it in the libraries that we use? Seems to be specific.

superpower1: The idea is to change the body text color in all adslot-ui component.

Issue #603, opened 2017-09-25
Merge adslot-ui and alexandria components into the same components folder


omgaz:

veegandhi:
Prioritising this issue as this will make the new documentation task a lot more easier.

PR #951, opened 2019-12-06
chore: merge alexandria with adslot-ui

lightbringer1991:
Resolves https://github.com/Adslot/adslot-ui/issues/603
Merge `alexandria` and `adslot-ui` directories into `components` directory

```bash
# new structure
src/
  components/
    <react components>
  lib/
  prop-types/
  styles/
```

TODO

- [x] update README file
- [x] update `scaffold` script
- [x] check integrated libraries to update `displayName` if required

Potential breaking due to some `displayName` change.

- [x] Yes
- [ ] No


src/components/Switch/index.jsx
```diff
@@ -42,7 +38,7 @@ class Switch extends React.Component {
           disabled={disabled}
           onChange={this.handleChange}
           className={className}
-          dts={dts}
+          data-test-selector={dts}
```
superpower1: Shouldn't this dts been put on label?

lightbringer1991: I think it's not in this PR's scope. We can decide on where to put the selector when we need it.

PR #952, opened 2019-12-09
[ADS-3552] chore: extend datepicker component to enable/disable inline editing

glenkurniawan-adslot:
## Description
Extend react-datepicker component to enable/disable inline editing

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/DatePicker/index.jsx
+++ b/src/components/DatePicker/index.jsx
@@ -1,0 +1,35 @@
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactDatePicker from 'react-datepicker';
+import './styles.scss';
+
+const adslotDatePickerPropTypes = {
+  disableInlineEditing: PropTypes.bool,
+};
+
+class DatePicker extends React.PureComponent {
+  handleDateChangeRaw = event => {
+    event.preventDefault();
+  };
+
+  render() {
+    const { disableInlineEditing } = this.props;
+
+    const datePickerProps = disableInlineEditing ? { onChangeRaw: this.handleDateChangeRaw } : {};
+
+    return (
+      <div data-test-selector={this.props.dts}>
+        <ReactDatePicker {..._.omit(this.props, _.keys(adslotDatePickerPropTypes))} {...datePickerProps} />
+      </div>
+    );
+  }
+}
+
+DatePicker.propTypes = { ...adslotDatePickerPropTypes };
+
+DatePicker.defaultProps = {
+  disableInlineEditing: false,
+};
+
+export default DatePicker;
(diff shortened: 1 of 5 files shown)


src/components/DatePicker/index.jsx
```diff
@@ -0,0 +1,35 @@
...
+    event.preventDefault();
+  };
+
+  render() {
+    const { disableInlineEditing } = this.props;
+
+    const datePickerProps = disableInlineEditing ? { onChangeRaw: this.handleDateChangeRaw } : {};
+
+    return (
+      <div data-test-selector={this.props.dts}>
```
lightbringer1991: would be nice to have a className for this one, say `date-picker-component`.

src/components/DatePicker/index.spec.jsx
```diff
@@ -0,0 +1,15 @@
...
+
+describe('DatePicker', () => {
+  it('should render with defaults', () => {
+    const component = shallow(<DatePicker className="test" />);
+    expect(component.props().handleDateChangeRaw).to.be.undefined;
+  });
+
+  it('should render with onChangeRaw when disableInlineEditing = true', () => {
+    const component = shallow(<DatePicker className="test" disableInlineEditing={true} />);
+    expect(component.props().hasOwnProperty('onChangeRaw')).to.be.true;
```
lightbringer1991:
```jsx
import ReactDatePicker from 'react-datepicker';
...
expect(component.find(ReactDatePicker).prop('onChangeRaw')).to.be.a('function');
```

same goes with your previous test, I don't think it tests what you think it tests.

src/components/DatePicker/index.spec.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import DatePicker from '.';
+
+describe('DatePicker', () => {
+  it('should render with defaults', () => {
+    const component = shallow(<DatePicker className="test" />);
```
lightbringer1991: we're going towards having this variable name as `wrapper` instead of `component`.

src/components/DatePicker/styles.scss
```diff
@@ -0,0 +1,179 @@
...
+      cursor: pointer;
+      padding-right: 4px;
+
+      &::-ms-clear {
+        display: none;
+      }
+    }
+  }
+
+  &__close-icon{
```
lightbringer1991: space between selector and bracket (not sure why prettier didn't catch this one)

www/examples/DatePickerExample.jsx
```diff
@@ -16,13 +16,26 @@ class DatePickerExample extends React.Component {
...
+        <DatePicker
+          className="form-control"
+          dateFormat="DD MMM YYYY"
+          selected={this.state.startDate}
+          onChange={this.setSelectedDate}
+          placeholderText="Select Date"
+          disableInlineEditing={true}
+          isClearable="true"
+        />
+      </div>
```
lightbringer1991: maybe have a checkbox that toggles the inline editing, so we only need to render 1 datepicker.

PR #953, opened 2019-12-12
[ADS-3552] chore: extend datepicker component to enable/disable inline editing

glenkurniawan-adslot:
## Description
Extend react-datepicker component to enable/disable inline editing
NB: I also extend the test and karma.conf so I can run the test on WSL.. :)

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/DatePicker/index.jsx
+++ b/src/components/DatePicker/index.jsx
@@ -1,0 +1,35 @@
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactDatePicker from 'react-datepicker';
+import './styles.scss';
+
+const adslotDatePickerPropTypes = {
+  disableInlineEditing: PropTypes.bool,
+};
+
+class DatePicker extends React.PureComponent {
+  handleDateChangeRaw = event => {
+    event.preventDefault();
+  };
+
+  render() {
+    const { disableInlineEditing } = this.props;
+
+    const datePickerProps = disableInlineEditing ? { onChangeRaw: this.handleDateChangeRaw } : {};
+
+    return (
+      <div className="aui--date-picker" data-test-selector={this.props.dts}>
+        <ReactDatePicker {..._.omit(this.props, _.keys(adslotDatePickerPropTypes))} {...datePickerProps} />
+      </div>
+    );
+  }
+}
+
+DatePicker.propTypes = { ...adslotDatePickerPropTypes };
+
+DatePicker.defaultProps = {
+  disableInlineEditing: false,
+};
+
+export default DatePicker;
--- a/src/components/DatePicker/index.spec.jsx
+++ b/src/components/DatePicker/index.spec.jsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import sinon from 'sinon';
+import ReactDatePicker from 'react-datepicker';
+import { shallow } from 'enzyme';
+import DatePicker from '.';
+
+describe('DatePicker', () => {
+  it('should render with defaults', () => {
+    const component = shallow(<DatePicker className="test" />);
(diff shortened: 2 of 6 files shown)


src/components/DatePicker/index.jsx
```diff
@@ -0,0 +1,35 @@
...
+    event.preventDefault();
+  };
+
+  render() {
+    const { disableInlineEditing } = this.props;
+
+    const datePickerProps = disableInlineEditing ? { onChangeRaw: this.handleDateChangeRaw } : {};
+
+    return (
+      <div className="date-picker-component" data-test-selector={this.props.dts}>
```
superpower1: Some of our components have `aui-` class name prefix but others haven't, just wondering if `aui-` prefix is the way forward.

src/components/DatePicker/index.spec.jsx
```diff
@@ -0,0 +1,25 @@
...
+    const component = shallow(<DatePicker className="test" />);
+    expect(component.find(ReactDatePicker).prop('onChangeRaw')).to.equal(undefined);
+  });
+
+  it('should render with onChangeRaw when disableInlineEditing = true', () => {
+    const component = shallow(<DatePicker className="test" disableInlineEditing />);
+    expect(component.find(ReactDatePicker).prop('onChangeRaw')).to.be.a('function');
+  });
+
+  it('should render with onChangeRaw when disableInlineEditing = true', () => {
```
superpower1: This description is wrong.

superpower1:
If you have copy over the styles to `src/components/DatePicker/styles.scss`, I think you should delete `src/styles/_react-datepicker-custom.scss`.

src/components/DatePicker/index.spec.jsx
glenkurniawan-adslot: oops.. description has been updated now

glenkurniawan-adslot:
> If you have copy over the styles to `src/components/DatePicker/styles.scss`, I think you should delete `src/styles/_react-datepicker-custom.scss`.

deleted.

src/components/DatePicker/index.jsx
lightbringer1991: yes it is the way forward, maybe change it to `aui-date-picker`

src/components/DatePicker/index.jsx
```diff
@@ -0,0 +1,35 @@
...
+        <ReactDatePicker {..._.omit(this.props, _.keys(adslotDatePickerPropTypes))} {...datePickerProps} />
+      </div>
+    );
+  }
+}
+
+DatePicker.propTypes = { ...adslotDatePickerPropTypes };
+
+DatePicker.defaultProps = {
+  disableInlineEditing: false,
```
lightbringer1991: if we want to disable inline editing in our date picker usages as default, this value should be `true` instead.

glenkurniawan-adslot: no, Vandish and Haofu said that the default should enable inline editing, I assume that's because how react-datepicker behaves as the default

src/components/DatePicker/index.jsx
glenkurniawan-adslot: ok.. changed to aui-date-picker as suggested

superpower1: Following the convention I think it should be `aui--date-picker-component`

glenkurniawan-adslot: hmm.. I found this on the template: return (<div className="aui--x-x-x">
I think aui--date-picker is the most appropriate.. what do you guys think?

lightbringer1991: sounds good to me :D, yes it should be `aui--`

PR #956, opened 2019-12-18
refactor: improve alert input component

### LINKED ISSUE #957: Improve AlertInput component
`AlertInput` structure seems to be more complicated comparing to what it should be.
Proposing a new DOM structure for the component (see attached PR)

## Types of issue

lightbringer1991:
Simplify `AlertInput` DOM structure
Resolves #957 

## Description

`AlertInput` has an unusual DOM structure, and therefore makes the CSS complicated. By removing some layers of the DOM, and implement `Popover.WithRef` into the component, we end up with a super simple structure and easy to understand.
```jsx
// before
<div class="alert-input-component">
  <span class="aui--popover-element">
    <div class="alert-input-component-wrapper">
      <span class="alert-input-component-wrapper-addon">Max</span>
      <span class="alert-input-component-wrapper-flex-wrapper">
        <input class="alert-input-component-wrapper-input" type="text" value="">
      </span>
      <span class="alert-input-component-wrapper-addon">impressions</span>
    </div>
  </span>
</div>

// after
<div class="aui--alert-input">
  <span class="aui--alert-input__addon">Max</span>
  <input class="aui--alert-input__input" value="">
  <span class="aui--alert-input__addon">impressions</span>
</div>
```

Potentially breaking since the DOM structure changes, therefore some styles are invalid.

- [x] Yes
- [ ] No

after

### DIFF
--- a/src/components/AlertInput/index.jsx
+++ b/src/components/AlertInput/index.jsx
@@ -3,18 +3,37 @@ import { Popover } from 'adslot-ui';
 import classnames from 'classnames';
 import React from 'react';
 import PropTypes from 'prop-types';
+import { popoverPlacements } from '../Popover/constants';
 import './styles.scss';
 
-export const baseClass = 'alert-input-component';
+export const baseClass = 'aui--alert-input';
 
 export default class AlertInput extends React.PureComponent {
-  constructor(props) {
-    super(props);
-    this.state = {
-      isFocused: false,
-      isPopoverVisible: false,
-    };
-  }
+  static propTypes = {
+    className: PropTypes.string,
+    dts: PropTypes.string,
+    disabled: PropTypes.bool,
+    prefixAddon: PropTypes.node,
+    suffixAddon: PropTypes.node,
+    alertStatus: PropTypes.oneOf(['success', 'info', 'warning', 'error']),
+    popoverPlacement: PropTypes.oneOf(popoverPlacements),
+    alertMessage: PropTypes.string,
+    onValueChange: PropTypes.func,
+    onBlur: PropTypes.func,
+    onFocus: PropTypes.func,
+  };
+
+  static defaultProps = {
+    alertStatus: 'success',
+    popoverPlacement: 'bottom',
+  };
+
+  state = {
+    isFocused: false,
+    isPopoverVisible: false,
+  };
+
+  componentRef = React.createRef();
 
   handleMouseEnter = () => {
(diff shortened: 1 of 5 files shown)


superpower1:
I am ok with `<Popover.WithRef />` however if you don't provide `getContainer` prop it won't be able able to flip over on the edge when you use it inside a container like a Modal or something. It will use the body as container instead. Don't know if it's necessary to extend this prop to AlertInput.

src/components/AlertInput/index.jsx
```diff
@@ -3,18 +3,37 @@ import { Popover } from 'adslot-ui';
 import classnames from 'classnames';
 import React from 'react';
 import PropTypes from 'prop-types';
+import { popoverPlacements } from '../Popover/constants';
 import './styles.scss';
 
 export const baseClass = 'alert-input-component';
```
superpower1: As we are introducing `aui--` prefix to new component so maybe change this component as well.

lightbringer1991:
> I am ok with `<Popover.WithRef />` however if you don't provide `getContainer` prop it won't be able able to flip over on the edge when you use it inside a container like a Modal or something. It will use the body as container instead. Don't know if it's necessary to extend this prop to AlertInput.

interesting, let me test that out

www/examples/AlertInputExample.jsx
```diff
@@ -1,45 +1,46 @@
...
+    default:
+      return { status: 'success' };
   }
+};
 
-  changeValue(value) {
+class AlertInputExample extends React.Component {
+  state = {
+    hasPrefix: false,
+    hasPostfix: true,
```
superpower1: `hasSuffix` instead? As it's called suffix in the component.

www/examples/AlertInputExample.jsx
```diff
@@ -49,17 +50,51 @@ class AlertInputExample extends React.Component {
...
+          <Checkbox
+            inline
+            name="prefix"
+            checked={this.state.hasPrefix}
+            label="Show Prefix"
+            onChange={this.togglePrefix}
+          />
+          <Checkbox
+            inline
+            name="postfix"
```
superpower1: Same as above, change to `suffix` maybe.

veegandhi:
This issue needs more discussion as `AlertInput` as a component need far more improvement then just its DOM structure. Please open a UI issue linking this card so we could discuss it on our biweekly sessions.

PR #973, opened 2020-02-07
docs: add new docs for component t-z

superpower1:
## Description

Migrate Documentation
T-Z
 ┣ TabExample.jsx
 ┣ TextEllipsisExample.jsx
 ┣ TextareaExample.jsx
 ┣ TileGridExample.jsx
 ┣ TotalsExample.jsx
 ┣ TreePickerExample.jsx
 ┣ UserListPickerExample.jsx
 ┣ VerticalNavigationExample.jsx

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/TextEllipsis/index.jsx
+++ b/src/components/TextEllipsis/index.jsx
@@ -7,6 +7,9 @@ import './styles.scss';
 class TextEllipsisComponent extends React.PureComponent {
   static propTypes = {
     children: PropTypes.node.isRequired,
+    /**
+     * Can use `placement` and `trigger` props from <a href="/popover">Popover</a> to control popover.
+     */
     popoverProps: PropTypes.shape(_.pick(Popover.propTypes, ['placement', 'trigger'])),
   };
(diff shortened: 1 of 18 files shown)


src/components/TextEllipsis/index.jsx
```diff
@@ -7,6 +7,9 @@ import './styles.scss';
 class TextEllipsisComponent extends React.PureComponent {
   static propTypes = {
     children: PropTypes.node.isRequired,
+    /**
+     * Can use `placement` and `trigger` props from <a href="/popover">Popover</a> to control popover.
+     */
     popoverProps: PropTypes.shape(_.pick(Popover.propTypes, ['placement', 'trigger'])),
```
superpower1: Here we are expecting the popover props to be only `placement` and `trigger` but we are not actually stoping user from putting other props in `popoverProps`. Wondering if I should put a filter in `TextEllipsis` to filter out props that is not placement and trigger. Please comment.

PR #974, opened 2020-02-07
New header for docs.

veegandhi:
## Description

- New header format
- Introduction, logos, bells and whistles

- [ ] Yes
- [x] No

Screen Shot 2020-02-13 at 4 06 19 pm

Screen Shot 2020-02-13 at 4 06 27 pm


.gitattributes
```diff
@@ -1 +1,24 @@
-* text eol=lf
+* text=auto
```
veegandhi: Thanks @lteacher for sorting these out 🎉

www/components/Header/index.jsx
```diff
@@ -1,30 +1,37 @@
...
+          <Contributors />
+          <Popover popoverContent="Github" placement="bottom" theme="dark">
+            <a href="https://github.com/Adslot/adslot-ui">
+              <img src="./assets/github.png" />
+            </a>
+          </Popover>
+        </div>
+      </PageTitle>
+    </div>
+  </React.Fragment>
```
lightbringer1991: do we need this `React.Fragment`?

.gitattributes
lteacher: Yeah lol err i did just randomly copy from somewhere else so dunno if it misses anything but prob just has extra stuff that we dont need which am not too worried about anyways.

.gitattributes
```diff
@@ -1 +1,24 @@
...
+*.html       text eol=lf
+*.md         text eol=lf
+*.properties text eol=lf
+*.yml        text eol=lf
+*.json       text eol=lf
+*.config     text eol=lf
+*.inc        text eol=lf
+*.manifest   text eol=lf
+*.rdf        text eol=lf
+*.jade       text eol=lf
```
adityasabnis: `*.pug` instead of `*.jade` ?

adityasabnis: just checked that there are neither pug nor jade files in adslot-ui.
Don't need any changes

adityasabnis:
Merging 🚀

PR #981, opened 2020-02-20
feat: add ref to ActionPanel

### LINKED ISSUE #962: Allow ActionPanel to have ref
Allow ActionPanel to accept `ref`.

## Types of issue

## Description

Sometimes we want popovers inside ActionPanel to flip on edges. This is currently done by passing the `ref` to the inner element.
```jsx
<ActionPanel isModal>
  <div ref={someRef}>...</div>
</ActionPanel>
```

However, this brings a limitation that the popover will flip on body boundary, not on the modal boundary.

This issue proposes adding ref to the top level, and attach it to the top level div.
```jsx
// example usage
<ActionPanel ref={someRef}>...</ActionPanel>
```

Harris-YWang:
## Description

Provide a ref prop to the top level for ActionPanel Component so that Devs could better use this component.

Resolves #962 

Also, refactor this component a bit and modify corresponding test cases.

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/ActionPanel/index.jsx
+++ b/src/components/ActionPanel/index.jsx
@@ -1,49 +1,50 @@
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useEffect } from 'react';
 import Button from 'react-bootstrap/lib/Button';
 import ReactDOM from 'react-dom';
 import './styles.scss';
 
-class ActionPanel extends React.PureComponent {
-  componentDidMount() {
-    if (this.props.isModal) document.body.classList.add('modal-open');
-  }
+const ActionPanel = React.forwardRef((props, ref) => {
+  const { title, className, size, onClose, children, actionButton, isModal, closeIcon } = props;
 
-  componentWillUnmount() {
-    if (this.props.isModal) document.body.classList.remove('modal-open');
-  }
+  const addBodyClass = classname => document.body.classList.add(classname);
+  const removeBodyClass = classname => document.body.classList.remove(classname);
 
-  render() {
-    const { title, className, size, onClose, children, actionButton, isModal, closeIcon } = this.props;
+  useEffect(() => {
+    if (isModal) addBodyClass('modal-open');
 
-    const actionPanel = (
-      <React.Fragment>
-        <div className={isModal ? 'aui--action-panel-backdrop' : 'hide'} />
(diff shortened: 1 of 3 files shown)


src/components/ActionPanel/index.jsx
```diff
@@ -1,49 +1,52 @@
...
-  componentWillUnmount() {
-    if (this.props.isModal) document.body.classList.remove('modal-open');
-  }
+  const addBodyClass = classname => document.body.classList.add(classname);
+  const removeBodyClass = classname => document.body.classList.remove(classname);
 
-  render() {
-    const { title, className, size, onClose, children, actionButton, isModal, closeIcon } = this.props;
+  useEffect(() => {
+    // Set up
```
veegandhi: You can remove these comments.

src/components/ActionPanel/index.jsx
```diff
@@ -55,6 +58,7 @@ ActionPanel.propTypes = {
   actionButton: PropTypes.node,
   closeIcon: PropTypes.node,
   isModal: PropTypes.bool,
+  actionPanelRef: PropTypes.func,
```
veegandhi: just `ref`

lightbringer1991: I think the consensus is using `React.forwardRef` to pass on the ref

Harris-YWang: Solved. `React.forwardRef` does have some new advantages, although it wasn't a breaking update comparing to a custom prop `ref`.

src/components/ActionPanel/index.jsx
```diff
@@ -1,49 +1,50 @@
...
+import React, { useEffect } from 'react';
 import Button from 'react-bootstrap/lib/Button';
 import ReactDOM from 'react-dom';
 import './styles.scss';
 
-class ActionPanel extends React.PureComponent {
-  componentDidMount() {
-    if (this.props.isModal) document.body.classList.add('modal-open');
-  }
+const ActionPanel = React.forwardRef((props, ref) => {
```
lteacher: We do seem to be dropping the `PureComponent` off from a few components, not sure if anyone has thoughts on that

veegandhi: This one should have been dropped as the props has `children` in it. Was redundant in first place

PR #987, opened 2020-03-05
refactor: merge SvgSymbol and SvgSymbolCircle

### LINKED ISSUE #832: Merge SvgCircle & SvgSymbol
## Types of issue

- [ ] New feature (non-breaking change which adds functionality)
- [x] Something else (idea, build tool, documentation improvement)

## Description

Looks like the circle can only be a prop instead of a complete separate component.

Harris-YWang:
## Description

As we want to depreciate `<SvgSymbolCircle {...props} />`, and main project still requires dependency on this component. So, we managed to merge these two component and adding a new prop `isCircle` to support the current use.

Other projects may need update the use of this `<SvgSymbolCircle {...props} />` to `<SvgSymbol isCircle {...props} />` after the new version release of `adslot-ui`.

For the minor function change, we choose to remain previous props `classSuffixes`. Also, a wrapper component to manage background colour is created in place of previous component. 

Resolves #832
 
- [x] Yes
- [ ] No


src/components/SvgSymbol/SvgSymbolWrapper.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { classSuffixHelper } from '../../lib/utils';
+
+const SvgSymbolWrapper = props => {
```
veegandhi: Curious on why do we need a wrapper? Though this would just be a change controlled by class.

www/examples/SvgSymbolExample.jsx
```diff
@@ -4,7 +4,12 @@ import { SvgSymbol } from '../../src';
 
 class SvgSymbolExample extends React.PureComponent {
   render() {
-    return <SvgSymbol href="./assets/svg-symbols.svg#checklist-incomplete" classSuffixes={['70']} />;
+    return (
```
veegandhi: You only need to update `mdx` file. example folder is redundant and pending a removal cleanup

src/components/SvgSymbol/SvgSymbolWrapper.jsx
Harris-YWang: now, we give symbol a circular coloured background (as a wrapper in this case). guess we probably want other kinds of backgrounds/wrapper design in the future? Otherwise, I'll just combine wrapper and symbol into one file.

veegandhi: I don't think if there is a future design then we'll add onto this component. Specially for an SVG wrapping. If this possible to control with `classNames` should be just done using classNames and the wrapper component should be dropped.

Harris-YWang: Great. will do as so

PR #991, opened 2020-03-17
chore: clean up examples folder

glenkurniawan-adslot:
- Delete the old files in www/examples
- Move the files in www/mdexamples folder to www/examples
- Delete the old www/components and move the necessary files to www/containers/components

## Description

Clean up the old www/examples folder (before we go to the mdx route)

- [ ] Yes
- [x] No


www/containers/props.json
```diff
@@ -103,6 +103,7 @@
   "src/components/ActionPanel/index.jsx": [
     {
       "description": "",
+      "displayName": "ActionPanel",
```
Harris-YWang: you can add `ActionPanel.displayName = 'ActionPanel';` to `src/components/ActionPanel/index.jsx`, otherwise, this line will disappear when we run `npm run start`. 

Currently, we have two ways for `displayName`, `AlertInput` and `AvatarComponent` with 'Component' following, maybe we can change some to improve consistency.

adityasabnis:
Merging 🚀 🚀

PR #995, opened 2020-03-31
feat: upgrade carousel

veegandhi:
## Description

Upgrade carousel to react slick to support variable width items and not just images.

- [x] Yes
- [ ] No

image


www/containers/SidePanel.jsx
```diff
@@ -16,14 +16,14 @@ export class SidePanel extends Component {
   };
 
   render() {
-    const activeGroup = routes.find(route => route.path === this.props.location.pathname);
+    const groupedRoutes = _.groupBy(routes, 'group');
```
veegandhi, line 19: Sneaking in a docs change to open all side panel accordions so the search results appear better.

src/components/Carousel/index.jsx
```diff
@@ -1,53 +1,40 @@
 import classNames from 'classnames';
-import Carousel from 'nuka-carousel';
+// import Carousel from 'nuka-carousel';
```
superpower1, line 2: If this is not needed just remove it?

src/components/Carousel/index.jsx
```diff
@@ -1,53 +1,39 @@
...
   const { className, children } = props;
 
   return (
-    <Carousel
-      {...props}
-      renderCenterLeftControls={PrevButton}
-      renderCenterRightControls={NextButton}
-      className={classNames(baseClass, className)}
-    >
+    <Slider {...props} className={classNames(baseClass, className)}>
```
superpower1, line 15: Wondering if we should allow user to use all the props of Slider? Because for some of our components using 3rd party lib we only allow limited props being passed to the component.

veegandhi: Yes. In this instance I am inclined to pass all of them as they are controlled enough by react-slick. And we have already have three different customisations of carousel needed in main project. So limiting them is not adding any value. 

Good question though.

PR #997, opened 2020-04-01
fix: fix react-docgen processing non-component files

lightbringer1991:
Resolve the following (and similar) errors when generating docs.
```bash
Error with path "src/components/Popover/WithRef.jsx": Error: No suitable component definition found.
  Error: No suitable component definition found.
```

- [ ] Yes
- [x] No

### DIFF
--- a/scripts/generate-docs
+++ b/scripts/generate-docs
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+FILES_TO_IGNORE=("index.spec.jsx" "constants.js" "styles.js" "styles.spec.js" "dataProcessor.js" "dataProcessor.spec.js" "mocks.js" "mocks.jsx")
+DIRECTORIES_TO_IGNORE=("fastStatelessWrapper")
+
+COMMAND="react-docgen src/components -o www/containers/props.json --pretty --resolver findAllComponentDefinitions"
+
+for ignoreFile in ${FILES_TO_IGNORE[@]}; do
+  COMMAND="$COMMAND --exclude $ignoreFile"
+done
+
+for ignoreDir in ${DIRECTORIES_TO_IGNORE[@]}; do
+  COMMAND="$COMMAND --ignore $ignoreDir"
+done
+
+eval $COMMAND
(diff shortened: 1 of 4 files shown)


src/components/Popover/WithRef.jsx
```diff
@@ -28,21 +28,25 @@ const WithRef = ({
...
+          wrapperStyles={wrapperStyles}
+          dts={dts}
+          title={title}
+          popoverContent={popoverContent}
+          arrowStyles={arrowStyles}
+          innerRef={popperRef}
+        />,
+        boundariesElement
+      )}
+    </React.Fragment>
```
lightbringer1991, line 49: `react-docgen` cannot identify `ReactDOM.createPortal` as a valid component, therefore putting it into fragment helps the library to parse this file successfully.

scripts/generate-docs
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+FILE_TO_IGNORE=("index.spec.jsx" "constants.js" "styles.js" "styles.spec.js" "dataProcessor.js" "dataProcessor.spec.js" "mocks.js" "mocks.jsx")
```
veegandhi, line 3: should be FILE**S**

scripts/generate-docs
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+FILES_TO_IGNORE=("index.spec.jsx" "constants.js" "styles.js" "styles.spec.js" "dataProcessor.js" "dataProcessor.spec.js" "mocks.js" "mocks.jsx")
+DIRECTORIES_TO_IGNORE=("fastStatelessWrapper")
+
+COMMAND="react-docgen src/components -o www/containers/props.json --pretty --resolver findAllComponentDefinitions"
+
+for ignoreFile in ${FILES_TO_IGNORE[@]}; do
+  COMMAND="$COMMAND --exclude $ignoreFile"
```
bchew, line 9: Does `--exclude` and `--ignore` not allow for an array/multiple?

lightbringer1991: https://github.com/reactjs/react-docgen/issues/126
it uses `node-dir` to parse these options.
I couldn't find an example of an array option, but multiple `--exclude` works.

pphminions:
Merging 🚀

PR #998, opened 2020-04-02
refactor: use node for icons instead of svg symbol href

superpower1:
## Description

This is part of the process to deprecate SvgSymbol in the future. For those components that are using svg symbol href for icons are all changed to use react node for icons which allow users to pass in their own icons.

- [x] Yes
- [ ] No

### DIFF
--- a/www/containers/props.json
+++ b/www/containers/props.json
@@ -1206,19 +1206,6 @@
           "required": false,
           "description": ""
         },
-        "svgSymbol": {
-          "type": {
-            "name": "shape",
-            "value": "SvgSymbol.propTypes",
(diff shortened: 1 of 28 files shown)


www/containers/props.json
```diff
@@ -2208,9 +2189,7 @@
         },
         "emptySvgSymbol": {
           "type": {
-            "name": "shape",
-            "value": "SvgSymbol.propTypes",
-            "computed": true
+            "name": "node"
           },
```
Harris-YWang, line 2193: L2426 not updated

pphminions:
Merging 🚀

PR #1003, opened 2020-04-07
feat: rich-text-editor

veegandhi:
## Description

Create a new wysiwyg editor based of draft js (https://draftjs.org/)

- [ ] Yes
- [x] No

image


src/components/RichTextEditor/ToolbarButton.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button } from 'adslot-ui';
+import classnames from 'classnames';
+
+function ToolbarButton({ onToggle, label, active }) {
```
superpower1, line 6: Just wondering why not using an arrow function?

veegandhi: The both are the same, but I'll change it for consistency to match the arrow function syntax used across. 

I use the react snippets to generate the components scaffold which creates them as function 😄

adityasabnis:
Merging 🚀 🚀

PR #1008, opened 2020-04-15
refactor: change action panel title type to node

superpower1:
In the new design of product detail modal, we have the publisher logo image and some action buttons on title so Action Panel needs to support custom title as react node.                             

## Description

Change `title` prop type to node. 

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/ActionPanel/index.jsx
+++ b/src/components/ActionPanel/index.jsx
@@ -25,7 +25,7 @@ const ActionPanel = React.forwardRef((props, ref) => {
       <div className={classNames('aui--action-panel-wrapper', { 'aui--action-panel-modal-wrapper': isModal })}>
         <div className={classNames('aui--action-panel', className, `is-${size}`, { 'action-modal': isModal })}>
           <div className={classNames('aui--action-panel-header', { 'has-actions': actionButton })}>
-            <span className="title">{title}</span>
+            <div className="title">{title}</div>
             <span className="actions">
               <Button
                 onClick={onClose}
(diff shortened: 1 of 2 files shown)


src/components/ActionPanel/index.jsx
```diff
@@ -25,7 +26,7 @@ const ActionPanel = React.forwardRef((props, ref) => {
       <div className={classNames('aui--action-panel-wrapper', { 'aui--action-panel-modal-wrapper': isModal })}>
         <div className={classNames('aui--action-panel', className, `is-${size}`, { 'action-modal': isModal })}>
           <div className={classNames('aui--action-panel-header', { 'has-actions': actionButton })}>
-            <span className="title">{title}</span>
+            {_.isString(title) ? <span className="title">{title}</span> : title}
```
veegandhi, line 29: I think you should just render title within the span even if its a node.

veegandhi:
You'll need to restart the dev server and push the prop changes with the commit I think so it will be updated in the documentation.

src/components/ActionPanel/index.jsx
superpower1: Don't know about that, if user give a title as node, I think they won't like a wrapper with predefined styles which they may need to overwrite. But I am ok either way.

veegandhi: The reason why that title is in a span because there are text size and elipsis requirement. Even if they provide the node, they should not change. If overwriting is required it should be done in local project.

veegandhi: What you could change though is to change that span to a div so they get a block element to render the custom node.

PR #1011, opened 2020-04-27
feat: enable <TileGrid> item title to be react node

### LINKED ISSUE #1010: Custom Tile Grid
Sometimes we need to put icon in the title of each item of Tile Grid, which requires us to change the expected type from `string` to `React.Node`

## Types of issue

- [X] New feature (non-breaking change which adds functionality)

## Description

Change the `title` prop type in `Item Object` from `string` to `React.Node` which allows users to put any kind of content.

Harris-YWang:
## Description

- Enable the key title of the prop `items` to be `React.Node` to receive svg icon as part of it.

- The prop `title` of the `<TileGrid />` component should not be required, as sometimes there is no title for it.

- btw, fixing up warnings and errors of homepage shown in the browser console  

Resolves #1010 

- [ ] Yes
- [X] No

### DIFF
--- a/src/components/TileGrid/index.jsx
+++ b/src/components/TileGrid/index.jsx
@@ -50,7 +50,7 @@ const TileGrid = ({ title, items, onItemClick, distributed }) => {
 
   return (
     <div className={baseClass}>
-      <strong className={`${baseClass}-title`}>{title}</strong>
+      {title ? <strong className={`${baseClass}-title`}>{title}</strong> : null}
       <ul className={`${baseClass}-list`}>{cardList}</ul>
     </div>
   );
@@ -61,7 +61,7 @@ TileGrid.defaultProps = {
 };
 
 TileGrid.propTypes = {
-  title: PropTypes.string.isRequired,
+  title: PropTypes.node,
   /**
    * 	The shape of item object is defined below
    */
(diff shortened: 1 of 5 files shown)


src/components/TileGrid/index.jsx
```diff
@@ -61,15 +61,15 @@ TileGrid.defaultProps = {
 };
 
 TileGrid.propTypes = {
-  title: PropTypes.string.isRequired,
+  title: PropTypes.string,
```
lightbringer1991, line 64: should we also make this title node? We seem to just have icons everywhere in the title these days.

Harris-YWang: sounds good

src/components/TileGrid/index.jsx
```diff
@@ -50,7 +50,7 @@ const TileGrid = ({ title, items, onItemClick, distributed }) => {
 
   return (
     <div className={baseClass}>
-      <strong className={`${baseClass}-title`}>{title}</strong>
+      {title ? <strong className={`${baseClass}-title`}>{title}</strong> : null}
```
veegandhi, line 53: title is required prop. ternary is probably redundant here

www/containers/Props.jsx
```diff
@@ -51,7 +51,7 @@ const Props = ({ componentName, customMapper }) => {
                     <code>{HtmlParser(_.get(prop, 'required', ''))}</code>
                   </td>
                   <td>
-                    <code>{HtmlParser(_.get(prop, 'defaultValue.value', ''))}</code>
+                    <code>{_.get(prop, 'defaultValue.value', '')}</code>
```
veegandhi, line 54: I think this will impact a lot of other docs.

Harris-YWang: manually checked, it looks good for now

Issue #1013, opened 2020-04-28
Theme and color management 


Harris-YWang:
Recently, more components have their custom theme/color design, however, the props and parameter names are not very consistent in the whole project. we need enhancement or refactor sometime to improve it.

## Types of issue

- [ ] Bug (High impact or critical to a production build)
- [ ] Bug (Low impact or not working as expected)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Something else (idea, build tool, documentation improvement)

## Description

Here are some ideas, please review them and mark them if reasonable and feasible.

- [ ] `color` and `colour` are both used in the project, like we name prop `color` for `<Avatar />` component while we have `colourStyle` prop for `<Spinner />`. It's better to use only one spelling.

- [ ] the parameter names are not consistent. for `<Popover />`, users use `themes = warn` while for `<Alert />`, users use `type = warning` for a quite similar purpose.

- [ ] the color aliases are not consistent. 

  - for a blue color `#006dcc`, the alias used in `<Button />` is `primary` while the alias used in `<CountBadge />` is `info`.
  - for a cyan color `#49afcd`, the alias used in other components (like `<Alert />`, `<Button />` ...) is `info`.
  - similar cases also happen to the alias `light`

- [ ] we should also manage the color when the props `theme` is undefined, like using `default` as a default value of this `prop`. Now, if `theme` is undefined, the color cannot be predicted before manual test.
- [ ] It looks like we are gonna have two color systems after `<StatusPill />`. This component uses the common aliases but it means different color codes (here `primary` means `#c1e1ec`) from others. If we want a second system, we might need to give these aliases remarkably names instead of keeping using old aliases`['primary', 'success', 'warning', 'error', 'light']`  

- [ ] for different components, we have different prop names to manage its theme or color. Although it makes sense for a specific component, we could re-evaluate if we need to improve the consistency of the whole library.

  - `type` for `<Alert />`
  - `color` for `<Avatar />`
  - `bsStyle` for `<Button />`
  - `status` for `<CountBadge />`
  - `themes` for `<Popover />`
  - `colourStyle` for `<Spinner />`
  - `displayStyle` for `<StatusPill />`
  - `accent` for `<Tag />`

[...]

PR #1014, opened 2020-04-28
feat: add theme to InformationBox

### LINKED ISSUE #994: Add prop 'theme' to <InformationBox />
Sometimes, we may want a different theme/type of `<InformationBox />` in different scenarios.

Especially, we may need this box with different designs of the icon used here and different background color to imply different accents/meanings.

Maybe we can add `theme` prop which is similar to the one used with the component `<StatusPill />`, so `theme` could be one of  `["primary", "success", "warning", "error", "light"]`.

> Also, consider the `edit` link to trigger something on the right/top corner.

For more use cases, please refer to the screenshot part.

## Types of issue

## Screenshots

Currently, we have only one theme/type of InformationBox in `adslot-ui`, the icon has a default white border surrounding the icon
Screen Shot 2020-03-27 at 4 47 57 pm
Alternatively, we may have a design like below one:
Screen Shot 2020-03-27 at 5 43 50 pm

Harris-YWang:
## Description

- Add a new prop `theme` to `<InformationBox />` to take one of `['primary', 'success', 'warning', 'error', 'light']` for easy use in other projects, set `light` to be default as it's completely the same as the legacy design.

- Remove the redundant icon halo, add a new sample to the doc.

- Refine prop types using for theme/color and add descriptions for docs.

- One issue here is we are using `#dbf0fd` for `primary` color of the info box, however, this color is inconsistent to the current `adslot-ui` library. So, choose a very similar color `#e0f0ff` from the current color palette as a new alternative.

- new use cases of `<InformationBox />` (checked with Pedros)
Screen Shot 2020-05-08 at 11 21 00 am

Resolves #994

FYI, further talks about color management and enhancement, please refer to #1013 

- [x] Yes
- [ ] No

### DIFF
--- a/src/components/Button/index.jsx
+++ b/src/components/Button/index.jsx
@@ -17,10 +17,14 @@ const adslotButtonPropTypes = {
   inverse: PropTypes.bool,
   dts: PropTypes.string,
   isLoading: PropTypes.bool,
+  /**
+   * PropTypes.oneOf(['primary', 'success', 'info', 'warning', 'danger', 'link'])
+   */
+  bsStyle: PropTypes.oneOf(['primary', 'success', 'info', 'warning', 'danger', 'link']),
 };
 
 const Button = props => {
-  const { inverse, size, children, dts, className, isLoading, disabled } = props;
+  const { bsStyle, inverse, size, children, dts, className, isLoading, disabled } = props;
   const baseClass = 'aui--button';
   const classes = classNames(baseClass, className, {
     'btn-inverse': inverse && !/btn-inverse/.test(className),
@@ -36,6 +40,7 @@ const Button = props => {
 
   return (
     <BootstrapButton
+      bsStyle={bsStyle}
       {..._.omit(props, _.keys(adslotButtonPropTypes))}
       disabled={isLoading || disabled}
       className={classes}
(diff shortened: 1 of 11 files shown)


src/components/Button/index.jsx
```diff
@@ -36,6 +40,7 @@ const Button = props => {
 
   return (
     <BootstrapButton
+      {...{ bsStyle }}
```
veegandhi, line 43: `bsStyle={bsStyle}`

src/components/Button/index.jsx
```diff
@@ -17,10 +17,14 @@ const adslotButtonPropTypes = {
...
   isLoading: PropTypes.bool,
+  /**
+   * PropTypes.oneOf(['primary', 'success', 'info', 'warning', 'danger', 'link'])
+   */
+  bsStyle: PropTypes.oneOf(['primary', 'success', 'info', 'warning', 'danger', 'link']),
 };
 
 const Button = props => {
-  const { inverse, size, children, dts, className, isLoading, disabled } = props;
+  const { bsStyle, inverse, size, children, dts, className, isLoading, disabled } = props;
```
superpower1, line 27: Wondering why you have to take `bsStyle` out and pass it into BootstrapButton without any changes? As I know, `bsStyle` is one of the bootsrap button prop that will be given to it on L44.

Harris-YWang: As `bsStyle` is a very common and useful `prop` used everywhere, so adding this `prop` doc here for convenience. WIth this one, people don't have to find out BootstrapButton doc to check the expected values.

PR #1017, opened 2020-05-06
chore: use common eslint rules

veegandhi:
## Description

- extent adslot's eslint
- fix linting errors
- drop code sandbox as no more needed

- [ ] Yes
- [x] No


.eslintrc
```diff
@@ -1,101 +1,15 @@
...
     "expect": false
   },
-  "extends": ["airbnb", "prettier"],
-  "plugins": ["lodash", "react"],
+  "extends": "adslot",
   "settings": {
     "lodash": {
       "version": 4
-    },
-    "import/resolver": {
```
xiaofan2406, line 23: should double check this resolver, its alias related.

package.json
```diff
@@ -83,29 +83,25 @@
...
-    "eslint-plugin-import": "^2.8.0",
-    "eslint-plugin-jsx-a11y": "^6.0.3",
-    "eslint-plugin-lodash": "^2.6.1",
-    "eslint-plugin-react": "^7.6.1",
+    "eslint": "^6.8.0",
+    "eslint-config-adslot": "^0.1.0",
+    "eslint-loader": "^4.0.2",
+    "eslint-plugin-import": "^2.20.2",
+    "eslint-plugin-lodash": "^7.1.0",
+    "eslint-plugin-react": "^7.19.0",
```
xiaofan2406, line 104: eslint-plugin-import
eslint-plugin-lodash
eslint-plugin-react

no longer needed

veegandhi: I know right, for some reason it is complaining its a requirement for "eslint-config-adslot". I'll see what's causing this.

.eslintrc
veegandhi: this is ok to drop. It was introduced as part of airbnb config extension of v8.  we should be on v18 now and webpack configs were updated since then.

PR #1019, opened 2020-05-07
fix: default button issue

Harris-YWang:
## Description

Align the default color with the current color palette used.

As `<Button />` can be given props `bsStyle = "default", inverse={true}`, a tweak here is to remove focus issue of this kind of button.

- [ ] Yes
- [x] No


pphminions:
merging 🚀

PR #1023, opened 2020-05-11
feat: new image cropper component

superpower1:
## Description

```
ImageCropper.propTypes = {
  title: PropTypes.string,
  src: PropTypes.string.isRequired,
  alt: PropTypes.string,
  onCrop: PropTypes.func.isRequired,
  onCancel: PropTypes.func.isRequired,
  width: PropTypes.number,
  height: PropTypes.number,
};
```

- [ ] Yes
- [x] No

Screen Shot 2020-05-11 at 4 40 58 pm

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -73,7 +73,7 @@ module.exports = webpackMerge(commonConfig, {
         },
       },
       {
-        test: /\.scss/,
+        test: /\.((c|sc)ss)$/i,
         use: [
           'style-loader',
           {
--- a/src/components/ImageCropper/index.jsx
+++ b/src/components/ImageCropper/index.jsx
@@ -1,0 +1,72 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import Cropper from 'cropperjs';
+import 'cropperjs/dist/cropper.css';
+import Button from '../Button';
+import ActionPanel from '../ActionPanel';
+import './styles.scss';
+
+// https://github.com/fengyuanchen/cropperjs/blob/v2/README.md
+const mapOptions = {
+  dragMode: 'move',
+  viewMode: 2,
+  guides: false,
+  movable: true,
+  cropBoxMovable: true,
+  cropBoxResizable: true,
+  toggleDragModeOnDblclick: false,
+};
+
+const ImageCropper = ({ title, src, alt, onCancel, onCrop, width, height }) => {
+  const cropperRef = React.useRef();
+  const imageRef = React.useRef();
+  const [isSaving, setIsSaving] = useState(false);
+
+  React.useEffect(() => {
+    cropperRef.current = new Cropper(imageRef.current, mapOptions);
+
+    return () => {
+      setIsSaving(false);
+      cropperRef.current.destroy();
+    };
+  }, []);
+
+  const uploadButton = (
+    <Button
+      bsStyle="primary"
+      onClick={async () => {
+        setIsSaving(true);
+        await onCrop(cropperRef.current.getData());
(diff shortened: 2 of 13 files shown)


src/components/ImageCropper/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+      isLoading={isSaving}
+    >
+      Upload
+    </Button>
+  );
+
+  return (
+    <ActionPanel title="Image Upload" onClose={onCancel} actionButton={uploadButton}>
+      <div className="image-cropper" style={{ width: width || '100%', height: height || '100%' }}>
+        <img src={src} ref={imageRef} style={{ maxWidth: '100%', display: 'block' }} alt={alt} />
```
adityasabnis, line 49: why inline styles?

config/webpack.config.dev.js
```diff
@@ -73,7 +73,7 @@ module.exports = webpackMerge(commonConfig, {
         },
       },
       {
-        test: /\.scss/,
+        test: /\.((c|sa|sc)ss)$/i,
```
lightbringer1991, line 76: hmm can't find any `.sass` file, can we revert this?

src/components/ImageCropper/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+    cropperRef.current = new Cropper(imageRef.current, mapOptions);
+
+    return () => {
+      cropperRef.current.destroy();
+    };
+  }, []);
+
+  const uploadButton = (
+    <Button
+      onClick={async () => {
```
Harris-YWang, line 35: Normally, we are using `<Button bsStyle="primary" inverse />` as the 'actionButton' of the `<ActionPanel />`, otherwise, we might have to give a `prop` for custom button design

www/examples/ImageCropper.mdx
```diff
@@ -0,0 +1,25 @@
...
+```jsx live=true
+const Example = () => {
+  const [data, setData] = React.useState('');
+  const onCrop = value => setData(JSON.stringify(value));
+
+  return (
+    <>
+      <ImageCropper src="./assets/tileGrid/example-image.jpg" onCancel={_.noop} onCrop={onCrop} height={400} />
+      <br />
+      <textarea
```
Harris-YWang, line 14: maybe `<Textarea />` is better here

src/components/ImageCropper/index.jsx
superpower1: Will move the style to css file

config/webpack.config.dev.js
superpower1: Just want to support css file but yes we don't have any sass file so will change it to `/\.((c|sc)ss)$/i`.

src/components/ImageCropper/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+        setIsSaving(true);
+        await onCrop(cropperRef.current.getData());
+        setIsSaving(false);
```
knilink, line 38: might set state after unmount.

not sure whether it's better to use 'loading/saving' props and let user handle the loading state.

src/components/ImageCropper/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+      }}
+      isLoading={isSaving}
+    >
+      Upload
+    </Button>
+  );
+
+  return (
+    <ActionPanel title="Image Upload" onClose={onCancel} actionButton={uploadButton}>
```
veegandhi, line 47: Change of plans sightly. Lets make `title` a prop which defaults to `Image Upload`

PR #1024, opened 2020-05-14
fix: button css overwrite issue

Harris-YWang:
## Description

After discussion with @veegandhi, we determine to fix `<Button />` css overwriting problem with `adslot-ui`. In this way, we can avoid breaking changes.

The issue was caused by `btn-default` when the component was not explicitly defined its `bsStyle`, it will give a `btn-default` and if we give the component with `className="btn-primary"`, the `.btn-primary` the selector doesn't overweigh the `btn-default` selector, which means the css wouldn't be overwritten by expected style.

For this problem, changing `btn-default` css styles to `aui--btn-default` and set this classname as per the required situation.

Now, the `<Button />` should behave its way without any breaking changes.

*Buddy test is required for this project only.

- [ ] Yes
- [X] No

Screen Shot 2020-05-14 at 4 07 28 pm

### DIFF
--- a/src/components/Button/index.jsx
+++ b/src/components/Button/index.jsx
@@ -29,6 +29,10 @@ const Button = props => {
   const classes = classNames(baseClass, className, {
     'btn-inverse': inverse && !/btn-inverse/.test(className),
     'btn-large': size === 'large',
+    'aui--btn-default':
+      (!bsStyle || bsStyle === 'default') &&
+      (!className ||
+        ['btn-default', 'btn-inverse', 'btn-default btn-inverse', 'btn-inverse btn-default'].includes(className)),
   });
 
   const renderSpinner = () =>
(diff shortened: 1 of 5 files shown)


src/components/Button/index.jsx
```diff
@@ -29,6 +29,10 @@ const Button = props => {
   const classes = classNames(baseClass, className, {
     'btn-inverse': inverse && !/btn-inverse/.test(className),
     'btn-large': size === 'large',
+    'aui--btn-default':
+      !bsStyle &&
+      (!className ||
+        ['btn-default', 'btn-inverse', 'btn-default btn-inverse', 'btn-inverse btn-default'].indexOf(className) > -1),
```
veegandhi, line 35: You can just use [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)

PR #1026, opened 2020-05-25
feat: add google analytics tracking

vinteo:
## Description

Add Google Analytics tracking for production Adslot UI website

- [ ] Yes
- [x] No

### DIFF
--- a/www/containers/GoogleAnalytics.jsx
+++ b/www/containers/GoogleAnalytics.jsx
@@ -1,0 +1,22 @@
+import { useEffect } from 'react';
+import { withRouter } from 'react-router-dom';
+
+const GA_TRACKING_ID = 'UA-18273448-24';
+
+const GoogleAnalytics = ({ history, location }) => {
+  useEffect(() => {
+    const gtag = window.gtag;
+
+    if (history.action === 'PUSH' && typeof gtag === 'function') {
+      gtag('config', GA_TRACKING_ID, {
+        page_title: document.title,
+        page_location: window.location.href,
+        page_path: location.pathname,
+      });
+    }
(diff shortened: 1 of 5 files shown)


www/containers/props.json
```diff
@@ -1863,6 +1863,17 @@
       "displayName": "ImageCropper",
       "methods": [],
       "props": {
+        "title": {
```
vinteo, line 1866: not sure why this was added automagically?

www/containers/GoogleAnalytics.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { withRouter } from 'react-router-dom';
+
+const GA_TRACKING_ID = 'UA-18273448-24';
+
+class GoogleAnalytics extends React.Component {
+  componentWillUpdate({ location, history }) {
```
veegandhi, line 7: this one is an unsafe life cycle. You can probably just `useEffect` on location and history

www/containers/props.json
veegandhi: I think the it was previously not committed with the feature

www/index.html
```diff
@@ -1,13 +1,24 @@
-<!doctype html>
+<!DOCTYPE html>
```
veegandhi, line 1: I know it's slightly confusing but this html is for dev bundling purposes so we don't load external dependency. For production we use the `index.html` in `src`. So this code will belong there.

vinteo: VSCode formatted the HTML, no changes in this file now besides indentation

pphminions:
merging 🚀

PR #1031, opened 2020-06-05
[ASC-65] refactor: test cases with jest and reac-testing-library

Harris-YWang:
## Description

It's a good taste to migrate the testing cases from `chai`, `karma`, and `enzyme` to light-weighted `jest`, `jest-dom`, and `react-testing-library`. 

The feature of the `react-testing-library` is to test components on the `js-dom` with different props and states. It's easy to understand without considering the React lifecycle. It can provide more confidence to developers on what the component will look like on the page.

Sometimes, we can test if the expected props transferred to another component with `enzyme`, but these props may not be used/displayed on the DOM due to the inappropriate logic. With `react-testing-library`, we can just focus on the final DOM (page) and identify those logic defects and props mistakes.

With the new test cases, some of the previous bugs are also identified, which are ASC-89 and ASC-91, we may discuss and fix them later.

A shortage of this new `react-testing-library` is a few materials/best-practices here, and some third-party support isn't adequate either. Although taking the time to learn to cope with some issues, there are still several lines not being covered until now. I've created ASC-90 to record these lines. Hopefully, some of you could help me on the Fixup Friday someday.

- [ ] Yes
- [x] No


src/components/Popover/index.jsx
```diff
@@ -44,6 +44,8 @@ class Popover extends React.PureComponent {
 
   state = {
     isPopoverOpen: this.props.isOpen,
+    // isPopoverOpen: false,
+    // isForcedOpen: false,
   };
```
Harris-YWang, line 49: potential an update later, please refer to ASC-91

package.json
```diff
@@ -157,11 +152,13 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.4.4",
+    "babel-jest": "^26.0.1",
     "cropperjs": "^1.5.6",
     "diff-match-patch": "^1.0.4",
     "draft-js": "^0.11.5",
     "draft-js-export-html": "^1.4.1",
     "draft-js-import-html": "^1.4.1",
+    "fs-extra": "^8.1.0",
```
Harris-YWang, line 161: a necessary lib for build, don't know why it's unnecessary previously

package.json
```diff
@@ -157,11 +152,13 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.4.4",
+    "babel-jest": "^26.0.1",
```
bchew, line 155: Does this need to be a dependency or can it be a `devDependency`?

package.json
bchew: Perhaps one of the packages that was removed has that as a dependency previously - as above, can this be a `devDependency` too?

src/components/Accordion/index.spec.jsx
```diff
@@ -1,144 +1,102 @@
...
 import React from 'react';
-import sinon from 'sinon';
+import { render, cleanup, fireEvent } from '@testing-library/react';
+
+import Accordion from '.';
 import PanelMocks from '../Panel/mocks';
 
-describe('AccordionComponent', () => {
+beforeEach(() => {
+  jest.spyOn(console, 'error').mockImplementation(() => {});
```
lightbringer1991, line 9: can use `_.noop` for the mock implementation here.

src/components/ActionPanel/index.jsx
```diff
@@ -22,10 +22,21 @@ const ActionPanel = React.forwardRef((props, ref) => {
   const actionPanel = (
     <div ref={ref}>
       <div className={isModal ? 'aui--action-panel-backdrop' : 'hide'} />
-      <div className={classNames('aui--action-panel-wrapper', { 'aui--action-panel-modal-wrapper': isModal })}>
-        <div className={classNames('aui--action-panel', `is-${size}`, { 'action-modal': isModal }, className)}>
-          <div className={classNames('aui--action-panel-header', { 'has-actions': actionButton })}>
-            <div className="title">{title}</div>
+      <div
+        data-testid="action-panel-modal-wrapper"
```
lightbringer1991, line 26: this is definitely interesting. I didn't know about the `data-testid` stuff. 
recommend a read for everyone https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change

Harris-YWang:
> Can you also check on this error in the output: https://travis-ci.org/github/Adslot/adslot-ui/builds/716813863#L976-L989
> 
> Might be something else to add to our warning/error check list too

Solved

vinteo:
Merging...

PR #1033, opened 2020-06-15
feat: toastify component

devanshiAdslot:
https://adslot.atlassian.net/browse/ASC-70

## Description
Add react-toastify to adslot-ui.

- [ ] Yes
- [x] No

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16189,6 +16189,56 @@
         "scheduler": "^0.13.6"
       }
     },
+    "react-toastify": {
+      "version": "6.0.5",
+      "resolved": "https://registry.npmjs.org/react-toastify/-/react-toastify-6.0.5.tgz",
+      "integrity": "sha512-1YXSb6Jr478c1TJEyVpxLHFvtmeXGMvdpZc0fke/7lK+MoLBC+NFgB74bq+C2SZe6LdK+K1voEURJoY88WqWvA==",
+      "requires": {
+        "classnames": "^2.2.6",
+        "prop-types": "^15.7.2",
+        "react-transition-group": "^4.4.1"
+      },
+      "dependencies": {
+        "@babel/runtime": {
+          "version": "7.10.2",
+          "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.10.2.tgz",
+          "integrity": "sha512-6sF3uQw2ivImfVIl62RZ7MXhO2tap69WeWK57vAaimT6AZbE4FbqjdEJIN1UqoD6wI6B+1n9UiagafH1sxjOtg==",
+          "requires": {
+            "regenerator-runtime": "^0.13.4"
+          }
+        },
+        "csstype": {
+          "version": "2.6.10",
+          "resolved": "https://registry.npmjs.org/csstype/-/csstype-2.6.10.tgz",
+          "integrity": "sha512-D34BqZU4cIlMCY93rZHbrq9pjTAQJ3U8S8rfBqjwHxkGPThWFjzZDQpgMJY0QViLxth6ZKYiwFBo14RdN44U/w=="
+        },
+        "dom-helpers": {
+          "version": "5.1.4",
+          "resolved": "https://registry.npmjs.org/dom-helpers/-/dom-helpers-5.1.4.tgz",
+          "integrity": "sha512-TjMyeVUvNEnOnhzs6uAn9Ya47GmMo3qq7m+Lr/3ON0Rs5kHvb8I+SQYjLUSYn7qhEm0QjW0yrBkvz9yOrwwz1A==",
+          "requires": {
+            "@babel/runtime": "^7.8.7",
+            "csstype": "^2.6.7"
+          }
+        },
+        "react-transition-group": {
+          "version": "4.4.1",
+          "resolved": "https://registry.npmjs.org/react-transition-group/-/react-transition-group-4.4.1.tgz",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -58,7 +58,7 @@
   ],
   "husky": {
     "hooks": {
-      "pre-commit": "npm run protect-master && lint-staged",
```
Harris-YWang, line 61: will consider the solution with @glenkurniawan-adslot later

package-lock.json
```diff
@@ -16189,6 +16189,56 @@
         "scheduler": "^0.13.6"
       }
     },
+    "react-toastify": {
+      "version": "6.0.5",
+      "resolved": "https://nexus.corp.adslot.com/repository/adslot-npm/react-toastify/-/react-toastify-6.0.5.tgz",
```
bchew, line 16194: The default npm registry (https://registry.npmjs.org) should be used

package.json
bchew: Technically this is a legacy requirement when Github did not have branch protection. Since we do have it on master now, we probably don't need it anymore so we could drop it.

The script itself can be removed too if it's no longer going to be used

src/components/Toast/ToastNotification.jsx
```diff
@@ -0,0 +1,55 @@
...
+import { toastPlacements } from './constants';
+import 'react-toastify/dist/ReactToastify.css';
+import './styles.scss';
+
+const ToastNotificationComponent = props => {
+  const { theme, title, message } = props;
+  const toastClass = `toast-type-${theme}`;
+  const toastMessage = (
+    <React.Fragment>
+      <table className="toast-body-message">
```
vinteo, line 14: just wondering if a table layout is needed. can it be done using `div`s and/or css flex

src/components/Toast/ToastContainer.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ToastContainer } from 'react-toastify';
+import { toastPlacements } from './constants';
+import './styles.scss';
+import 'react-toastify/dist/ReactToastify.css';
```
veegandhi, line 6: Css and scss should be the other way around. I think the order will mater when override is needed

src/components/Toast/ToastNotification.jsx
```diff
@@ -0,0 +1,55 @@
...
+            <td>{message}</td>
+          </tr>
+        </tbody>
+      </table>
+    </React.Fragment>
+  );
+
+  const options = _.omit({ className: 'toast-notification-body', ...props }, ['title', 'theme', 'message']);
+
+  const toastList = new Set();
```
veegandhi, line 29: Is this `toastList` redundant? It would be better to return l#30 to the user incase toastId is needed to progamatically destroy the toast

src/components/Toast/styles.scss
```diff
@@ -0,0 +1,57 @@
+@import '~styles/color';
+
+.toast-container-component {
+  width: auto;
```
veegandhi, line 4: If the width is `auto` and `max-width` is applied if the content overflows it ui with look glitchy.

src/components/Toast/styles.scss
```diff
@@ -0,0 +1,57 @@
...
+    margin-top: 18px;
+    padding: 0;
+
+    .toast-body-message {
+      font-size: 14px;
+      color: $color-white;
+      margin: 18px;
+
+      .toast-type-info {
+        font-weight: 700;
```
veegandhi, line 21: `font-weight` `font-size` and `margin-right` should be abstracted to one base class and then only colour should be applied per class.

src/components/Toast/ToastContainer.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+const ToastContainerComponent = props => <ToastContainer className="toast-container-component" {...props} />;
```
veegandhi, line 8: classNames within Adslot ui should be prefixed with `aui--` so we avoid classes within usage. So in this instance you'll have `aui--toast-container`. Drop the `-component` as its self explanatory.

www/examples/Toast.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+<DesignNotes>
+  <div>
+    <p>
+      Toasts are used to provide action-based feedback. The Toast.Container can be defined at the highest app layer,
+      while the Toast.Notification can be used multiple times to create multiple notifications. The themes show
+      different colours like:
+    </p>
+    <p>
+      <span className="text-green">Success</span> : indicates that an action was processed successfully.
```
veegandhi, line 33: I think the explanation associated with each class name is use case specific. the part after `:` should be dropped.

www/examples/Toast.mdx
```diff
@@ -0,0 +1,87 @@
...
+      <br />
+      <span className="text-red">Attention</span> : indicates that an action failed completely or something is
+      critically wrong.
+      <br />
+      <span className="text-cyan">Info</span> : provides information related to a performed action.
+    </p>
+  </div>
+
+  <div style={{ paddingTop: 5 }}>
+    See <a href="https://fkhadra.github.io/react-toastify">React Toastify Documentation</a>
```
veegandhi, line 45: See React Toastify Documentation **_for more info_**

src/components/Toast/ToastContainer.jsx
```diff
@@ -0,0 +1,42 @@
...
+  autoClose: 5000,
+  hideProgressBar: true,
+  newestOnTop: false,
+  closeOnClick: true,
+  rtl: false,
+  draggable: true,
+  pauseOnHover: true,
+};
+
+export default ToastContainerComponent;
```
veegandhi, line 42: I am inclining towards `React.memo` (memoizing) this component. As all props are primitive data types and if this component is at the top level under redux provider it can rerender redundantly depending on the implementation.

PR #1037, opened 2020-07-02
[ASC-77][ASC-78] feat: tile and pill component

Harris-YWang:
## Description

We discussed `<Pill />` and `<Tile />` components in the UI/Grid meeting and made a plan about what props should be used this morning. 

As instructed, I made new `<Pill />` and `<Tile />` components, and refactor `<StatusPill />` with new `<Pill />`.

Now, `<Tile />` means a single generic small tile, can be used in `<Carousel />`. 

No breaking changes introduced to `<StatusPill />`.

Screen Shot 2020-07-02 at 2 02 48 pm

Screen Shot 2020-07-02 at 2 02 33 pm

- [ ] Yes
- [x] No


src/components/Pill/styles.scss
```diff
@@ -0,0 +1,32 @@
...
+  justify-content: center;
+  padding: 5px 12px;
+  border: 1px solid transparent;
+  border-radius: 12px;
+  user-select: none;
+  font-weight: $font-weight-bold;
+  line-height: 1;
+  color: $color-gray-darker;
+  background-color: $color-background;
+  border-color: $color-status-light;
```
lightbringer1991, line 16: can merge this with the `border` above?
`border: 1px solid $color-status-light`

src/components/Tile/index.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import { expandDts } from '../../lib/utils';
+import './styles.scss';
+
+const Tile = ({ className, title, logo, onClick, dts }) => {
```
lightbringer1991, line 7: instead of `logo`, I think we can change to `image` or `backgroundImage` to make it generic.
The tile essentially is an image with some text going on top right?

src/components/Tile/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+      <div className={`${baseClass}-title`}>{title}</div>
+    </div>
+  );
+};
+
+Tile.propTypes = {
+  /**
+   *  	Custom classnames
+   */
+  className: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
```
lightbringer1991, line 23: maybe just a string?
in which case are we using an array of strings for class names?

src/components/Pill/index.spec.jsx
```diff
@@ -0,0 +1,41 @@
...
+import Pill from '.';
+
+describe('<Pill />', () => {
+  let sandbox;
+
+  before(() => {
+    sandbox = sinon.createSandbox();
+  });
+
+  afterEach(() => sandbox.restore());
```
lightbringer1991, line 13: seems like `sandbox` is not in use, we can remove.

src/components/Tile/index.jsx
Harris-YWang: It depends, If the tile is given a logo, no text would be put on the top of it for our new cases. But our generic `<Tile /> component could support the use case you mentioned. As `<TileGrid />` is using `ImgLink` as the prop name, would you like to use it to keep this consistency, or which one else is preferred? cc: @lteacher @superpower1

src/components/Tile/index.spec.jsx
```diff
@@ -0,0 +1,61 @@
...
+  );
+
+describe('<Tile />', () => {
+  let sandbox;
+
+  before(() => {
+    sandbox = sinon.createSandbox();
+  });
+
+  afterEach(() => sandbox.restore());
```
lightbringer1991, line 22: unused `sandbox`

src/components/Tile/index.jsx
Harris-YWang: sure, it makes sense

src/components/Pill/styles.scss
```diff
@@ -0,0 +1,32 @@
+@import '~styles/variable';
+
+.aui--pill {
+  font-size: $font-size-base;
+  display: inline-block;
+  align-items: center;
+  justify-content: center;
```
superpower1, line 7: This two lines are not needed as your display is not flex.

Harris-YWang: > The props table is missing for Tile and Pill.

This one is because currently the `react-docgen` no longer generates props from files due to the #ASC-75 issue, would be fixed once that issue is resolved

PR #1038, opened 2020-07-03
Fix docgen failure if a single component cant be parsed

lteacher:
## Description

There is an issue with `react-docgen` where it can't detect some types of components. When that happens it throws and error which is then a promise rejection and can impact the prop doc generation. This fix adds a try catch to handle that with a warning for now as it doesn't seem that the parse will be detecting these components unless there are some magical fixes.

I added a comment to reference the tracked issue and simply log a warning message

- [ ] Yes
- [X] No

### DIFF
--- a/www/containers/props.json
+++ b/www/containers/props.json
@@ -4518,11 +4518,111 @@
       }
     }
   ],
-  "src/components/Toast/index.jsx": [
+  "src/components/Toast/ToastContainer.jsx": [
     {
       "description": "",
-      "displayName": "toastContainerComponent",
-      "methods": []
+      "displayName": "ToastContainerComponent",
+      "methods": [],
+      "props": {
+        "position": {
+          "type": {
+            "name": "enum",
+            "computed": true,
+            "value": "toastPlacements"
+          },
+          "required": false,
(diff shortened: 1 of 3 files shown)


www/containers/props.json
```diff
@@ -1860,7 +1860,6 @@
   "src/components/ImageCropper/index.jsx": [
     {
       "description": "",
-      "displayName": "ImageCropper",
```
lteacher, line 1863: Hmm not too sure why this is changed let me know if you have any clues

lteacher:
At first I ignored the Toast related stuff but not sure if the issue will be resolved so changed to just warn for now. Am thinking we can re-review later and ignore relevant components if necessary

www/containers/props.json
Harris-YWang: Seems this line was added manually. If we want to get this line back, just add `ImageCropper.displayName = 'ImageCropper';` to `components/ImageCropper/index.jsx`

superpower1: It doesn't seems to be added manually by looking gat the git history. Looking at `Carousel` component that doesn't have display name in the component but docgen also gave it a display name for some reason. @Harris-YWang

PR #1042, opened 2020-07-16
refactor: move custom class to the end

lightbringer1991:
Move all custom classes to the end of `className`

## Description

CSS rules are applied from left to right, so the later styles will override the former. Therefore to give the custom class the most control, we need to put it at the end of class generation.

- [ ] Yes
- [x] No


vinteo:
Merging...

PR #1043, opened 2020-07-17
fix: toast invalid hook issue

Harris-YWang:
## Description

Because the incompatibility between React.memo and Symphony's project and `<ToastContainer />` has been fully wrapped by `react-toastify` like a black-box, we think we don't have to wrap it again in adslot-ui for the moment.

Also, `react-toastify` use `toast()` to trigger a toast adding to the `<ToastContainer />`. Initially we wrapped it into a component, which should works well in JSX files. However, with Symphony's project, they are gonna use this function in the redux action, which is unclear for people to manage a component in the action.js file. So, adding a `Toast.renderNotification()` function, exactly has the same UI display as `<Toast.Notificaiton />`, but we can use it as a function rather than a functional component.

btw, fix small bugs of the doc.

- [ ] Yes
- [x] No

## Manual testing step?

Manual tested by @devanshiAdslot


src/components/Toast/ToastContainer.jsx
```diff
@@ -39,4 +39,4 @@ ToastContainerComponent.defaultProps = {
   pauseOnHover: true,
 };
 
-export default React.memo(ToastContainerComponent);
```
lightbringer1991, line 42: yea I think we shouldn't wrap anything in `adslot-ui`, users can do it themselves.

src/components/Toast/ToastNotification.jsx
```diff
@@ -1,29 +1,43 @@
...
+  const toastClass = getToastClass(theme);
+  const toastMessage = <ToastMessage toastClass={toastClass} title={title} message={message} />;
+  const toastOptions = options(props);
 
-  toast(toastMessage, options);
+  toast.info(toastMessage, toastOptions);
   return null;
 };
 
+export const renderNotification = ({ theme, title, message, ...props }) => {
```
knilink, line 34: maybe use a different name, such as just `info`?
as it's not a typical render function

Harris-YWang: how about something like `Toast.notify`?

superpower1:
Please don't commit your build files.

Harris-YWang:
> Please don't commit your build files.

Solved

PR #1049, opened 2020-07-30
refactor: add isSaving prop to image cropper

superpower1:
## Description

Add `isSaving` prop to ImageCropper component to let user to control whether it's saving. We used to handle `isSaving` internally by awaiting the `onCrop` function which will be an issue when this component is unmount before the onCrop function finish. So I think it's better to let user control whether it is saving.   

- [x] Yes
- [ ] No


pphminions:
merging 🚀

PR #1051, opened 2020-07-31
fix: add ref to carousel component

superpower1:
## Description

In order to use methods on react-slick, user has to give ref to Slider component. So we have to forward ref to Slider.
[react-slick doc](https://react-slick.neostack.com/docs)

For example:
```jsx
const example = () =>  {
  const images = [0,1,2];
  const carouselRef = React.useRef();
  const goToSlide = () => carouselRef.current.slickGoTo(images.length -1)
    return (
      <div>
        <button onClick={goToSlide}>Go To Last Slide</button>
        <Slider ref={carouselRef} {...settings} />
        {images.map((image, index) => (
          <div>
            <img src={baseUrl + "/abstract" + index + ".jpg"} />
          </div>
        ))}   
      </div>
    )
}
```

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/Carousel/index.jsx
+++ b/src/components/Carousel/index.jsx
@@ -8,15 +8,15 @@ import './styles.scss';
 
 const baseClass = 'aui--carousel-component';
 
-const CarouselComponent = props => {
+const CarouselComponent = React.forwardRef((props, ref) => {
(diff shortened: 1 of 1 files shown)


vinteo:
Merging...

PR #1054, opened 2020-08-06
feat: add prop collapsable

Harris-YWang:
## Description

Add a new prop `collapsable` to the component `<VerticalNav />` to enable users to control if the VerticalNav menu could be foldable into a burger icon.

- [ ] Yes
- [x] No

Screen Shot 2020-08-06 at 2 20 40 pm


Harris-YWang:
> The thing is that the prop word looked wrong to me lol... Unless im spazzing out completely I think its American English? usually we would have `collapsable`... Its a combination of collapse + able which makes perfect sense to me. Anyway i think we should change it haha although usually things like `color` im ok with

lol, will change it to `collapsable`

PR #1059, opened 2020-08-17
feat: add ref to Search

Harris-YWang:
## Description

- Add `ref` to `<Search />` component, so that we can give search input a focus if required
- Add missing coverage to `<Search />`

- [ ] Yes
- [X] No

### DIFF
--- a/www/containers/props.json
+++ b/www/containers/props.json
@@ -3517,7 +3517,6 @@
   "src/components/Search/index.jsx": [
     {
       "description": "",
-      "displayName": "Search",
       "methods": [],
       "props": {
         "className": {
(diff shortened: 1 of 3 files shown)


www/containers/props.json
```diff
@@ -3517,7 +3517,6 @@
   "src/components/Search/index.jsx": [
     {
       "description": "",
-      "displayName": "Search",
```
superpower1, line 3520: Just wondering why this happens for every component we change although it doesn't harm.

Harris-YWang: A good question, maybe let's create an ASC card to investigate it on Fix-up First Friday

pphminions:
merging 🚀

PR #1070, opened 2020-09-07
refactor: nav component

Harris-YWang:
## Description

Refactor `<Nav />` as we determine to remove `react-bootstrap` step by step.

Refer to the prop provided and used in the sample. This refactor should not introduce any breaking change, neither to the feature nor to the UI. (perhaps require a double check). 

- [ ] Yes
- [X] No

### DIFF
--- a/src/components/Navigation/index.jsx
+++ b/src/components/Navigation/index.jsx
@@ -1,36 +1,106 @@
-/* eslint-disable react/prop-types */
 import _ from 'lodash';
-import PropTypes from 'prop-types';
 import React from 'react';
-import Nav from 'react-bootstrap/lib/Nav';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
 import { expandDts } from '../../lib/utils';
 import './styles.scss';
 
-const NavigationComponent = props => {
-  const { barPosition, children, dts } = props;
-  const classes = `${barPosition}-bar`;
-  const navProps = _.omit(props, ['barPosition']);
+const Nav = ({ stacked, className, onSelect, activeKey, barPosition, children, dts }) => {
+  let navItems = [];
+  const [currentKey, setActiveKey] = React.useState(activeKey);
+
+  React.Children.forEach(children, child => {
+    navItems.push(React.cloneElement(child, { onSelect, activeKey: currentKey, setActiveKey }));
+  });
 
   return (
-    <Nav
-      {...navProps}
-      {...expandDts(dts)}
-      data-testid="navigation-wrapper"
-      bsClass="nav-borderless"
-      className={classes}
-    >
-      {children}
-    </Nav>
+    <nav data-testid="nav-component-wrapper" {...expandDts(dts)}>
+      <ul
+        data-testid="nav-component"
+        className={classnames('aui--nav', 'nav-borderless', `${barPosition}-bar`, { stacked: stacked }, className)}
+      >
+        {navItems}
+      </ul>
+    </nav>
   );
 };
 
-NavigationComponent.propTypes = _.assign({}, Nav.propTypes, {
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -48,6 +47,7 @@ import HelpIconPopover from './components/HelpIconPopover';
 import ListPicker from './components/ListPicker';
 import ListPickerPure from './components/ListPickerPure';
 import Nav from './components/Navigation';
+import { NavItem } from './components/Navigation';
```
Harris-YWang, line 50: This solution may solve the docgen issue of the `<Toast />`

src/components/Navigation/index.jsx
```diff
@@ -1,36 +1,77 @@
...
-      className={classes}
+    <li
+      data-testid="nav-item-component"
+      role="presentation"
+      className={classnames('aui--nav-item', className, { active: activeKey === eventKey }, { disabled: disabled })}
+      onClick={disabled ? _.noop : onItemClick}
     >
-      {children}
-    </Nav>
+      {href ? (
```
Harris-YWang, line 40: if given a `href`, then put the `<a>` element here, otherwise leave it because I'm thinking we may use `<Link />` of the `react-router` inside the NavItem later

src/components/Navigation/index.jsx
```diff
@@ -1,36 +1,77 @@
...
+        {navItems}
+      </ul>
+    </nav>
+  );
+};
+
+export const NavItem = ({ className, disabled, activeKey, setActiveKey, eventKey, href, onSelect, children }) => {
+  const onItemClick = () => {
+    setActiveKey(eventKey);
+    onSelect(eventKey);
```
Harris-YWang, line 30: use original naming prop 'onSelect', but `onselect` event is supposed to be used for `<input />` only. please correct me if I'm wrong

vinteo (approved):

src/components/Navigation/index.jsx
```diff
@@ -1,36 +1,104 @@
...
+Nav.propTypes = {
+  /**
+   * NavItems are be positioned horizontally or vertically.
+   */
+  stacked: PropTypes.bool,
+  className: PropTypes.string,
+  /**
+   * A callback fired when any children is selected.
+   */
+  onSelect: PropTypes.func,
```
lightbringer1991, line 61: maybe add a sample of how to use this function, e.g. `onSelect(eventKey)` so that it's easier for people to implement when looking at the doc.

src/components/Navigation/index.jsx
```diff
@@ -1,36 +1,104 @@
...
-    <Nav
-      {...navProps}
-      {...expandDts(dts)}
-      data-testid="navigation-wrapper"
-      bsClass="nav-borderless"
-      className={classes}
+    <li
+      data-testid="nav-item-component"
+      role="presentation"
+      className={classnames('aui--nav-item', className, { active: activeKey === eventKey }, { disabled: disabled })}
```
lightbringer1991, line 38: I reckon the custom class name should be at the end, so it can override all previous styles.

lightbringer1991: also you can combine the object `{ active: activeKey === eventKey, disabled }`

Harris-YWang: Got it, thanks

src/components/Navigation/index.jsx
lightbringer1991: yea I feel like for NavItem it should be `onClick`, and for `Nav` it's `onChange`. But happy to go with the previous naming.

superpower1 (approved):

lteacher (approved):

lightbringer1991 (approved):
LGTM

PR #1072, opened 2020-09-09
refactor: file picker component

superpower1:
## Description
1. Refactor File Picker to functional component
1. Add `onClick` prop for click event on file name
1. Add `onChange` prop for change event on file name
1. Add `value` prop to make the file name controllable

- [ ] Yes
- [ ] No


src/components/FilePicker/index.jsx
```diff
@@ -1,129 +1,169 @@
...
+      <input
+        data-testid="file-picker-form-control"
+        className={classNames('form-control', { clickable: _.isFunction(onClick) })}
+        type="text"
+        readOnly
+        placeholder={placeholder}
+        value={value || fileName}
+        title={value || fileName}
+        onClick={clickFile}
+      />
```
lightbringer1991, line 86: @superpower1 @minhe86 can you put `dts` in this input as well, thank you.

lightbringer1991: something like `data-test-selector={`${dts}-form-control`}`, if `dts` exists.

Harris-YWang (commented):
Screen Shot 2020-09-09 at 4 20 19 pm
the borders of input and button are not aligned, is it what we want? lol

Harris-YWang (commented):
Perhaps add a new example to doc, and when hovering on the file, if we should give an underline to the file name?

Harris-YWang (approved):
LGTM

src/components/FilePicker/index.jsx
superpower1: I will put dts to the wrapper div instead

vinteo (approved):

superpower1:
> Screen Shot 2020-09-09 at 4 20 19 pm
> the borders of input and button are not aligned, is it what we want? lol

Fixed.

src/components/RichTextEditor/index.jsx
```diff
@@ -67,6 +67,10 @@ RichTextEditor.propTypes = {
   /** Editor State: Instance of <a href="https://draftjs.org/docs/api-reference-editor-state">draft-js editor state</a> */
   value: PropTypes.instanceOf(EditorState),
   onChange: PropTypes.func,
+  /**
+   * data-test-selector of the rich text editor
+   */
+  dts: PropTypes.string,
```
superpower1, line 73: Sneak in a change to add `dts` to RichTextEditor.

Harris-YWang (approved):

src/components/FilePicker/index.jsx
```diff
@@ -1,129 +1,169 @@
...
+   */
+  onSelect: PropTypes.func.isRequired,
+  /**
+   * function called when user click the file name
+   */
+  onClick: PropTypes.func,
+  /**
+   * function called when the file name changes
+   */
+  onChange: PropTypes.func,
```
adityasabnis, line 151: I can't exactly recollect the discussion, however, didn't we decide to cut down the handler functions.
do we need `onSelect, onClick, onChange, onRemove` for this picker

lteacher: I think we were planning to retain the `onSelect` for scenarios where the value is handled by the component

src/components/FilePicker/index.jsx
```diff
@@ -1,129 +1,169 @@
...
-  state = {
-    isFileSelected: false,
-    fileName: '',
-  };
+  const selectFile = event => {
+    if (!isFileSelected) {
+      const file = event.target.files[0];
+      setIsFileSelected(true);
+      if (_.isFunction(onChange)) {
+        onChange(file.name);
```
lteacher, line 32: I think we should probably provide the actual file for the on change instead of just the name and allow the external component to put the file's name into the value. In those scenarios they would also not need to use `onSelect` since they have the power to do whatever and they also have the file itself to manage rather than it being hidden in the component.

lightbringer1991 (approved):
LGTM

src/components/FilePicker/index.jsx
```diff
@@ -1,129 +1,169 @@
...
+    if (isFileSelected || !_.isEmpty(value)) {
+      fileInput.current.value = null;
+      setIsFileSelected(false);
+      if (_.isFunction(onChange)) {
+        onChange('');
+      }
+      setFileName('');
+
+      if (_.isFunction(onRemove)) {
+        onRemove();
```
knilink, line 54: Not quite sure about this. Seem to be a little bit confusing to me triggering multiple handlers.
IMO should only keep either `onChange('')` or `onRemove()`, otherwise there will be multiple ways of doing the samething.

should be either
```js
// handle remove with onChange('')
handleChange = (value) => {
  if(value==='') {/* handle remove */}
}
```
or 
```js
// setValue('') with onRemove()
handleRemove = () => setValue('')
```

PR #1085, opened 2020-10-09
fix: fix ui button

glenkurniawan-adslot:
## Description

After testing with local on grid and direct web, I found that the button that we have on both of them has additional `btn` className.. Haofu suggested that `btn` className comes from `react-bootstrap`, since we want to remove bootstrap dependency, I refactored any `btn` className as well as components that uses `react-bootstrap Button`.

- [ ] Yes
- [x] No


Harris-YWang (approved):
LGTM

src/components/ConfirmModal/__snapshots__/index.spec.jsx.snap
```diff
@@ -28,12 +28,13 @@ exports[`<ConfirmModal /> should show modal when \`show\` is true 1`] = `
...
       >
         <button
-          class="btn btn-primary"
+          class="aui--button btn-primary"
           data-test-selector="confirm-modal-confirm"
           data-testid="confirm-modal-confirm"
-          type="button"
         >
-          Confirm
+          <div>
```
superpower1, line 35: Wondering why there is a div wrapper here?

glenkurniawan-adslot: @superpower1 because of this: 
```
<button
      data-testid="button-wrapper"
      disabled={isLoading || disabled}
      className={classes}
      {...expandDts(dts)}
      {..._.omit(props, _.keys(adslotButtonPropTypes))}
    >
      {renderSpinner()}
      **<div className={isLoading ? 'aui--button-children-container' : null}>{children}</div>**
    </button>
```

superpower1 (approved):

Harris-YWang (changes requested):
Seems you've made some changes after my last review, and some stylings are broken again. please double check the style of each button per https://ui.adslot.com/button  Thanks mate.

Also, here is a console.error from the unit tests as follows, and it'd better to fix:
Screen Shot 2020-10-16 at 7 55 01 am

glenkurniawan-adslot:
> Seems you've made some changes after my last review, and some stylings are broken again. please double check the style of each button per https://ui.adslot.com/button Thanks mate.

@Harris-YWang Yeah, Haofu suggested to remove the `btn` className, instead of adding it (because btn is coming from bootstrap.. I will work on the fix.. thanks

Harris-YWang (approved):
Compared with the existing button design, everything looks good to me now. :+1:

lightbringer1991:
seems like my PR is duplicated https://github.com/Adslot/adslot-ui/pull/1089
will close my PR and track this one instead

src/components/Button/index.jsx
```diff
@@ -9,11 +9,11 @@ import { expandDts } from '../../lib/utils';
 import './styles.scss';
 
 const Button = props => {
-  const { bsStyle, inverse, size, children, dts, className, isLoading, disabled } = props;
+  const { bsStyle, bsSize, inverse, size, children, dts, className, isLoading, disabled } = props;
   const baseClass = 'aui--button';
   const classes = classNames(baseClass, className, {
```
lightbringer1991, line 14: I think we need to put `className` at the end of the class chain so the custom styles can override previous classes if needed.

src/components/ConfirmModal/__snapshots__/index.spec.jsx.snap
```diff
@@ -28,12 +28,13 @@ exports[`<ConfirmModal /> should show modal when \`show\` is true 1`] = `
         data-testid="confirm-modal-footer"
       >
         <button
-          class="btn btn-primary"
+          class="aui--button btn-primary"
```
lightbringer1991, line 31: no need to add `aui--button` here, the component will generate it automatically

src/components/Button/index.jsx
glenkurniawan-adslot: ok.. sounds good... will update this

src/components/TreePicker/Node/index.jsx
```diff
@@ -69,9 +69,7 @@ class TreePickerNodeComponent extends React.PureComponent {
           {selected ? (
             <GridCell classSuffixes={['button']} dts="button-remove">
               <Button
-                block
-                bsSize="xsmall"
-                className="btn-inverse"
+                className="btn-inverse button-xs"
```
lightbringer1991, line 72: I think we should use `inverse` prop instead of declaring class here.
```
<Button inverse className="button-xs" {...props}>....</Button>
```

glenkurniawan-adslot: they are basically doing the same thing.. so I think either is fine.. ex: we can use btn-large or bsSize="large"

src/components/ConfirmModal/__snapshots__/index.spec.jsx.snap
glenkurniawan-adslot: @lightbringer1991 ok.. will update this

src/components/Button/index.jsx
```diff
@@ -9,22 +9,26 @@ import { expandDts } from '../../lib/utils';
...
-  });
+  const classes = classNames(
+    baseClass,
+    {
+      'btn-inverse': inverse && !/btn-inverse/.test(className),
+      'btn-large': size === 'large' || _.includes(['lg', 'large'], bsSize),
+      'aui--btn-default':
+        (!bsStyle || bsStyle === 'default') &&
+        (!className ||
+          ['btn-default', 'btn-inverse', 'btn-default btn-inverse', 'btn-inverse btn-default'].includes(className)),
```
lightbringer1991, line 22: I think `btn-default` no longer exists, it's now `aui--btn-default`, so we can simplify the conditions here a bit.

I guess it would be breaking for the following suggestions, just throwing it out here. We drop all checking conditions for class names, make the user responsible for what class they want to put in, and force them to use props where possible. E.g.
```
const classes = classNames(
  baseClass,
  {
    'btn-inverse': inverse,
    'aui--btn-default': !bsStyle,
    'btn-large': size === 'large',
    [`btn-${bsStyle}`]: !_.isEmpty(bsStyle),
  },
  className
);
```

lightbringer1991 (approved):
LGTM

src/components/ConfirmModal/__snapshots__/index.spec.jsx.snap
glenkurniawan-adslot: @lightbringer1991 as mentioned, this is needed because the snapshot breaks the <Button> to <button><div></div></button>

lteacher (approved):

PR #1089, opened 2020-10-18
fix: fix custom class name for button component

lightbringer1991:
- Refactor class names for button component.
- Put the custom class at the end of the class chain to ensure all custom styles are applied.
- Change internal Button usage in other components to conform to the prop structure.

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/Button/index.jsx
+++ b/src/components/Button/index.jsx
@@ -11,15 +11,17 @@ import './styles.scss';
 const Button = props => {
   const { bsStyle, inverse, size, children, dts, className, isLoading, disabled } = props;
   const baseClass = 'aui--button';
-  const classes = classNames(baseClass, className, {
-    'btn-inverse': inverse && !/btn-inverse/.test(className),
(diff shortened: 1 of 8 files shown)


src/components/Button/index.jsx
```diff
@@ -11,15 +11,17 @@ import './styles.scss';
 const Button = props => {
   const { bsStyle, inverse, size, children, dts, className, isLoading, disabled } = props;
   const baseClass = 'aui--button';
-  const classes = classNames(baseClass, className, {
-    'btn-inverse': inverse && !/btn-inverse/.test(className),
-    'btn-large': size === 'large',
-    'aui--btn-default':
-      (!bsStyle || bsStyle === 'default') &&
-      (!className ||
-        ['btn-default', 'btn-inverse', 'btn-default btn-inverse', 'btn-inverse btn-default'].includes(className)),
```
lightbringer1991, line 20: `btn-default` is now non-existent.

PR #1090, opened 2020-10-19
fix: fix carousel incorrect prop

lightbringer1991:
Fix Carousel incorrect prop

## Description

According to the API https://react-slick.neostack.com/docs/example/auto-play, there is no `autoplayInterval`, but only `autoplaySpeed`, which is the prop that we need.

- [ ] Yes
- [x] No


Harris-YWang (approved):
LGTM

pphminions (approved):
LGTM 🟠

pphminions:
merging 🚀

PR #1097, opened 2020-11-16
feat: enable button to support href

Harris-YWang:
## Description

With `react-bootstrap` `<Button />` component, it can work as an anchor with the prop `href`. In our main project, this `href` is frequently used now. 

Bring this function back to better support.

- [ ] Yes
- [X] No

### DIFF
--- a/src/components/Button/index.jsx
+++ b/src/components/Button/index.jsx
@@ -9,7 +9,7 @@ import { expandDts } from '../../lib/utils';
 import './styles.scss';
 
 const Button = props => {
-  const { bsStyle, bsSize, inverse, size, children, dts, className, isLoading, disabled } = props;
+  const { bsSize, bsStyle, children, className, disabled, dts, href, inverse, isLoading, size } = props;
   const baseClass = 'aui--button';
   const classes = classNames(
     baseClass,
@@ -32,6 +32,15 @@ const Button = props => {
       </div>
     ) : null;
 
+  const renderChildren = () =>
+    href ? (
+      <a data-testid="button-anchor" className="aui--button-anchor" href={href}>
+        {children}
+      </a>
+    ) : (
+      children
+    );
+
   return (
     <button
       data-testid="button-wrapper"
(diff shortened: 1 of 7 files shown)


vinteo (approved):
Just making sure the clickable area will be the whole button as there might be some padding?

src/components/Button/index.jsx
```diff
@@ -9,7 +9,7 @@ import { expandDts } from '../../lib/utils';
 import './styles.scss';
 
 const Button = props => {
-  const { bsStyle, bsSize, inverse, size, children, dts, className, isLoading, disabled } = props;
+  const { bsSize, bsStyle, children, className, disabled, dts, href, inverse, isLoading, size } = props;
```
superpower1, line 12: Just notice that we have `bsSize` and `size`, is it possible to combine them together to just use `size`? As we are not using bootstrap button any more, maybe we should consider removing `bsStyle`? Just something I noticed not related to your changes of this pr.

superpower1 (approved):

lteacher (approved):

src/components/Button/index.jsx
Harris-YWang: @glenkurniawan-adslot hey mate, what do you think of? Since you added both props here, I feel like both ways are accepted

glenkurniawan-adslot: @Harris-YWang @superpower1 the overlap between size and bsSize should be fixed when we clean up the Button (we have a card for this, and it would be a breaking changes). And yes, I believe we should be removing bsStyle as it is bootstrap styling...

glenkurniawan-adslot (approved):

src/components/Button/index.jsx
Harris-YWang: @superpower1 let's remove 'bs' props with another pr then

Harris-YWang:
> Just making sure the clickable area will be the whole button as there might be some padding?

Checked and fixed

knilink:
maybe add support for prop `target` as well?

lightbringer1991:
hmm don't we already have `bsStyle="link"`? Which generates something that looks like a link.
How is this one different from the one we already have?

Harris-YWang:
> hmm don't we already have `bsStyle="link"`? Which generates something that looks like a link.
> How is this one different from the one we already have?

As we removed react-bootstrap, and `bsStyle` changes the ui only , we need a `href` to implement the function. In our main project, the `redirect` function is not limited to a `bsStyle='link'`

Harris-YWang:
> maybe add support for prop `target` as well?

A new ASC card is created for this idea, let's discuss this one in the next meeting. https://adslot.atlassian.net/browse/ASC-107

PR #1100, opened 2020-11-20
[ASC-105] feat: new paragraph component

Harris-YWang:
## Description

- New component `<Paragraph />` with a Read More button, and it can be used to display `description` in data
  - It supports `briefWordCount` to determine the threshold to display the brief or full content.
  - It can display plain text and one rich format text - `HTML` when specified.

- As we still explore the requirement of this component, please leave your ideas here to further improve this one.

- [ ] Yes
- [x] No

Kapture 2020-11-20 at 16 11 30

### DIFF
--- a/src/components/Paragraph/index.jsx
+++ b/src/components/Paragraph/index.jsx
@@ -1,0 +1,82 @@
+import _ from 'lodash';
+import React from 'react';
+import { htmlToText } from 'html-to-text';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import { expandDts } from '../../lib/utils';
+import Button from '../Button';
+import './styles.scss';
+
+const Paragraph = ({ briefWordCount, content, className, dts, isHtml }) => {
+  const [readMore, setReadMore] = React.useState(false);
+  const baseClass = 'aui--paragraph';
+  const paragraph = isHtml ? htmlToText(content, { wordwrap: false }) : content;
+  const paragraphWordCount = _.get(paragraph, 'length');
+
+  const brief = _.truncate(paragraph, {
+    length: briefWordCount,
+    separator: ' ',
+  });
+
+  const toggleReadMore = () => {
+    setReadMore(!readMore);
+  };
+
+  return (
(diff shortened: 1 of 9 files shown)


vinteo (approved):

src/components/Paragraph/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import { expandDts } from '../../lib/utils';
+import Button from '../Button';
+import './styles.scss';
+
+const Paragraph = ({ briefWordCount, content, className, dts, html }) => {
+  const [readMore, setReadMore] = React.useState(false);
+  const baseClass = 'aui--paragraph';
+  const paragraph = html ? htmlToText(content, { wordwrap: false }) : content;
```
superpower1, line 13: Do we really need this flag? Can we always use `htmlToText` for all content?

Harris-YWang: I guess we need this flag, since we may have some formats like `\n` in the plain text, however, `htmlToText` will remove these ones.

And another reason may be sometimes users don't want to convert the content to HTML, although it could be. 

As this one is a root component, we can make to adapt to more use cases. Users can set up this flag if they need it.

src/components/Paragraph/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+   */
+  className: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
+  /**
+   * 	Generate "data-test-selector" on the paragraph
+   */
+  dts: PropTypes.string,
+  /**
+   *    Define if the content is HTML type or not
+   */
+  html: PropTypes.bool,
```
superpower1, line 74: Maybe change it to `isHtml`?

superpower1 (approved):

lteacher (approved):
lgtm

src/components/Paragraph/index.jsx
Harris-YWang: updated to `isHtml`

PR #1104, opened 2020-11-25
[ASC-108][ADS-6053] Upgrade react datepicker

knilink:
- upgraded react datepicker to the latest
- added util to transform momentjs format to unicode format

- [ ] Yes
- [ ] No


src/components/DatePicker/index.jsx
```diff
@@ -2,29 +2,77 @@ import _ from 'lodash';
...
+  React.useEffect(() => {
+    if (ref) ref.current = _ref.current;
+    _datePickerRef(_ref.current);
+    return () => {
+      if (ref) ref.current = null;
+      _datePickerRef(null);
+    };
+  }, [_datePickerRef, ref, _ref]);
+  return React.cloneElement(originalCustomInput, { ...props, [originalCustomInputRef]: _ref });
+});
```
knilink, line 48: I think the reason we fork `react-datepicker` is to get the input ref.
This hack can achieve the same thing.

src/components/DatePicker/index.jsx
```diff
@@ -2,29 +2,77 @@ import _ from 'lodash';
...
+        {...props}
+        selected={momentToDate(selected)}
+        startDate={momentToDate(startDate)}
+        endDate={momentToDate(endDate)}
+        minDate={momentToDate(minDate)}
+        maxDate={momentToDate(maxDate)}
+        onChange={handleChange}
+      />
+    );
+  });
```
knilink, line 35: new `react-datepicker` has dropped the `moment`, here adding a adapter to support it back.

src/components/DatePicker/index.jsx
knilink: another possible way to patch `react-datepicker` without forking is via extending the class
e.g.
``` jsx
class PatchedDatePicker extends ReactDatePicker {
  superRenderDateInput = this.renderDateInput
  renderDateInput = (...args) => {
    return React.cloneElement(this.superRenderDateInput(...args), {
      ....
    });
  }
}
```
not sure whether this is a good practice

dist/adslot-ui-main.css
```diff
@@ -1,5 +1,5 @@
 @charset "UTF-8";
```
Harris-YWang, line 1: Seems we need to remove the built files with this pr

Harris-YWang (commented):
seems also need to fix example files

dist/adslot-ui-main.css
knilink: ah yeah, i rebase this branch against 27.9 because is this the version used in direct web at the moment.
this issue will be gone after rebase back against master

knilink:
as discussed during the retro, would prefer to wait until the release of [temporal](https://github.com/tc39/proposal-temporal) and removing `momentjs` from the main project first.

src/components/DatePicker/index.jsx
```diff
@@ -1,29 +1,89 @@
...
+          maxDate={momentToDate(maxDate)}
+          dateFormat={useMomentFormat ? transform(dateFormat) : dateFormat}
+          onChange={handleChange}
+        />
+      );
+    }
   );
-};
+
+const RefHack = React.forwardRef(({ originalCustomInputRef, originalCustomInput, _datePickerRef, ...props }, ref) => {
```
xiaofan2406, line 36: I dont think we need to support `customInputRef` as the fork does anymore.

As v4 has a complete new API, and has it's own way of customising the input. We can just refactor away any places uses `customInputRef`.

xiaofan2406 (commented):
maybe update the doc page in the same PR as well?

src/components/DatePicker/transformFormat.js
```diff
@@ -0,0 +1,142 @@
+// https://github.com/moment/moment/blob/e96809208c9d1b1bbe22d605e76985770024de42/src/lib/format/format.js#L4
```
vinteo, line 1: why do we need to have a copy of moment code? why can't we use `moment` directly?

knilink: need to make some changes for the following reason
- join local format. moment format local string in a different step and doing it the same way will increase complexity.
https://github.com/Adslot/adslot-ui/blob/b6659ad988b55115aeadec65de6587572bc510fe/src/components/DatePicker/transformFormat.js#L7-L10
- test the token. the original regex will match all string.
https://github.com/Adslot/adslot-ui/blob/b6659ad988b55115aeadec65de6587572bc510fe/src/components/DatePicker/transformFormat.js#L84-L86

package.json
```diff
@@ -89,6 +89,7 @@
     "cross-env": "^7.0.3",
     "css-loader": "^6.5.0",
     "css-minimizer-webpack-plugin": "^3.1.1",
+    "date-fns": "^2.25.0",
```
bchew, line 92: Latest version is now 2.26.0

package.json
```diff
@@ -112,7 +113,7 @@
     "prettier": "^2.4.1",
     "prism-react-renderer": "^1.2.1",
     "prop-types": "^15.7.2",
-    "react-datepicker": "github:Adslot/react-datepicker#2a1ebb8618e1996c321331ef01d290424272d126",
+    "react-datepicker": "^4.3.0",
```
bchew, line 116: Latest version is 4.4.0

bchew (approved):

xiaofan2406 (dismissed):

bchew (approved):
🏆

bchew:
Merging :rocket:

PR #1107, opened 2020-12-08
fix: default button classname bug

Harris-YWang:
## Description

With the last release, the <Button /> default styles will be broken if given a className like 'some class'. This pr is to fix this bug.

- [ ] Yes
- [x] No

Before Change:
Screen Shot 2020-12-08 at 1 55 17 pm

After Change:
Screen Shot 2020-12-08 at 1 56 20 pm

### DIFF
--- a/src/components/Button/styles.scss
+++ b/src/components/Button/styles.scss
@@ -64,7 +64,7 @@
     border-color: $color-button-default-hover-border;
   }
 
-  &.aui--btn-default {
+  &.btn-default {
     background-color: $color-white;
     border-color: $color-gray-light;
     color: $color-gray-dark;
@@ -156,7 +156,8 @@
     }
   }
 
-  &.btn-link {
+  &.btn-link,
+  &.btn-link.btn-default {
     background-color: transparent;
     border-color: transparent;
     color: $color-button-link-color;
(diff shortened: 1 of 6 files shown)


glenkurniawan-adslot (approved):

src/components/Button/styles.scss
```diff
@@ -64,7 +64,8 @@
     border-color: $color-button-default-hover-border;
   }
 
-  &.aui--btn-default {
+  &.aui--btn-default,
+  &.btn-default {
```
superpower1, line 68: If you are going to give the same style for `btn-default` do we still need `aui--btn-default`?

Harris-YWang: Hey mate @glenkurniawan-adslot why did we have the `aui--btn-default` initially?

glenkurniawan-adslot: @Harris-YWang hmm... not quite sure, but for some reason the other class are `btn-xxx` except for `btn-default`, I named it `aui--btn-default` I think it's a mistake on my part.. and it might be better to only use the `btn-default` and remove the `aui--btn-default`, so everything is consistent

vinteo (approved):

src/components/Button/styles.scss
Harris-YWang: Got it, no worries man, I will make further tweak for this one

glenkurniawan-adslot: @Harris-YWang thanks.. and sorry about this..

superpower1 (approved):

adityasabnis (approved):

lteacher (approved):

lightbringer1991 (approved):
LGTM

vinteo:
Merging...

PR #1134, opened 2021-03-18
fix: plugin for dist and popover modifiers

Harris-YWang:
Hot fix for `data-testid` is still in dist files

Fix modifiers not able to be overwritten

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/Popover/Popper.jsx
+++ b/src/components/Popover/Popper.jsx
@@ -19,8 +19,6 @@ export const renderArrowStyles = (placement, arrowStyles, container = null) => {
   const verticalPosition =
     _.get(container, 'clientHeight') >= DEFAULT_ARROW_POSITION * 2 ? DEFAULT_ARROW_POSITION : null;
 
-  // console.log(horizontalPosition);
-  // console.log(_.get(container, 'clientWidth'));
   let calculatedArrowStyles = {};
   switch (true) {
     case _.includes(['bottom-start', 'top-start'], placement) && !_.isNull(horizontalPosition):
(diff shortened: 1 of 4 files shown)


src/components/Popover/Popper.jsx
```diff
@@ -64,22 +62,25 @@ const Popper = ({
...
 
-  const { styles, attributes, update } = usePopper(refElement, popperElement, {
-    modifiers: [
-      { name: 'arrow', options: { element: arrowElement } },
-      {
-        name: 'offset',
-        options: {
-          offset: [0, 6],
-        },
+  // react-popper not using subsequent settings to overwrite preceding ones
```
Harris-YWang, line 65: ^ If the modifiers is `[{name:'flip', options:{}}, {name:'flip', options:{altBoundary: true}}]`, the second object is not working. So, set up default modifiers only if there is no explicitly defined by users

superpower1 (approved):

vinteo (approved):

bchew:
Merging :rocket:

src/components/Popover/Popper.jsx
```diff
@@ -64,22 +62,25 @@ const Popper = ({
...
+    });
+  if (!_.find(modifiers, { name: 'flip' }))
+    defaultModifiers.push({
+      name: 'flip',
+      options: {
+        altBoundary: true,
       },
-    ].concat(modifiers),
+    });
+
```
superpower1, line 81: You can change this part to:
```
_(modifiers).keyBy('name').merge(defaultModifiers).values().value()
```
which will allow user to partially change the default modifier

Harris-YWang: not necessary but could be an improvement, please create a new pr if still want it

PR #1150, opened 2021-05-05
fix: fix hover dropdown menu does not display single menu item

lightbringer1991:
Fix Hover Dropdown Menu does not display when there is only 1 menu item.

- [ ] Yes
- [x] No


src/components/HoverDropdownMenu/index.jsx
```diff
@@ -38,16 +38,32 @@ const HoverDropdownMenu = ({ arrowPosition, headerText, hoverComponent, children
...
         <Popover
-          placement={`bottom-${arrowPosition === 'left' ? 'start' : 'end'}`}
+          popoverClassNames={popoverClassNames}
+          placement={placement}
           triggers={['disabled']}
           isOpen={isOpen || mouseInPopover}
           title={headerText}
           popoverContent={<ul className="list-unstyled">{children}</ul>}
           popperRef={setPopperNode}
+          arrowStyles={arrowPosition === 'none' ? { display: 'none' } : null}
```
Harris-YWang, line 66: maybe add a `modifiers` props here as well, in case you want to customize it later

Harris-YWang: you can either expose `popoverModifiers` or considering pre-defined modifiers like set the width of the popper as the width of the reference

Harris-YWang (approved):
LGTM

superpower1 (approved):

src/components/HoverDropdownMenu/index.jsx
Harris-YWang: sorry, I just found we have a `props` named `popoverProps` in the `<TextEllipsis />` so maybe pass `popoverProps` to the `<Popover />`

PR #1168, opened 2021-06-21
[ASC-75][ADS-5897] chore: upgrade react to 17

devharris7:
## Description

After deprecating `react-bootstrap`, our step is to upgrate react to version `17.0.0`.

- Upgrade `react-select` to eliminate the warning of deprecated `componentWillReceiveProps` from React 17
- React 17 uses `focusIn` and `focusOut` to mock `onFocus` and `onBlur` instead of `focus` and `blur`. https://reactjs.org/blog/2020/08/10/react-v17-rc.html#aligning-with-browsers
- React 17 supports `async` act to mock up opening popover, which removes a previous test warning.
- Also, we abused `console.error = jest.fn()` more or less. Fix all of these mock method and make sure to restore `console.error` at end of each test case.
- `enzyme` didn't have ` enzyme-adapter-react-17`, which means enzyme testing cases may not be compatible with new React 17. For `adslot-ui`, we are using `react-testing-library` mostly, this defect is not affecting badly. 
- For the upgrade of `webpack 5`, there might be extra setting for React 17, so if webpack 5 is deployed first, this upgrade pr need to fix some conflicts regarding `webpack 5` later

- [X] Yes
- [ ] No


xiaofan2406:
no longer required

PR #1175, opened 2021-07-02
[ASC-129] docs: add popover example for triggers

Harris-YWang:
## Description

As discussed, we can use `<Popover triggers>` to replace some use cases of `<Popover.WithRef>` and make code leaner. Record the use case as a demo.
Kapture 2021-07-02 at 10 00 34

- [ ] Yes
- [x] No

### DIFF
--- a/www/examples/Popover.mdx
+++ b/www/examples/Popover.mdx
@@ -74,6 +74,39 @@ render(
 
 <br />
 <br />
+<h4>Use Popover triggers as an alternative to Popover.WithRef.</h4>
+<h4>Code will be leaner with the following way.</h4>
+
+```jsx live=true
+const Example = () => {
+  const [safeMode, setSafeMode] = React.useState(true);
+  const toggleSafeMode = () => setSafeMode(!safeMode);
+
+  return (
+    <React.Fragment>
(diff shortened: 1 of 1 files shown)


www/examples/Popover.mdx
```diff
@@ -72,6 +72,39 @@ render(
...
+```jsx live=true
+const Example = () => {
+  const [safeMode, setSafeMode] = React.useState(true);
+  const toggleSafeMode = () => setSafeMode(!safeMode);
+
+  return (
+    <React.Fragment>
+      <Button onClick={toggleSafeMode}>{`Turn ${safeMode ? 'off' : 'on'} Safe Mode`}</Button>
+      <Popover
+        isOpen={safeMode}
```
renrizzolo, line 89: if this line is removed it still trigger on hover when `safeMode` is true, right?

Harris-YWang: Thinking this line is only used for the initial state of the popover, referring to the above demo.

superpower1 (approved):

glenkurniawan-adslot (approved):

vinteo (approved):

bchew:
Merging :rocket:

PR #1179, opened 2021-07-08
fix: ActionPanel body content can always overflow

renrizzolo:
## Description

- Set ActionPanel max-height and overflow by default so we don't have to add it manually in each custom component's css
  - the ActionPanel as a modal should always have a max-height so that content can overflow on small screens or when there is a lot of content
- also fix a small bug with the fadein animation that caused a bit of a flicker when opening a modal

Fixes ADS-4854

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/ActionPanel/styles.scss
+++ b/src/components/ActionPanel/styles.scss
@@ -18,6 +18,11 @@
     background-color: $color-white;
     margin: 30px auto;
     border: 0;
+
+    .aui--action-panel-body {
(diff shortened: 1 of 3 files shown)


Harris-YWang (approved):
LGTM

pphminions (approved):
LGTM 🍏

superpower1 (approved):

pphminions:
merging 🚀

PR #1194, opened 2021-08-27
fix: [carousel] prevent clicks when swiping

renrizzolo:
When dragging the carousel and letting go on top of a child with an `onClick`, that `onClick` will be fired.

## Description

This PR adds a hook that can be used to prevent unwanted clicks from firing when swiping.

### Usage
```jsx
import { Carousel } from 'adslot-ui'
 const Component = () => {
  // the returned value is an object that has the required click capture handlers
  const itemProps = Carousel.usePreventSwipeClicks(); 
  return (
    <Carousel>
      <button {...itemProps} onClick={someClickHandler}></button>
    </Carousel>
  )
}
```

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/Carousel/index.jsx
+++ b/src/components/Carousel/index.jsx
@@ -36,4 +36,36 @@ CarouselComponent.defaultProps = {
   dots: true,
 };
 
+const SWIPE_DELTA = 3;
+
+const usePreventCarouselSwipeClicks = () => {
+  const [mousePos, setMousePos] = React.useState({});
+
+  const onMouseDownCapture = (e) => {
+    setMousePos({ clientX: e.clientX, clientY: e.clientY });
+  };
+
+  const onClickCapture = (e) => {
+    const deltaX = Math.abs(mousePos.clientX - e.clientX);
+    const deltaY = Math.abs(mousePos.clientY - e.clientY);
+
+    if (deltaX > SWIPE_DELTA || deltaY > SWIPE_DELTA) {
+      e.stopPropagation();
+      e.preventDefault();
+    }
+  };
+
+  return {
+    onMouseDownCapture,
+    onClickCapture,
+  };
+};
+
+/**
+ * A hook to prevent child click handlers from firing immediately after swiping the Carousel
+ *
+ * @returns {object} - to be spread onto any carousel items with `onClick` handlers
+ */
+CarouselComponent.usePreventSwipeClicks = usePreventCarouselSwipeClicks;
+
 export default CarouselComponent;
--- a/src/components/Carousel/index.spec.jsx
+++ b/src/components/Carousel/index.spec.jsx
@@ -9,6 +9,19 @@ const getByDataIndex = queryByAttribute.bind(null, 'data-index');
 const getById = queryByAttribute.bind(null, 'id');
 const queryAllByClass = queryAllByAttribute.bind(null, 'class');
 
+const CarouselHookTest = React.forwardRef(({ onClick }, ref) => {
+  const props = Carousel.usePreventSwipeClicks();
+  return (
+    <Carousel ref={ref} slidesToShow={1}>
+      {_.map(new Array(5), (value, index) => (
+        <button {...props} onClick={onClick} id={`btn-${index}`} key={index}>
+          <img src={`path/to/image-${index}.jpg`} alt={index} id={`img-${index}`} />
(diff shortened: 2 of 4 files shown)


src/components/Carousel/index.jsx
```diff
@@ -10,10 +11,28 @@ const baseClass = 'aui--carousel-component';
...
+      event.stopPropagation();
+      event.preventDefault();
+    }
+  };
 
   return (
-    <Slider {...props} ref={ref} className={classNames(baseClass, className)}>
-      {children}
+    <Slider {...props} ref={mergedRef} className={classNames(baseClass, className)}>
+      {React.Children.map(children, (child) => (
```
xiaofan2406, line 31: This can cause issues when using fragments in children.

had a quick google search, https://github.com/akiran/react-slick/issues/848#issuecomment-438903613.
something like this be a good solution for now. maybe extract the logic, and export it has a hook.

renrizzolo: I tried this on fragments and it seems to work fine.
Regardless, that example avoids mapping the children by using an items prop which would be a breaking change for us.

src/index.js
```diff
@@ -34,6 +34,7 @@ import AlertInput from './components/AlertInput';
 import ButtonGroup from './components/ButtonGroup';
 import Breadcrumb from './components/Breadcrumb';
 import Carousel from './components/Carousel';
+import { usePreventCarouselSwipeClicks } from './components/Carousel/hooks';
```
renrizzolo, line 37: Not sure if this is too verbose?
Also, there aren't any other exported hooks so I wasn't sure where would be best to export it from.
Maybe better to put the hook in the same file as the component?
`import Carousel, { usePreventCarouselSwipeClicks } from './components/Carousel';`

Harris-YWang (dismissed):
Need to run `npm start` to update `props.json`, LGTM for the rest part

src/index.js
Harris-YWang: I suppose you can refer to the <RichTextEditor />, and maybe rename it to `usePreventSwipeClick` then?

renrizzolo: @Harris-YWang do you mean adding it to the Carousel itself so it's used like `Carousel. usePreventSwipeClick()`?

Harris-YWang: yea, as this hook is only used for Carousel, I suppose. what do you think of?

renrizzolo: Yep, makes sense, as you said it's only for Carousel.

Harris-YWang (approved):
LGTM

src/components/Carousel/index.jsx
```diff
@@ -2,6 +2,7 @@ import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import React from 'react';
 import Slider from 'react-slick';
+import mergeRefs from 'react-merge-refs';
```
xiaofan2406, line 5: this lib breaks the assumption that the ref object is always the "same" between renders, as it returns a function, not a ref object.

if we need to stay with this implementation, i prefer 

```js
function useMergedRefs(...refs) {
  const targetRef = React.useRef(null);

  React.useEffect(() => {
    refs.forEach((ref) => {
      if (!ref) return;

      if (typeof ref === 'function') {
        ref(targetRef.current);
      } else {
        ref.current = targetRef.current;
      }
    });
  }, [refs]);

  return targetRef;
}
```

as the returned result is a ref object.

src/components/Carousel/index.jsx
```diff
@@ -36,4 +37,35 @@ CarouselComponent.defaultProps = {
...
+    setIsSwiping(carouselRef.current?.innerSlider?.state?.swiping);
+  };
+
+  const onClickCapture = (event) => {
+    if (isSwiping) {
+      event.stopPropagation();
+      event.preventDefault();
+    } else {
+      if (carouselRef.current?.innerSlider) {
+        carouselRef.current.innerSlider.clickable = true;
```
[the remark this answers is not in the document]
renrizzolo: Fair enough. I couldn't find it in the docs so I've refactored to not use it.

xiaofan2406 (approved):

superpower1 (approved):

Harris-YWang (approved):

bchew:
Merging :rocket:

PR #1227, opened 2021-10-11
chore: add cancel text props on action panel

glenkurniawan-adslot:
## Description

Add a cancelText props on Action Panel, so the user can specify the text for close/cancel button when action panel is being used as modal (default is 'Cancel');

- [ ] Yes
- [x] No


glenkurniawan-adslot:
> Changes LGTM. Just wondering whether we should also support custom button, for example a different icon. If yes I think it's better to make it as a node.

hmm.. we can change the icon using `closeIcon` props (but the icon is used only for non modal action panel).. Do we want to support the close/cancel button (when it is being used as modal) not as a button?

PR #1349, opened 2022-02-04
[ADS-6456] fix: tree shaking & regeneratorRuntime

renrizzolo:
- add `sideEffects` for tree shaking of esm build (ref: https://webpack.js.org/guides/tree-shaking/)
- add `@babel/plugin-transform-runtime` to fix undefined `regeneratorRuntime` in esm/cjs builds. 
  - async/await in `RichTextEditor` transpiles to use `regeneratorRuntime` but it's not actually imported/polyfilled.
  - This is due to not using `corejs` builtins in the esm/cjs builds, but also having a `browserslist` that would cause polyfilling of async/await. 
  
## Description
I have tested the tree shaking as below - looks to be working properly (these are unminified stats)

```
import { Button } from 'adslot-ui';
```
Without `sideEffects` field, using esm build:
```
dist/assets/index.f07c8cc8.js      128.05 KiB / gzip: 23.11 KiB
dist/assets/vendor.a8808d79.js     2080.81 KiB / gzip: 424.74 KiB
```

With `sideEffects` field, using esm build:

```
dist/assets/index.53979bf6.js      5.92 KiB / gzip: 1.96 KiB
dist/assets/vendor.83d8442c.js     407.97 KiB / gzip: 90.17 KiB
```

For us, the common chunk size could be reduced by around 8%.

### minified adslot-ui bundle

| chunk                     | parsed    | gzip      |
| ------------------------- | --------- | --------- |
| common                    | 4.46 MB   | **1.17 MB**   |
| \|---node_modules         | 1.61 MB   | 472.01 KB |
| \|---adslot-ui (inc deps) | 1.2 MB    | 347.71 KB |
| moduleX        | 789.29 KB | 139.81 KB |
| Total                     | 7.5 MB    | 2.11 MB   |

### ESM adslot-ui

| chunk              | parsed    | gzip      |
| ------------------ | --------- | --------- |
| common             | 4.26 MB   | **1.11 MB**   |
| \|---node_modules  | 1.4 MB    | 412.27 KB |
| \|---adslot-ui     | 290.16 KB | 73.87 KB  |
| moduleX | 789.29 KB | 139.81 KB |
| Total              | 7.3 MB    | 2.05 MB   |

### ESM adslot-ui with tree shaking

| chunk              | parsed    | gzip      |
| ------------------ | --------- | --------- |
| common             | 4.19 MB   | **1.09 MB**   |
| \|---node_modules  | 1.33 MB   | 392.48 KB |
| \|---adslot-ui     | 274.25 KB | 71.13 KB  |
| moduleX | 865.34 KB | 161.3 KB  |
| Total              | 7.31 MB   | 2.06 MB   |

### Why does x increase?
all of adslot-ui is included in that module, so the portions that are unused in `common` are included in `moduleX`.

- [ ] Yes
- [x] No

### DIFF
--- a/config/webpack.config.prod.js
+++ b/config/webpack.config.prod.js
@@ -89,6 +89,7 @@ module.exports = webpackMerge(commonConfig, {
       },
       {
         test: /\.((c|sc)ss)$/i,
+        sideEffects: true,
         use: [
           MiniCssExtractPlugin.loader,
           {
--- a/package.json
+++ b/package.json
@@ -10,6 +10,10 @@
     "lib",
     "es"
   ],
+  "sideEffects": [
+    "./src/**/*.scss",
+    "./dist/*.css"
+  ],
   "license": "MIT",
   "scripts": {
     "generate-docs": "node ./scripts/generateDocs.js",
@@ -71,6 +75,7 @@
     "@babel/cli": "^7.17.0",
     "@babel/core": "^7.17.0",
     "@babel/preset-env": "^7.16.11",
+    "@babel/plugin-transform-runtime": "^7.17.0",
     "@babel/preset-react": "^7.16.7",
     "@commitlint/cli": "^16.1.0",
     "@commitlint/config-conventional": "^16.0.0",
--- a/babel.config.js
+++ b/babel.config.js
@@ -35,11 +35,21 @@ const env = {
   },
   esm: {
     presets: [['@babel/preset-env', { useBuiltIns: false, modules: false }]],
-    plugins: ['jsx-remove-data-test-id', inlineSvgPlugin, ['transform-remove-imports', { test: '\\.(scss|css)$' }]],
+    plugins: [
+      ['@babel/plugin-transform-runtime', { regenerator: true }],
+      'jsx-remove-data-test-id',
+      inlineSvgPlugin,
+      ['transform-remove-imports', { test: '\\.(scss|css)$' }],
+    ],
   },
   cjs: {
     presets: [['@babel/preset-env', { useBuiltIns: false, modules: 'cjs' }]],
-    plugins: ['jsx-remove-data-test-id', inlineSvgPlugin, ['transform-remove-imports', { test: '\\.(scss|css)$' }]],
+    plugins: [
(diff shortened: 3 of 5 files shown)


bchew (dismissed):

knilink (dismissed):

package.json
```diff
@@ -10,6 +10,14 @@
     "lib",
     "es"
   ],
+  "sideEffects": [
+    "**/*.css",
+    "**/*.scss",
+    "./es/index.js",
+    "./es/lib/utils.js",
```
knilink, line 17: have you tried `./es/**/*.js`?

renrizzolo: @knilink If all of individual components are added, that would be saying 'always include this component' when tree shaking, regardless of if it was imported or not.

renrizzolo: turns out we don't have any files with side effects, so I've changed it to `false`

xiaofan2406 (approved):

bchew (approved):

shn2016 (dismissed):

package.json
```diff
@@ -10,6 +10,10 @@
     "lib",
     "es"
   ],
+  "sideEffects": [
+    "./src/**/*.scss",
```
renrizzolo, line 14: Did some further testing and it seems that for production builds, the styles will be tree shaken out of the jsx files, so they need to be included

config/webpack.config.prod.js
```diff
@@ -89,6 +89,7 @@ module.exports = webpackMerge(commonConfig, {
       },
       {
         test: /\.((c|sc)ss)$/i,
+        sideEffects: true,
```
renrizzolo, line 92: Because the docs is part of the same package as the lib, we need to signify that the docs styles should be included.

package.json
```diff
@@ -10,6 +10,10 @@
...
+    "./dist/*.css"
```
renrizzolo, line 15: I found that even `import 'adslot-ui/dist/adslot-ui-main.css';` in a separate webpack app would tree shake the css file out.

knilink (approved):

bchew (approved):
Merging :rocket:

bchew (approved):

PR #1350, opened 2022-02-06
[ADS-6363][ASC-22] feat: add type definitions

renrizzolo:
## Description
- Generate type definition files for all components
- Add scripts to facilitate automatic type generation
  - created a babel plugin to 'normalize' the PropTypes, then use `react-to-typescript-definitions` to convert PropTypes to types. Types are saved to `d.ts` files alongside each component.
  - some manual overrides are added so 3rd party lib types are applied - `scripts/generate-types/typesPostFixes.js`
  
A few surface changes were done for all of this to operate smoothly
- make all component names consistent  - the exports need to match the type defs, and they are generated from the component name
- remove static class properties
- simplify some prop type definitions that were too complicated to parse

## Benefits 
Type definitions unlock a huge boost for developer experience, even for plain JS projects. That's the driving motivator behind this PR. 

With a typescript enabled code editor like VSCode, all component props and their definitions and descriptions are available and will autocomplete. Not only is this a huge productivity boost, it also reduces the chance for misspelled props or prop values.

The gifs below show some examples of: prop autocompletion/suggestion, prop descriptions via comments, type autocompletion/suggestion, type checking, member expression types. (`// @ts-check` is not needed, I'm just using it here to show when types would be incorrect.) 

Kapture 2022-02-04 at 16 06 09

Kapture 2022-02-04 at 16 18 41

## Usage

Generate types for all components
`npm run generate-types`

<details>
<summary>More...</summary>

Generate types for a single component
`npm run generate-types -- --only=Button`

Generate types for a single component (this will run for all jsx files in the folder matching Button)
`npm run generate-types -- --only=Button`

Generate types for a specific file
`npm run generate-types -- --only=Button/index.jsx`

globs are accepted:
`npm run generate-types -- --only="RichTe*`

Copy generated types to es/*
`npm run generate-types -- --copy`

Generate types with debugging output
`npm run generate-types -- --debug`

Check validity of generated types
`npm run type-check`
</details>

## Proposed workflow for updating/adding types

Maintaining consistency of the type definitions is important, and it's up to us as contributors to do so. 

Note that I haven't added `generate-types` to the `dist` script. This is to avoid any unintentional breaking changes slipping through. The onus is on the dev to generate and commit type definitions when making changes to or adding new components/props.

**In most cases it will be as simple as running `npm run generate-types`, then verifying and committing the changes.**

Of course, the Prop Types themselves are responsible for the type definitions, so please make sure Prop Types are as correct as possible (e.g, if you know the prop accepts certain strings only, the definition should be `PropTypes.oneOf(['string1', 'string2'])` as opposed to `PropTypes.string`

### Prop descriptions
Comments (a la JSDoc) above a prop type will carry across to the type def, so feel free to describe the prop there to help other users!

### Advanced usage
Advanced usage, e.g when wrapping types of a 3rd party library into a component, should be manually adjusted in `scripts/generate-types/typesPostFixes.js`. 

## Caveats
- nested `PropTypes.shape` will not work
- JS in prop type definitions won't be evaluated (see below)

This works:
```js
const reusableTypes = ['primary', 'secondary', 'warning']
Component.PropTypes = {
  theme: PropTypes.shape(reusableTypes)
}
```

This **doesn't** work:
```js
const reusableTypes = ['primary', 'secondary', 'warning']
const filteredTypes = reusableTypes.filter((type) => type !== 'warning')
Component.PropTypes = {
 theme: PropTypes.oneOf(filteredTypes),
}
```

## Other options to explore/discuss
1. now we have a baseline of types, forget the auto-generation, and manually curate them
2. as above but create a definitely typed repo `@types/adslot-ui` to separate them further
3. convert the lib to typescript entirely

I feel that **1** would be potentially better due to being able to define the types as precisely as we need, without having to rely on modifications or be limited by the PropTypes lib – but would require more work and TS knowledge from contributors.

- [x] Yes
- [ ] No

Any typescript projects using `adslot-ui` may need to update types accordingly. E.g if custom types were added they should be replaced with the types from the library.


scripts/generate-types/copyTypes.js
```diff
@@ -0,0 +1,22 @@
+const fs = require('fs-extra');
```
knilink, line 1: i think `fs-extra` can be deprecated in favour of `fs/promises`

renrizzolo: @knilink I'm using fs-extra's `outputFile` quite a bit which doesn't have an equivalent in node as far as I know (it creates directories if they don't exist when writing a file, so it's much more convenient)

xiaofan2406: if it can be rewritten to reduce an extra package dep, it would be preferred

scripts/build.js
```diff
@@ -99,13 +104,22 @@ function build(previousFileSizes) {
...
+      } catch {
+        await fs.promises.mkdir(buildPath);
+      }
+      fs.readdirSync(buildPath).forEach((f) => {
+        fs.rmSync(`${buildPath}/${f}`, { recursive: true });
+      });
+    }
     // Merge with the public folder
-    if (process.env.NODE_ENV === 'production') {
-      process.env.DEMO_ASSETS ? copyDemoAssets() : copyPublicFolder();
```
renrizzolo, line 108: I don't think `copyPublicFolder()` is ever used (`DEMO_ASSETS` and `production` are always used together).
Can anyone confirm?

scripts/generate-types/copyTypes.js
renrizzolo: added a separate commit to remove `fs-extra`

vinteo (approved):

src/components/Alert/index.d.ts
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
+
+export type AlertType = "success" | "info" | "warning" | "danger";
+
+export interface AlertProps {
+  /**
+   * ['success', 'info', 'warning', 'danger']
+   */
```
shn2016, line 8: this comment is a bit duplicated though, so is other similar one

src/components/Breadcrumb/index.d.ts
```diff
@@ -0,0 +1,13 @@
+import * as React from "react";
+
+export interface BreadcrumbProps {
+  rootNode?: any;
```
shn2016, line 4: found this in default `rootNode: { id: 'all', label: 'All' }`, and it looks same type with `BreadcrumbNodeNode` in `src/components/Breadcrumb/Node/index.d.ts`, is there some consideration for this?

src/components/Checkbox/index.d.ts
```diff
@@ -0,0 +1,48 @@
+import * as React from "react";
+
+export type CheckboxChecked = any | any | "partial";
```
shn2016, line 3: cant' make boolean | "partial" ?

shn2016:
Great work!

shn2016 (approved):

src/components/Alert/index.d.ts
renrizzolo: The redundant prop comments are for to the docs site's prop descriptions. The type generation just pulls all the comments from the prop types.

xiaofan2406 (dismissed):

src/components/Checkbox/index.d.ts
renrizzolo: will see if there's a way to get this to be a boolean. Not sure why it's giving any actually.

src/components/Breadcrumb/index.d.ts
renrizzolo: Good find, fixed.

src/components/Alert/index.d.ts
renrizzolo: When we move to storybook, the prop values will be extracted properly so we'll be able to remove these comments then.

scripts/generate-types/generateTypes.js
```diff
@@ -0,0 +1,144 @@
...
+      parsed.map(async (code, i) => {
+        const result = await generateFromSource(null, code, {
+          babylonPlugins: ['exportDefaultFrom', 'transformImports'],
+        });
+
+        const component = allComponents[i];
+        console.log(chalk.green.bold(`Generated type defs for ${component.componentName}`));
+
+        const output = typesPostFixes(component.componentName, result);
+        const prettifiedOutput = prettier.format(output, { parser: 'typescript', ...pkg.prettier });
```
renrizzolo, line 126: added prettier settings from package.json so output is consistently formatted

xiaofan2406 (approved):

knilink (approved):

PR #1354, opened 2022-03-01
[ASC-144] fix: richTextEditor onChange issue

shn2016:
## Description

The `RichTextEditor` use draft-js-editor-plugin library, and it allows component trigger `onChange` at initialRender, see [source code here](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/packages/editor/src/Editor/index.tsx) from `113-139`.
And because of the nature of [EditorState](https://draftjs.org/docs/api-reference-editor-state/#content):

> The current text content state
> The current selection state
> The fully decorated representation of the contents
> Undo/redo stacks
> The most recent type of change made to the contents

It is triggering `onChange` fairly frequent. 

And regarding html `<p><br></p>`, it is inserted by default when empty editor state convert to html.

## Changes
This PR includes changes:
1.  type of value: from `EditorState` to `string`;
2. onChange: from `EditorState onChange` to `value(html input) onChange` ;
3. if the value is `<p><br></p>`, return ``

- [x] Yes
- [ ] No

Will need to update all the affected files as API change.

### DIFF
--- a/src/components/RichTextEditor/index.jsx
+++ b/src/components/RichTextEditor/index.jsx
@@ -18,6 +18,20 @@ import './styles.css';
 
 const { EditorState, Modifier, convertToRaw, RichUtils } = draftJs;
 
+const editorStateToHTML = (input) => stateToHTML(input.getCurrentContent());
+const editorStateFromHTML = (input) => EditorState.createWithContent(stateFromHTML(input));
+const isEditorStateEmpty = (html) => html === '<p><br></p>';
+
+const getInitialEditorState = (value, initialValue) => {
+  if (value) {
+    return editorStateFromHTML(value);
+  }
+  if (initialValue) {
+    return editorStateFromHTML(initialValue);
+  }
+  return EditorState.createEmpty();
+};
+
 const RichTextEditor = ({
   className,
   value,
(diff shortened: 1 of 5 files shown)


src/components/RichTextEditor/index.jsx
```diff
@@ -67,17 +80,18 @@ const RichTextEditor = ({
     return { plugins, MentionSuggestions };
   }, []);
 
+  const isHTMLChanged = (oldState, newState) => editorStateToHTML(oldState) !== editorStateToHTML(newState);
```
xiaofan2406, line 83: double check how much is this costing. as it happens on every key stroke

src/components/RichTextEditor/index.jsx
```diff
@@ -75,16 +89,16 @@ const RichTextEditor = ({
   }, []);
 
   const handleOnChange = (newState) => {
-    if (!value) {
-      setEditorState(newState);
-    }
-    if (onChange) {
-      onChange(newState);
+    const oldHTML = editorStateToHTML(editorState);
+    const newHTML = editorStateToHTML(newState);
```
shn2016, line 93: `editorStateToHTML` process editorState to content and use markdown to generate html, works fast
optimised to avoid extra calc.

xiaofan2406 (approved):

renrizzolo (dismissed):

xiaofan2406 (approved):

PR #1355, opened 2022-03-02
[ASC-143] Button update

renrizzolo:
## Description

- Refactor buttons based on ui component audit. 
- Changed/renamed props still work but have a deprecation error in dev.
- remove support for Button component becoming an anchor with the `href` attribute
- add `Anchor` component, which shares the API of Button (aside from `isLoading`), for rendering anchor links
- remove sass map and mixin usage for upcoming sass deprecation

### Styles
- remove bootstrap button scss dep entirely
- simplify/remove redundant scss for button variants
- add focus ring with `focus-visible` (keyboard interactions only)
- remove the 1px transform to the `:active` state
- unify the border/background colours for solid buttons (i.e no slightly darker shade border on hover)

### API
- change `theme` prop to `color`
- remove support for `warn` and `info` colors (edit: removed in docs only. These are still part of the global theme colors map thus still applicable to buttons)
- add `variant` prop to consolidate inverse/ borderless/solid variants
- add `icon` prop and associated styles
- add `round` boolean prop (icon with no child only)

```jsx
<Button color="primary" variant="inverse" size="large">
  Label
</Button>
``` 

### Docs
- separate examples into sections, add wording from design guide

## Breaking changes
- `Button` no longer renders `a` tags when using `href` (use `Anchor` instead)
- internal classnames have changed: `btn-*` is now `aui-*` e.g `btn-primary` is now `aui-primary`. These were made more generic because they are being used on both Button and Anchor.

Screen Shot 2022-03-04 at 12 29 56 pm

- [x] Yes
 The link variant has changed from a padded button that looks like a link to an inline style link with no padding.
The closest equivalent to the old link button is `<Button color="primary" variant="borderless" />`
- [ ] No

---
### Migration

Aside from the prop name changes, these are some additional guidelines for Button/Anchor usage below.
**In general, you should not need to override the css of a Button/Anchor component.** 

- Icons
Always use the icon prop for icons. This will ensure the correct sizing for icon buttons with no label, and the correct spacing for icon buttons with a label.

Old:
```diff
// icon only
- <Button>{svgIcon}<Button>
// icon with text
- <Button>{svgIcon} Content<Button>
```

New:
```jsx
// icon only
<Button icon={svgIcon} aria-label="Search" />
// icon with text
<Button icon={svgIcon}>Search</Button>
```

- Links
Links with href must use the `Anchor` component
Links with onClick and no href should use `Button`, but consider whether a link variant is appropriate, rather than e.g a borderless button.

Old:
```diff
- <Button className="btn-link" href="...">...</Button>
```

New:
```jsx
// navigation or other javascript action
<Button variant="link" onClick={}>...</Anchor>
// anchor link
<Anchor variant="link" href="...">...</Anchor>
```

---
### Upgrading (for applications using `adslot-ui` Button): 
Create the below `transform.js` in the root of your project

Run `npx jscodeshift --extensions="js,jsx" --ignore-pattern="**/node_modules/**" ./`
(See: [jscodeshift cli options](https://github.com/facebook/jscodeshift#usage-cli))

Shouldn't need any intervention in most cases except for filling out an aria-label for icon buttons, but please verify the diffs, as this hasn't been thoroughly tested. 

N.B the underlying AST lib jscodeshift uses will make some formatting changes to unrelated files! Running prettier will revert these changes (except for minified files, potentially).

*Buttons with icons as children should be manually changed to the new prop: `icon={...}`. This is important to have the correct dimensions (when using the icon prop, the button will have even width/height)
<details>
<summary>transform.js</summary>

```js
module.exports = function transformer(file, api) {
  const j = api.jscodeshift;

  const $j = j(file.source);

  return $j
    .find(j.ImportDeclaration, { source: { value: 'adslot-ui' } })
    .forEach((path) => {
      const defaultImport = j(path).find(j.ImportSpecifier);
      if (defaultImport.size() === 0) {
        return;
      }

      const hasButton = j(path)
        .find(j.ImportSpecifier)
        .find(j.Identifier)
        .some((n) => n.get('name').value === 'Button');
      if (!hasButton) {
        return;
      }

      const buttons = $j.findJSXElements('Button');

      buttons.forEach((buttonPath) => {
        const $button = j(buttonPath);

        if ($button.size() === 0) {
          return;
        }
        inverseToVariant(j, $button);
        linkThemeToVariant(j, $button);
        themeToColor(j, $button);
        borderlessClassToVariant(j, $button);
        largeToSize(j, $button);
        iconWithoutAriaLabel(j, $button);
        return;
      });
    })
    .toSource();
};

function iconWithoutAriaLabel(j, b) {
  if (b.find(j.JSXAttribute, { name: { name: 'icon' } }).size() === 0) {
    return;
  }
  if (b.find(j.JSXAttribute, { name: { name: 'aria-label' } }).size() === 1) {
    return;
  }
  b.get('attributes').push(j.jsxAttribute(j.jsxIdentifier(`aria-label=""`)));
}

function inverseToVariant(j, b) {
  if (b.find(j.JSXAttribute, { name: { name: 'inverse' } }).size() === 0) {
    return;
  }
  let variable;
  b.find(j.JSXAttribute, { name: { name: 'inverse' } }).forEach((nodePath) => {
    if (j(nodePath).get('value').value) {
      // WARNING: 'variable' doesn't take into account whole expressions or class members
      // e.g inverse={this.stuff > 0}
      // would become variant={stuff ? 'inverse' : undefined}
      variable = j(nodePath).find(j.JSXExpressionContainer).find(j.Identifier).get('name').value;
    }

    j(nodePath).remove();
  });
  if (b.find(j.JSXAttribute, { name: { name: 'theme' }, value: { value: 'link' } }).size() === 1) {
    return;
  }

  if (b.find(j.JSXAttribute, { name: { name: 'variant' } }).size() > 0) {
    return;
  }

  b.get('attributes').push(
    j.jsxAttribute(j.jsxIdentifier(variable ? `variant={${variable} ? 'inverse' : undefined}` : `variant="inverse"`))
  );
}

function linkThemeToVariant(j, b) {
  if (b.find(j.JSXAttribute, { name: { name: 'theme' }, value: { value: 'link' } }).size() === 0) {
    return;
  }

  b.find(j.JSXAttribute, { name: { name: 'theme' }, value: { value: 'link' } }).forEach((nodePath) => {
    j(nodePath).remove();
  });

  b.get('attributes').push(j.jsxAttribute(j.jsxIdentifier(`variant="link"`)));
}

function themeToColor(j, b) {
  if (b.find(j.JSXAttribute, { name: { name: 'theme' } }).size() === 0) {
    return;
  }
  b.find(j.JSXAttribute, { name: { name: 'theme' } }).forEach((nodePath) => {
    nodePath.node.name = 'color';
  });
}

function largeToSize(j, b) {
  if (b.find(j.JSXAttribute, { name: { name: 'large' } }).size() === 0) {
    return;
  }
  b.find(j.JSXAttribute, { name: { name: 'large' } }).forEach((nodePath) => {
    nodePath.node.name = 'size';
    nodePath.node.value = '"large"';
  });
}

function borderlessClassToVariant(j, b) {
  if (b.find(j.JSXAttribute, { name: { name: 'className' } }).size() === 0) {
    return;
  }
  let modified;
  b.find(j.JSXAttribute, { name: { name: 'className' } }).forEach((nodePath) => {
    if (j(nodePath).get('value')?.value?.value?.includes('btn-borderless')) {
      modified = true;
      nodePath.node.value.value = nodePath.node.value.value.replace('btn-borderless', '').trim();
    }
    if (modified) {
      b.get('attributes').push(j.jsxAttribute(j.jsxIdentifier(`variant="borderless"`)));
    }
  });
}

```
</details>


src/components/Button/index.jsx
```diff
@@ -9,18 +9,59 @@ import { expandDts } from '../../lib/utils';
...
+  // eslint-disable-next-line no-console
+  if (round && !icon) console.warn('Adslot UI Button: round can only be used with an icon.');
+
+  if (!!icon && _.isEmpty(children) && !rest['aria-label'])
+    // eslint-disable-next-line no-console
+    console.error('Adslot UI Button: an aria-label is required on icon buttons.');
+
+  if (isLink && (color !== 'default' || size === 'large'))
+    // eslint-disable-next-line no-console
+    console.warn('Adslot UI Button: buttons with the "link" variant do not inherit size and color properties.');
```
xiaofan2406, line 47: let's create a util function to do all this error & warning.

during dev env, we should just throw errors, and all these should be dropped for prod build.

check https://github.com/alexreardon/tiny-invariant as an example.

xiaofan2406, line 47: let's create a util function to do all this error & warning.

during dev env, we should just throw errors, and all these should be dropped for prod build.

check https://github.com/alexreardon/tiny-invariant as an example.

src/lib/utils.js
```diff
@@ -24,7 +24,16 @@ export const classSuffixHelper = ({ classSuffixes, suffixOptions, componentClass
...
 export const expandDts = (dtsString) => (dtsString ? { 'data-test-selector': dtsString } : {});
 
+const isProduction = process.env.NODE_ENV === 'production';
+const prefix = 'AdslotUI';
+export const invariant = (condition, message) => {
+  if (!isProduction && !condition) {
+    throw new Error(`${prefix} ${message}`);
+  }
+};
+
```
renrizzolo, line 34: @xiaofan2406 throw in dev and ignore in prod. No need for an optional console.warn instead of throwing?

src/components/Anchor/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+import { variants } from '../Button';
+import { sizes } from '../Button';
+
+const Anchor = (props) => {
+  const { href, color, size, variant, round, fullWidth, icon, children, className, disabled, dts, isLoading, ...rest } =
+    props;
+  const isLink = variant === 'link' || className?.match(/\b(btn-link)\b/)?.[0];
+
+  invariant(
+    props.hasOwnProperty('href'),
```
xiaofan2406, line 17: `href`

xiaofan2406 (approved):

PR #1407, opened 2022-06-07
chore: upgrade react-to-typescript-definitions

renrizzolo:
## Description
I was able to do the required upgrades to react-to-typescript-definitions, so the lodash version should now be at `4.17.21`. https://github.com/KnisterPeter/react-to-typescript-definitions/pull/1405

- [ ] Yes
- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
     "react-html-parser": "^2.0.2",
     "react-live": "^2.4.1",
     "react-router-dom": "^6.3.0",
-    "react-to-typescript-definitions": "^3.1.0",
+    "react-to-typescript-definitions": "^3.1.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -174,9 +174,6 @@
     "react": "^16.14.0",
     "react-dom": "^16.14.0"
   },
-  "overrides": {
-    "lodash": "$lodash"
```
renrizzolo, line 178: @bchew although this _should_ fix the issue, not sure if you just want to keep this anyway?

bchew: Yeah, wouldn't want to keep this if it's not required

bchew (approved):
Verified locally, thanks @renrizzolo

PR #1411, opened 2022-06-09
[ADS-6838] fix: export data picker helpers

xiaofan2406:
## Description

export `registerLocale` from `react-datepicker`
and all supported locales from `date-fns`

example usage:

```jsx
import {registerLocale} from 'adslot-ui';
import enAU from 'date-fns/locale/en-AU';

registerLocale(enAU);
```

- [ ] Yes
- [x] No

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,8 @@
 import './styles/_bootstrap-custom.scss';
 import './styles/_icheck-custom.scss';
 
+import { registerLocale } from 'react-datepicker';
+
 import Accordion from './components/Accordion';
 import ActionPanel from './components/ActionPanel';
 import Alert from './components/Alert';
@@ -90,6 +92,7 @@ export {
   ConfirmModal,
   CountBadge,
   DatePicker,
+  registerLocale,
   Empty,
   fastStatelessWrapper,
   FilePicker,


src/index.js
```diff
@@ -1,6 +1,9 @@
 import './styles/_bootstrap-custom.scss';
 import './styles/_icheck-custom.scss';
 
+import { registerLocale } from 'react-datepicker';
+import * as locales from 'date-fns/locale';
```
renrizzolo, line 5: This will increase the bundle size significantly. I don't think we should do this unless we implement tree shaking. imo if locales need to be used, they should be installed/imported by the consumer.

src/index.js
```diff
@@ -1,6 +1,9 @@
 import './styles/_bootstrap-custom.scss';
 import './styles/_icheck-custom.scss';
 
+import { registerLocale } from 'react-datepicker';
```
renrizzolo, line 4: We can export `setDefaultLocale ` and `getDefaultLocale` as well, for global usage

src/index.js
xiaofan2406: this isnt really the case, since we already have the esm module version, and no longer using the .min.js

src/index.js
xiaofan2406: we could when they are required. 

prefer to not to export all date-picker functions.

src/index.js
xiaofan2406: updated. as it does affect direct-web usage.

renrizzolo (approved):

renrizzolo:
merging...

PR #1414, opened 2022-06-24
[ADS-6917] RadioGroup and CheckboxGroup refactor

renrizzolo:
## Description

The main goal of this PR is to have consistent click areas and accessibility across all of our radios and checkboxes

- Make `RadioGroup` and `CheckboxGroup` accessible according to WAI/ARIA guidelines. This includes keyboard navigation with arrow keys for radio groups, tab navigation and activating with space for checkbox groups, as well as the appropriate `aria` attributes, roles.

- Add the new `box` variant to `RadioGroup` and `CheckboxGroup`, for rendering padded, button-style radio/checkbox groups

- add `getIsSelected` to `RadioGroup`, to match the `CheckboxGroup` API. This should only be used in rare cases.
- 
- add invariants to warn against using `checked` and `onChange` directly in `Checkbox`/`Radio` when within a `CheckboxGroup`/`RadioGroup`

- refactor checkbox and radio icon css, with some contrast improvements and subtle transitions. They're now much sharper due to using css for the containers and only svg for the checkbox itself.

- refactor the ListPicker component to have proper labels and click areas for the radio/checkbox (see gifs). Added hover/selected state. 

- [x] Yes

- Deprecated `inline` prop from `RadioGroup`/`CheckboxGroup` `inline-block` styles. (replaced by `orientation="horizontal"` that uses flex).
- - Deprecated the `size` prop from `Checkbox`. This was applying inline styles based on the number passed in. Don't think we need to support this, but if we do it should match other size props  e.g small/medium/large, and do the styling via classes.
- refactor classes to use `aui--radio` and `aui--checkbox` instead of `radio-component` and `checkbox-component`
- Some additional bootstrap related css removed

Keyboard navigation of radio group 
Kapture 2022-06-24 at 14 03 44

Kapture 2022-06-24 at 14 05 05

Keyboard navigation of ListPicker
Kapture 2022-06-24 at 15 37 34

Kapture 2022-06-24 at 15 43 24

Kapture 2022-06-24 at 14 03 31

### checkbox styles
old:
Screen Shot 2022-06-24 at 9 19 59 am

new:
Screen Shot 2022-06-24 at 9 20 02 am


src/components/RadioGroup/index.jsx
```diff
@@ -1,48 +1,95 @@
...
+import { useArrowFocus } from '../../hooks';
+import { expandDts, invariant } from '../../lib/utils';
+import '../RadioGroup/style.css';
+
+const itemClass = 'aui--radio';
+
+const RadioGroupContext = React.createContext({});
+
+const RadioGroupProvider = ({ children, value }) => {
+  const { value: radioValue, onChange, variant, name, getIsSelected } = value;
```
xiaofan2406, line 13: can we pls flatten these to `props`? same for the other ctx.

src/hooks/useArrowFocus.js
```diff
@@ -0,0 +1,119 @@
...
+      const nextEl = DOMList[n];
+      if (!nextEl || getIsDisabled(nextEl)) return;
+
+      nextEl.focus();
+      onFocus && onFocus(nextEl);
+    },
+    [getDOMList, onFocus]
+  );
+
+  React.useEffect(() => {
```
xiaofan2406, line 111: I think all these `useCallback` above are unnecessary, looking at how `useArrowFocus` is used internally, it is already always breaking these memoization.

I think you can just create `handleKeyDown` in this hook itself, and the only dependencies i am seeing are `selector`, `loop` and `orientation`. (may need the `ref` as well for eslint-happy)

xiaofan2406: as for `onFocus` you can

```
const onFocusRef = React.useRef(onFocus)

React.useLayoutEffect(() => {
 onFocusRef.current = onFocus
})
```

and use `onFocusRef.current` in the effect hook. this is a pattern i use for keeping callback function up2date.

src/hooks/useArrowFocus.js
```diff
@@ -0,0 +1,119 @@
...
+  )
+ * ```
+ * @param {object} options options object
+ * @param {object} options.ref ref attatched to the parent element
+ * @param {string} options.selector selector to pass to `querySelectorAll`. These are the elements that will be focused.
+ * @param {function} [options.onFocus] callback to be called when focussing an element. Receives element.
+ * @param {boolean} [options.loop] when true, navigating past the end of the list goes back to the beginning, and vice-versa
+ * @param {'vertical'|'horizontal'} [options.orientation] determines the arrow keys used based on the direction of the list
+ */
+const useArrowFocus = ({ ref, selector, onFocus, loop = true, orientation = 'vertical' }) => {
```
xiaofan2406, line 40: Q: what does the aria spec says about `tab` and `shift+tab`?

src/index.js
```diff
@@ -1,5 +1,4 @@
 import './styles/bootstrap-custom.css';
-import './styles/icheck-custom.css';
```
xiaofan2406, line 2: 🚀 👍 💯

src/components/CheckboxGroup/index.jsx
```diff
@@ -1,75 +1,82 @@
...
-            child.props.onChange(...args);
-            handleCheckboxChange(...args);
-          },
-        };
-        return <child.type {...childProps} />;
+      onChange(newValues, name, checkboxValue);
+    };
+
+    const isSelectedHandler = (itemValue) => {
+      if (getIsSelected) {
```
xiaofan2406, line 23: - `getIsChecked`? or 
- rename `checked` to `selected`

src/hooks/useArrowFocus.js
renrizzolo: @xiaofan2406 
For radio groups: Tab moves focus to the checked radio button (or the first if none are checked). Arrows move focus to and checks the next / prev radio.
For checkbox groups, no arrow navigation, regular tabbing for focus

src/components/Checkbox/index.jsx
```diff
@@ -1,49 +1,106 @@
...
+const Checkbox = ({
+  name,
+  value,
+  label,
+  dts,
+  disabled,
+  checked: checkedProp,
+  id,
+  className,
+  wrapperClass,
```
xiaofan2406, line 24: I noticed in master, we incorrectly apply `className` to both the wrapper and the `input`.

I personally think the `className` should just go on the wrapper div, and `wrapperClass` is not needed. for edge case where we want to overriding `input` styles, then we can do `> .checkbox-input`

renrizzolo: agreed

xiaofan2406 (dismissed):

xiaofan2406 (approved):

PR #1416, opened 2022-06-28
chore: add css linting

renrizzolo:
## Description
Adds `stylelint` https://stylelint.io/ with default preset and prettier compatability preset

- [ ] Yes
- [x] No

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,0 +1,9 @@
+{
+  "extends": ["stylelint-config-standard", "stylelint-config-prettier"],
+  "rules": {
+    "selector-class-pattern": null,
+    "color-function-notation": null,
+    "alpha-value-notation": "number",
+    "no-descending-specificity": [true, { "ignore": ["selectors-within-list"] }]
+  }
+}
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,0 +1,2 @@
+src/styles/bootstrap-custom.css
+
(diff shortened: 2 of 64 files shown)


.stylelintrc.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": ["stylelint-config-standard", "stylelint-config-prettier"],
+  "rules": {
+    "selector-class-pattern": null,
+    "color-function-notation": null,
+    "alpha-value-notation": "number",
+    "no-descending-specificity": [true, { "ignore": ["selectors-within-list"], "severity": "warning" }]
```
renrizzolo, line 7: there are a LOT of instances of this, and it's not auto-fixable.
I've changed it to a warning, but maybe we want to ignore it or attempt to fix the offending css.

.stylelintrc.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": ["stylelint-config-standard", "stylelint-config-prettier"],
+  "rules": {
+    "selector-class-pattern": null,
+    "color-function-notation": null,
```
renrizzolo, line 5: this will allow either modern (`rgba(0 0 0 / 25%)`) or legacy (`rgb(0, 0, 0, 0.25)`) color functions.
https://stylelint.io/user-guide/rules/list/color-function-notation

perhaps we should pick one of these?

xiaofan2406: i think its okay to disable this rule. we are still more familiar with the old notation

xiaofan2406 (approved):

xiaofan2406 (approved):

xiaofan2406 (approved):

PR #1455, opened 2022-07-06
Expose search props in searchable checklist

romi-h:
## Description

- Expose `<Search>` child component props in `<SearchableChecklist>` to support populating the checkable items from server side search.
  - `searchOnEnter`
  - `onSearch`
  - `onClear`
- Added props `hideTitle` to allow alternative design requirement

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/SearchableCheckList/index.d.ts
+++ b/src/components/SearchableCheckList/index.d.ts
@@ -31,7 +31,17 @@ export interface SearchableCheckListProps {
    * Number of checkbox items to render
    */
   displayCount?: number;
+  /**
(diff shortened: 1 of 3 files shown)


adityasabnis:
@romi-h 
Need to have flexibility to set remaining items. Since it would be a page calls
Currently it only subtracts shown items from total items which won't work for async calls

romi-h:
ah yea marking WIP as it depends on the design for https://github.com/Adslot/direct-web/pull/13874

adityasabnis (approved):

PR #1466, opened 2022-08-10
Icons test

renrizzolo:
## Description
This PR is for testing only, to make cleaning up existing icon svgs easier. Icons won't necessarily be in this repo.

- add icons list to docs for testing purposes
- add icon generation pipeline
  - svg files in `icons/src/svg` are converted to react components
  - runs through svgo
  - replaces fill with `currentColor` (currently as color prop that would be overridable, tbd)
  - adds a `size` prop that maps to svg width and height

These could be used directly:
```jsx
import { Adunit } from 'icons'

<Adunit/>
  ```
or 
```jsx
import * as Icons from 'icons'

<Icons.Adunit/>
``` 

- add generic Icon component
  - includes typing of name prop for all available icons
 These would be used like this:
```jsx
import Icon from 'icon'

<Icon name="Adunit"/>
```

not sure of the best API here, but a few things to consider/investigate
- tree shakeability
- importing/intellisense and usage ergonomics
- ease of migration for existing component


package.json
```diff
@@ -32,8 +33,10 @@
     "format": "prettier '{src,www,config}/**/*.{js,json,jsx,md,mdx,css}' --write",
     "lint": "npm run lint:eslint && npm run lint:prettier && npm run lint:stylelint",
     "prettier": "prettier \"{src,www,config}/**/*.{js,json,jsx,md,mdx,css}\" --write",
-    "svgo": "svgo --recursive --multipass --folder=src --config=svgo-config.js",
+    "svgo": "svgo --recursive --multipass --folder=src --config=svgo.config.js",
     "postversion": "git push -u origin $(git rev-parse --abbrev-ref HEAD) --follow-tags && npm publish && echo '…released.'",
+    "generate-icon-names": "node icons/src/generateIconNames.js",
+    "convert-icons": "npm run clean-icons && npx @svgr/cli --out-dir icons/src/react --ext=jsx --template=icons/template.js --expand-props=none --svg-props width={size} --svg-props height={size} --replace-attr-values=#5a5a5a={color},#333={color},currentColor={color} -- icons/src/svg && npm run generate-icon-names",
```
renrizzolo, line 39: This is the main transformation script. 
We don't need the `generate-icon-names` script if we choose to not use the `<Icon name="..."/>` component.

The main thing to note here is 
`--replace-attr-values=#5a5a5a={color}`
if all the source svgs have the same fill colour (e.g `#5a5a5a`), we can use this to replace the fill with `currentColor` string or a `{color}` prop.

the template.js file has the markup for the props etc.

PR #1472, opened 2022-09-05
fix: rich text editor btn aria label

chaofan232:
## Description

Fix the `rich text editor` missing button aria-label to improve direct-web accessibility.
Associated pr: https://github.com/Adslot/direct-web/pull/14209

- [ ] Yes
- [x] No

before:
Screen Shot 2022-09-05 at 9 59 06 pm

after
Screen Shot 2022-09-05 at 10 06 01 pm

### DIFF
--- a/src/components/RichTextEditor/ToolbarButton.jsx
+++ b/src/components/RichTextEditor/ToolbarButton.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import Button from '../Button';
 
-const ToolbarButton = ({ onToggle, label, active }) => {
+const ToolbarButton = ({ onToggle, label, active, ...rest }) => {
   const className = classnames('aui--toolbar-button', {
     active,
   });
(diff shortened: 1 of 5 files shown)


src/components/RichTextEditor/ToolbarButton.jsx
```diff
@@ -18,7 +18,7 @@ const ToolbarButton = ({ onToggle, label, active }) => {
   );
 
   return (
-    <Button className={className} onMouseDown={mouseDownHandler}>
+    <Button className={className} onMouseDown={mouseDownHandler} aria-label={btnAriaLabel}>
```
renrizzolo, line 21: Where possible, I don't think the prop name should be changed/obfuscated. 
Seeing as this is an internal component, I would just spread the rest of the props here, and add `aria-label` to `ToolbarButton` as needed

renrizzolo (approved):

shn2016 (approved):

bchew (approved):
Merging :rocket:

PR #1473, opened 2022-09-17
feat: searchable checklist refactor

adityasabnis:
- Searchable checklist currently has `All` option that selects all options, we need to change that to have it as an independent option.
- Handle use-case when user add `All` option key

## Description

The `All` option is confusing in instances where the options are loaded async. Since the loaded items are not all items. Hence we are changing it to be an independent option. However, if any other option is selected the `All` option is unselected

- [x] Yes
- [ ] No

## Manual testing step?

- SearchableChecklist `All` should not select all options
- Selecting other option should un-select `All`

https://user-images.githubusercontent.com/3688957/190937688-bfb4b55f-47f8-4fc9-902d-22e0d3a941ec.mov

### DIFF
--- a/src/components/SearchableCheckList/index.jsx
+++ b/src/components/SearchableCheckList/index.jsx
@@ -4,7 +4,13 @@ import PropTypes from 'prop-types';
 import Search from '../Search';
 import Checkbox from '../Checkbox';
 import CheckboxGroup from '../CheckboxGroup';
+import TextEllipsis from '../TextEllipsis';
 import './styles.css';
+
+const ALL_OPTION = {
+  value: 'all-option',
+  label: 'All',
+};
 
 const SearchableCheckList = ({
   context,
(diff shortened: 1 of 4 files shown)


src/components/Checkbox/index.jsx
```diff
@@ -93,7 +95,13 @@ const Checkbox = ({
...
             <div>
               <div data-testid="checkbox-label" className="aui--checkbox-label-text">
-                {label}
+                {hasTextTooltip ? (
+                  <Popover placement="top-start" popoverContent={label}>
+                    {label}
+                  </Popover>
+                ) : (
+                  label
+                )}
```
renrizzolo, line 104: Don't think this should be a concern of checkbox

src/components/SearchableCheckList/index.jsx
```diff
@@ -88,11 +85,11 @@ const SearchableCheckList = ({
             name={`${pluralLabel}-group`}
             value={selectedItemsKeys}
             onChange={(newSelectionList) => {
-              onChange(newSelectionList);
+              onChange(_.reject(newSelectionList, (value) => value === 'All'));
             }}
           >
             {_.map(itemsToRender, ({ value, label }) => (
-              <Checkbox key={`${value}-key`} label={label} value={value} dts={`${value}-dts`} />
+              <Checkbox key={`${value}-key`} label={label} value={value} hasTextTooltip dts={`${value}-dts`} />
```
renrizzolo, line 92: I think this can be simplified with
```jsx
label={<TextEllipsis>{label}</TextEllipsis>}
```

and adding width or max width to `.aui--searchable-checklist .aui--checkbox-label-text`

adityasabnis:
Since last review:
- CR feedback changes
- Added check for reserved key
@renrizzolo / @xiaofan2406 

Let me know, what you guy think about this implementation

src/components/SearchableCheckList/index.jsx
```diff
@@ -75,24 +81,29 @@ const SearchableCheckList = ({
...
           <Checkbox
-            checked={mainCheckBoxState}
-            value={`all-${pluralLabel}`}
-            label={`All ${pluralLabel}`}
+            checked={_.includes(itemKeys, ALL_OPTION.value)}
+            value={ALL_OPTION.value}
+            label={ALL_OPTION.label}
             onChange={(nextState) => {
-              onChange(nextState ? _.map(items, 'value') : []);
+              onChange(nextState ? [ALL_OPTION.value] : []);
```
romi-h, line 88: this seems like a breaking change..  maybe safer if we could add an option (in the `context`?) to indicate whether to use the new behaviour, or the default, old behaviour which calls `onChange` with values from all `items`?

renrizzolo:
My only concern is that, at least for the non-async usage, this is a bit of an anti-pattern to have the all option not show all checkboxes as checked, same with not having a partial state.

src/components/SearchableCheckList/index.jsx
adityasabnis: There will be a lot of checks to have backward compatibility.
Maybe we need a different component in that case?

adityasabnis:
> My only concern is that, at least for the non-async usage, this is a bit of an anti-pattern to have the all option not show all checkboxes as checked, same with not having a partial state.

This is a design decision. @pedroscoulloupas wants `All` to be an independent option

src/components/SearchableCheckList/index.jsx
adityasabnis: Can mark it as breaking change and proceed as per offline conversation with @romi-h

romi-h:
Will wait for @xiaofan2406 to weigh in. My earlier concern was to either keep old behaviour as default or mark PR as breaking change (which has been done).

PR #1475, opened 2022-09-19
[ADS-7249] feat: nested checkbox group

xiaofan2406:
## Description

- Allow nested checkbox group.
- Introduce a default `CheckboxGroup.All` checkbox item, as we want the `All` behaviour to be the consistent in different features.
- As a result, create a alias `CheckboxGroup.Item` for `Checkbox` when it is used under `CheckboxGroup`

- [ ] Yes
- [ ] No

### DIFF
--- a/src/components/CheckboxGroup/index.jsx
+++ b/src/components/CheckboxGroup/index.jsx
@@ -1,41 +1,63 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import _ from 'lodash';
 import { expandDts, invariant } from '../../lib/utils';
+import Checkbox, { shareCheckboxPropTypes } from '../Checkbox';
+import CheckboxGroupProvider, { useCheckboxGroup } from './CheckboxGroupContext';
 import '../RadioGroup/style.css';
+import './styles.css';
 
-const CheckboxGroupContext = React.createContext({});
+const CheckboxGroupItem = ({ value, ...rest }) => {
+  const groupCtx = useCheckboxGroup();
+  invariant(!_.isEmpty(groupCtx), 'CheckboxGroup.Item: must be used as children of CheckboxGroup');
+  invariant(!rest.name, 'CheckboxGroup.Item: name will be overridden by CheckboxGroup name');
+  invariant(!rest.variant, 'CheckboxGroup.Item: variant will be overridden by CheckboxGroup variant');
+  invariant(!rest.onChange, 'CheckboxGroup.Item: onChange will be overridden by CheckboxGroup onChange');
 
-const CheckboxGroupProvider = ({ children, name, value, onChange, getIsChecked, variant }) => {
-  const context = React.useMemo(() => {
-    const onCheckboxChange = (checkboxValue) => {
-      const newValues = value.includes(checkboxValue)
-        ? value.filter((item) => item !== checkboxValue)
-        : [...value, checkboxValue];
+  const { onItemChange, getIsItemChecked, name, variant, recordValue } = groupCtx;
 
-      onChange(newValues, name, checkboxValue);
-    };
+  React.useLayoutEffect(() => {
+    recordValue(value);
+  });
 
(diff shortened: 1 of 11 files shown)


src/components/CheckboxGroup/index.jsx
```diff
@@ -62,6 +38,7 @@ const CheckboxGroup = ({
         className={classnames('aui--checkbox-group', className, {
           'is-vertical': orientation === 'vertical',
           'is-default': variant === 'default',
+          'is-nested': isNested && nest,
```
renrizzolo, line 41: so `nest` is purely for the visual part of nesting? Maybe call it indent?

src/components/CheckboxGroup/index.jsx
```diff
@@ -93,10 +70,16 @@ CheckboxGroup.propTypes = {
...
   id: PropTypes.string,
+  nest: PropTypes.bool,
   /**
    *  @deprecated use orientation="horizontal" instead
    **/
   inline: PropTypes.bool,
 };
 
+CheckboxGroup.Item = CheckboxGroupItem;
+CheckboxGroup.All = CheckboxGroupAll;
```
renrizzolo, line 81: To type the dot notation components I think you'll need to define them all in this file.

src/components/CheckboxGroup/index.jsx
xiaofan2406: yep much better

src/components/CheckboxGroup/index.jsx
```diff
@@ -1,41 +1,63 @@
...
+      value={value}
+      variant={variant}
+      checked={getIsItemChecked(value)}
+      onChange={() => onItemChange(value)}
+    />
+  );
 };
 
-export const useCheckboxGroup = () => React.useContext(CheckboxGroupContext);
+CheckboxGroupItem.propTypes = Checkbox.propTypes;
```
renrizzolo, line 36: You can export checkbox propTypes object and use it like this, to get the types onto Item
```js
CheckboxGroupItem.propTypes = { ...checkboxPropTypes };
```
maybe those invariant props should be omitted for this component entirely, so something like 
```js
export const checkboxSharedPropTypes = {
  ...
}

Checkbox.propTypes = {
  ...checkboxSharedPropTypes,
  name: ....
  ...
}
```
```js
CheckboxGroupItem.propTypes = { ...checkboxSharedPropTypes };
```

src/components/CheckboxGroup/index.jsx
```diff
@@ -1,41 +1,63 @@
...
-      onCheckboxChange,
-    };
-  }, [getIsChecked, value, name, onChange, variant]);
-
-  return <CheckboxGroupContext.Provider value={context}>{children}</CheckboxGroupContext.Provider>;
+const CheckboxGroupItem = ({ value, ...rest }) => {
+  const groupCtx = useCheckboxGroup();
+  invariant(!_.isEmpty(groupCtx), 'CheckboxGroup.Item: must be used as children of CheckboxGroup');
+  invariant(!rest.name, 'CheckboxGroup.Item: name will be overridden by CheckboxGroup name');
+  invariant(!rest.variant, 'CheckboxGroup.Item: variant will be overridden by CheckboxGroup variant');
```
renrizzolo, line 15: shouldn't need this as variant isn't a checkbox prop

xiaofan2406: isnt `variant` a valid checkbox prop?

xiaofan2406: this is more or less a intermediate thing, since previsoulsy we had `Checkbox` directly as the child. would hopefully both of these on next release

renrizzolo: with existing checkbox, variant only comes from a checkbox group context (although there might be a case for having it on a single checkbox)

renrizzolo (approved):

chaofan232 (approved):

chaofan232:
merging...

PR #1507, opened 2022-10-18
fix: babel transpile jsx extension

knilink:
## Description

replace `jsx` extension using `babel-plugin-replace-import-extension` which doesn't seem to be a popular plugin.
or maybe can just rename all jsx files in the repo

- [ ] Yes
- [ ] No


xiaofan2406:
@knilink is this still needed?

PR #1523, opened 2022-11-24
React18

xiaofan2406:
## Description

- deprecate `fastStatelessWrapper`, as it does not do what it is supposed to do, use `React.memo`
- deprecate `HoverDropdownMenu`, as it is not used, and should be used with `Popover` and custom menu content
- deprecate `Toast.Notification`, as it has a bug and should use `Toast.notify` as a replacement
- upgrade React 18

*internal*
- lowered coverage percentage to 95% 
- react testing lib v13
- migrate most events to use user-events v14, (except for `RichTextEditor`)
- fix some tests had incorrect mocking for `console.error` which silenced lots of valid errors
- separated out `invariant` to a isolated file and ignore by coverage since it is auto mocked during test
- update linting setup for test files

- [x] Yes
- [ ] No


package.json
```diff
@@ -177,12 +179,17 @@
...
-    "react": "^16.14.0",
-    "react-dom": "^16.14.0"
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
   },
   "engines": {
     "node": "^18"
   },
+  "overrides": {
+    "@testing-library/dom": "^9.0.1",
```
xiaofan2406, line 189: https://github.com/testing-library/user-event/issues/1104

shn2016 (approved):

shn2016:
Merging...

vinteo (approved):

PR #1562, opened 2023-01-31
fix: add flattened vars json

renrizzolo:
Add a flattened JSON version of our design system tokens

## Description

This output can be used via postcss-simple-variables plugin, in order to not have to import the css itself

```js
const tokens = require('adslot-ui/system/internal/variables.json');

require('postcss-simple-vars')({
  variables: tokens
})
```

- [ ] Yes
- [x] No


system/config.js
```diff
@@ -63,5 +63,29 @@ module.exports = {
...
+     * ```
+     * const tokens = require('adslot-ui/system/internal/variables.json');
+     * 
+     * require('postcss-simple-vars')({
+        variables: tokens
+       })
+     * ```
+     */
+    postcssVariablesJSON: {
+      buildPath: `internal/`,
```
renrizzolo, line 78: Have put this under internal because, although it will be used 'externally', we don't want this to be used for referencing vars in JS directly.

xiaofan2406 (approved):

PR #1564, opened 2023-02-07
fix: allow checkbox item to be partial

xiaofan2406:
## Description

- when a child checkbox is `partial`, the parent checkbox should be `partial` as well
- change the toggle order to partial checkbox. in master it goes `partial -> false -> true -> partial`, now it is changed to `partial -> true -> false -> partial`

- [ ] Yes
- [x] No


pphminions (approved):
LGTM 🐈‍⬛

pphminions (approved):
LGTM 🪜

renrizzolo (approved):

PR #1569, opened 2023-03-01
fix: allow checkbox group within radio group

vinteo:
## Description

This allows the `CheckboxGroup` of variant `default` to be used within a `RadioGroup` with variant `box` without the CSS for the box `RadioGroup` affecting the `CheckboxGroup` items.

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/Checkbox/index.jsx
+++ b/src/components/Checkbox/index.jsx
@@ -74,8 +74,9 @@ const Checkbox = ({
       onKeyDown={onCheckboxKeyDown}
       {...expandDts(dts)}
     >
-      <label htmlFor={id}>
+      <label className="aui--checkbox-label-container" htmlFor={id}>
         <input
+          className="aui--checkbox-input"
           data-testid="checkbox-input"
           type="checkbox"
           name={name}
--- a/src/components/Checkbox/styles.css
+++ b/src/components/Checkbox/styles.css
@@ -7,19 +7,19 @@ $label-gap: 8px;
   cursor: pointer;
   display: block;
 
-  & label {
+  & .aui--checkbox-label-container {
     cursor: pointer;
     font-weight: $font-weight-normal;
     margin-bottom: 0;
     display: flex;
   }
 
-  & input {
+  & .aui--checkbox-input {
     display: none;
   }
 
   &.is-reverse {
-    & label {
+    & .aui--checkbox-label-container {
       flex-direction: row-reverse;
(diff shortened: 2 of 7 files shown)


pphminions (dismissed):
LGTM 🐈‍⬛

romi-h (dismissed):

src/components/Checkbox/index.jsx
```diff
@@ -74,8 +74,9 @@ const Checkbox = ({
       onKeyDown={onCheckboxKeyDown}
       {...expandDts(dts)}
     >
-      <label htmlFor={id}>
+      <label className="aui--checkbox-label-container" htmlFor={id}>
```
xiaofan2406, line 77: can we shorten this to just `aui--checkbox-label`

src/components/Checkbox/styles.css
```diff
@@ -7,19 +7,19 @@ $label-gap: 8px;
   cursor: pointer;
   display: block;
 
-  & label {
+  & label.aui--checkbox-label-container {
```
xiaofan2406, line 10: does it work without `label` now? reducing level would make overriding easier

src/components/Checkbox/styles.css
```diff
@@ -7,19 +7,19 @@ $label-gap: 8px;
...
     cursor: pointer;
     font-weight: $font-weight-normal;
     margin-bottom: 0;
     display: flex;
   }
 
-  & input {
+  & input.aui--checkbox-input {
```
xiaofan2406, line 17: same here, if they work without the tags, we can use the classname only

src/components/Checkbox/index.jsx
renrizzolo: `aui--checkbox-label` is used to wrap the label text within the label.
I think we should swap the class names so the label element has the label class and the text wrapper has the container class.

```jsx
<label className="aui--checkbox-label">
  <div className="aui--checkbox-input-container">
    ...
  <div className="aui--checkbox-label-container">
   ...
```

vinteo: there is already an `aui--checkbox-label`

vinteo: it looks like there are some overrides in direct web for that class

xiaofan2406 (approved):

renrizzolo (approved):

pphminions:
merging 🚀

PR #1576, opened 2023-03-07
chore: dedupe packages

knilink:
## Description

bundle contains duplicated modules
image

- [ ] Yes
- [x] No

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -121,11 +121,24 @@
       "dev": true
     },
     "node_modules/@ampproject/remapping": {
-      "version": "2.1.2",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.1.2.tgz",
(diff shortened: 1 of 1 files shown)


renrizzolo:
Not sure about editing the lockfile directly. Is it possible to use npm `overrides`?

I recall this from when I was looking at tree shaking ui – see:

https://github.com/Adslot/direct-web/pull/13506/files#diff-f1299ce9225bcc865f3ea5bf3ae73d0f068641b8231d5fe9c66b04e4caa25f8bR317-R318

knilink:
@renrizzolo npm install is supposed to dedupe packages but something duplicating will still happen when merging package-lock from different PRs. 

I just ran `npm dedupe` and I think reinstalling immutable dependent packages will get the same result.

bchew (approved):

PR #1577, opened 2023-03-08
fix: [useCollapse] cleanup resize observer, overflow css

renrizzolo:
## Description
- as per https://github.com/WICG/resize-observer/issues/38,  wrap the observer in `requestAnimationFrame` so we don't get the `Loop limit exceeded` error.
- Fix panel overflow/dispaly styles by adding transition state to `useCollapse`, so we can hide overflow only when animating
  - previously I set hidden overflow on the panel so you wouldn't see the content outside of the panel when toggling the collapse. Discovered we have some useage of Panel with Select inside them, which would be hidden by the overflow. 
- some cleanup to the collapse internals

- [ ] Yes
- [x] No

no overflow / without fix
Kapture 2023-03-09 at 11 29 13

with overflow/display transition classes
Kapture 2023-03-09 at 11 28 28

### DIFF
--- a/src/hooks/useCollapse.js
+++ b/src/hooks/useCollapse.js
@@ -3,60 +3,90 @@ import React from 'react';
 
 /**
  * @param {Object} useCollapse
- * @param {number} useCollapse.collapsedHeight height to collapse to (default 0)
+ * @param {number} useCollapse.collapsedHeight height to collapse to if the collapsed content's height is greater (default 0)
  * @param {number} useCollapse.collapsed optionally control collapsed state
+ * @param {number} useCollapse.transitionMs transition time in ms, for transitionState
  */
-export const useCollapse = ({ collapsedHeight = 0, collapsed: collapsedProp = false }) => {
+export const useCollapse = ({ collapsedHeight = 0, collapsed = false, transitionMs }) => {
   const containerRef = React.useRef();
-  const elHeightRef = React.useRef();
-  const collapsedHeightRef = React.useRef();
+  const animationFrameRef = React.useRef();
+  const transitionTimerRef = React.useRef();
 
-  const [height, _setHeight] = React.useState();
-  const [collapsed, setCollapsed] = React.useState(collapsedProp);
+  const [_height, _setHeight] = React.useState();
+  const [animationState, setAnimationState] = React.useState();
 
-  const setHeight = React.useCallback(
-    (elHeight) => {
-      if (!containerRef.current) return;
-      collapsedHeightRef.current = _.isNumber(collapsedHeight) ? Math.min(collapsedHeight, elHeight) : elHeight;
-      if (!collapsed) {
-        _setHeight(elHeight);
-      } else {
-        _setHeight(collapsedHeightRef.current);
-      }
-    },
-    [collapsedHeight, collapsed]
-  );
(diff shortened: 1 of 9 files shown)


shn2016 (approved):

romi-h (dismissed):

src/hooks/useCollapse.js
```diff
@@ -5,58 +5,96 @@ import React from 'react';
...
+    if (!collapsed) {
+      _setHeight(elHeight);
+    } else {
+      _setHeight(collapsedHeightRef.current);
+    }
+  }, [collapsedHeight, elHeight, collapsed]);
+
+  React.useEffect(() => {
     setCollapsed(collapsedProp);
   }, [collapsedProp]);
```
knilink, line 55: guess this is because both `collapsedProp` and `toggleCollapsed` are trying control `collapsed` state
i think should be either
`(defaultCollapsed, ...) => {collapsed, toggleCollapsed, ...}`
or
`(colleapsed, ...) => {...}`

src/hooks/useCollapse.js
```diff
@@ -5,58 +5,96 @@ import React from 'react';
...
-      for (const entry of entries) {
-        // borderBoxSize is not an array in old versions of Firefox
-        elHeightRef.current = _.castArray(entry.borderBoxSize)[0]?.blockSize ?? entry.contentRect.height;
+    if (!resizeObserverRef.current) {
+      resizeObserverRef.current = new ResizeObserver((entries) => {
+        // see https://github.com/WICG/resize-observer/issues/38
+        animationFrameRef.current = window.requestAnimationFrame(() => {
+          for (const entry of entries) {
+            // borderBoxSize is not an array in old versions of Firefox
+            setElHeight(_.castArray(entry.borderBoxSize)[0]?.blockSize ?? entry.contentRect.height);
```
knilink, line 74: i guess can just modify container height here with element ref so that the height state will not be needed

renrizzolo: not sure about this, I couldn't get it to work for all scenarios. I prefer to return the height from the hook.

src/hooks/useCollapse.js
renrizzolo: yes, defaultCollapsed is better

knilink: then i guess this can be removed as default props shouldn't be affecting the element state during its life cycle, the same as `defaultValue` to `input` component

https://github.com/Adslot/adslot-ui/blob/28fcf38c41a89e2d27f2a6e98b82c2f72ee7debf/src/hooks/useCollapse.js#L53-L55

renrizzolo: oh, it's still being used to control the component.

knilink: what i mean was both `toggleCollapsed` and `collapsedProp` control the `collapse`  state, should be just keeping one.

renrizzolo: I see what you mean, I'll remove the toggleCollapsed

renrizzolo: @knilink removed internal collapsed state and toggle function

knilink (approved):

romi-h (approved):

xiaofan2406 (approved):

PR #1724, opened 2023-10-16
fix icons on rich text editor component

glenkurniawan-adslot:

- [ ] Yes
- [x] No

before
image

after
image


glenkurniawan-adslot:
@xiaofan2406 there might be a way to do it properly with the plugins and configs, but i do not want to spend too much time on it.. a simple refactor to img works, but it might not be the most elegant solution

xiaofan2406 (approved):

PR #1761, opened 2024-02-27
[ASC-169] fix checkbox group all selection

glenkurniawan-adslot:
## Description

- update to find out checkbox items that are disabled and do not include them when checkboxgroup.all is being selected
- update to disable checkboxgroup.all when all the children are disabled

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/CheckboxGroup/CheckboxGroupContext.jsx
+++ b/src/components/CheckboxGroup/CheckboxGroupContext.jsx
@@ -20,6 +20,8 @@ const CheckboxGroupProvider = ({
   const name = parentCtx.name || nameProp;
   const value = parentCtx.value || valueProp;
 
+  const [disabledValues, setDisabledValues] = React.useState([]);
+
   const context = React.useMemo(() => {
     const getIsItemChecked = (checkboxValue) => {
       if (getIsChecked) return getIsChecked(checkboxValue, value);
@@ -28,11 +30,11 @@ const CheckboxGroupProvider = ({
     };
 
     const getIsAllChecked = (values) => {
+      if (_.isEmpty(values)) return false;
       const result = _(values)
         .map((item) => getIsItemChecked(item))
         .uniq()
         .value();
-
       return result.length === 1 ? result[0] : 'partial';
     };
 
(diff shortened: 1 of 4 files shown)


src/components/CheckboxGroup/CheckboxGroupContext.jsx
```diff
@@ -21,18 +21,37 @@ const CheckboxGroupProvider = ({
   const value = parentCtx.value || valueProp;
 
   const context = React.useMemo(() => {
+    const disabledComponents = [];
+    const findDisabledComponents = (c) => {
+      if (c.props?.children) {
+        findDisabledComponents(c.props.children);
+      } else if (Array.isArray(c)) {
+        _.forEach(c, (child) => {
```
xiaofan2406, line 29: as per offline discussion, we should avoid search for children, since fragment nesting is unlimited. try let the children "register" itself as it mounts

src/components/CheckboxGroup/index.spec.jsx
```diff
@@ -339,3 +339,142 @@ it('should work when the values are updated', async () => {
...
+      </CheckboxGroup>
+    );
+  };
+
+  render(<Component />);
+
+  await user.click(screen.getByTestId('button'));
+  const items = screen.queryAllByTestId('checkbox');
+  await user.click(within(items[0]).getByTestId('checkbox-input')); // click checkbox all
+  expect(items[1]).toBeChecked(); // terminator is checked
```
xiaofan2406, line 476: should we be asserting the parent disabled status as well before & after?

glenkurniawan-adslot: do you mean something like this:
parent initial status is disabled (children all disabled) then add a new children that is not disabled, expect parent is NOT disabled?

xiaofan2406: yea i think some tests like this should be useful

glenkurniawan-adslot: @xiaofan2406 i've added the test

xiaofan2406 (approved):

